PR #1, opened 2023-02-08
Feature/enable SwaggerUI

tnotheis:


Backbone.API/Configuration/BackboneConfiguration.cs
```diff
@@ -40,7 +40,7 @@ public class CorsConfiguration
     public class SwaggerUiConfiguration
     {
         [Required]
-        public string TokenUrl { get; set; }
+        public string TokenUrl { get; set; } = "";
```
daniel-almeida-konkconsulting, line 43: can we use string.Empty here?

daniel-almeida-konkconsulting, line 43: can we use string.Empty here?

daniel-almeida-konkconsulting (changes requested):

Backbone.API/Configuration/BackboneConfiguration.cs
tnotheis: I wanted to address changing "" to string.Empty in a separate PR.

daniel-almeida-konkconsulting (approved):

PR #2, opened 2023-02-08
Feature(devices)/add remaining projects to solution

tnotheis:


Modules/Devices/src/Devices.AdminCli/Program.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System.Reflection;
 using System.Text.Json;
-using Devices.Domain.Entities;
-using Devices.Infrastructure.Persistence.Database;
+using Backbone.Modules.Devices.Domain.Entities;
+using Backbone.Modules.Devices.Infrastructure.Persistence.Database;
```
daniel-almeida-konkconsulting, line 4: Shouldn't the Devices.AdminCli project be using a similar prefix in its namespace?
e.g. **Backbone.Modules.Devices.AdminCli**

daniel-almeida-konkconsulting (changes requested):

Modules/Devices/src/Devices.AdminCli/Program.cs
tnotheis: Good catch!

daniel-almeida-konkconsulting (approved):

PR #5, opened 2023-02-08
Fix/disable-deprecated-fluentvalidation-features

tnotheis:


tnotheis:
I moved the call to `AddFluentValidationAutoValidation` to the `AddCustomFluentValidation` method and removed the setting of the two deprecated properties `ImplicitlyValidateRootCollectionElements` and `ImplicitlyValidateChildProperties`.
For completeness: the deprecation of these properties is documented here: https://github.com/fluentValidation/FluentValidation/issues/1960

daniel-almeida-konkconsulting (approved):

PR #8, opened 2023-02-09
Ci/dont-show-build-warnings-during-pipeline-builds

tnotheis:


tnotheis:
The reason why I turned it off is that there are currently lots of warnings that we chose to ignore for now. So they are only messing up the log.

jkoenig134 (approved):

jkoenig134 (approved):

PR #9, opened 2023-02-10
Feature(messages)/add storage sanity check

JLSRKonk:


Messages.Jobs.SanityCheck/Infrastructure/Reporter/LogReporter.cs
```diff
@@ -0,0 +1,42 @@
+﻿using Backbone.Modules.Messages.Domain.Ids;
+
+namespace Messages.Jobs.SanityCheck.Infrastructure.Reporter
+{
+    public class LogReporter : IReporter
+    {
+        private readonly ILogger<LogReporter> _logger;
+        private readonly List<MessageId> _databaseids;
+        private readonly List<string> _blobids;
```
tnotheis, line 9: Casing...

tnotheis (changes requested):
Why are the jobs in the root directory the repository, and not in "Modules/Messages"?

Messages.Jobs.SanityCheck/Infrastructure/Reporter/LogReporter.cs
JLSRKonk: I am actually clueless as to why the projects were created in the root directory. I created them inside the 'Modules/Messages' folder, and they were displayed inside of that folder in VStudio, but for some reason the actual files were outside the folder... Either way, it's now fixed in the last commit.

tnotheis (approved):

PR #13, opened 2023-02-13
Feature(tokens)/add storage sanity check

JLSRKonk:
This service was the one that needed to use the TokenRepository. I added a new method there called GetAllTokenIds().
Also I noticed that the Azure BlobStorage for 'tokens' was returning blobs with prefixes other than the token one, is this intended?


tnotheis:
Humpf. Damn. I just noticed there is a conceptual problem regarding the blob storage. We need to introduce a separate blob storage interface for each module, similar to the db context interface (ITokensBlobStorage, IRelationshipsBlobStorage, ...)

Modules/Tokens/src/Tokens.Infrastructure/Persistence/Repository/TokenRepository.cs
```diff
@@ -64,6 +64,11 @@ public async Task<IdentityAddress> GetOwner(TokenId tokenId)
         return result.CreatedBy;
     }
 
+    public async Task<IEnumerable<TokenId>> GetAllTokenIds()
+    {
+        return await _readonlyTokensDbSet.Where(Token.IsNotExpired).Select(t => t.Id).ToListAsync();
```
tnotheis, line 69: Can you add a `includeExpired` parameter, which defaults to `false`?

tnotheis:
But we should do this in a separate PR

tnotheis (changes requested):

Modules/Tokens/src/Tokens.Infrastructure/Persistence/Repository/TokenRepository.cs
```diff
@@ -64,8 +64,13 @@ public async Task<IdentityAddress> GetOwner(TokenId tokenId)
...
-    public async Task<IEnumerable<TokenId>> GetAllTokenIds()
+    public async Task<IEnumerable<TokenId>> GetAllTokenIds(bool includeExpired = false)
     {
+        if (includeExpired)
+        {
+            return await _readonlyTokensDbSet.Select(t => t.Id).ToListAsync();
+        }
+
         return await _readonlyTokensDbSet.Where(Token.IsNotExpired).Select(t => t.Id).ToListAsync();
     }
```
tnotheis, line 75:
```suggestion
    public async Task<IEnumerable<TokenId>> GetAllTokenIds(bool includeExpired = false)
    {
        IQueryable<Token> query = _readonlyTokensDbSet;
        
        if(!includeExpired)
            query = query.Where(Token.IsNotExpired)
        
        return await _readonlyTokensDbSet.Select(t => t.Id).ToListAsync();
    }
```

tnotheis (changes requested):

Modules/Tokens/src/Tokens.Infrastructure/Persistence/Repository/TokenRepository.cs
tnotheis: Not sure if the syntax is 100% correct.

Modules/Tokens/src/Tokens.Jobs.SanityCheck/Infrastructure/DataSource/DataSource.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public async Task<IEnumerable<string>> GetBlobIdsAsync(CancellationToken cancellationToken)
+        {
+            var blobIds = await _blobStorage.FindAllAsync();
+            return await blobIds.ToListAsync(cancellationToken);
+        }
+
+        public async Task<IEnumerable<TokenId>> GetDatabaseIdsAsync(CancellationToken cancellationToken)
+        {
+            return await _tokenRepository.GetAllTokenIds();
```
tnotheis, line 26: Oh, the actual reason for asking for the change was, that I wanted you to check the content of expired tokens as well. :)
So you should pass true here. Or to be more expressive: pass `includeExpired: true`)

tnotheis (changes requested):

tnotheis (approved):

PR #18, opened 2023-02-17
Feature(challenges)/add api client

KONKdiogo-ferraz:
Add the Challenges API Client


Modules/Challenges/src/Challenges.API.Client/API/ChallengesClient.cs
```diff
@@ -0,0 +1,51 @@
...
+    public async Task<RestResponse<AccessTokenResponse>> GetAccessTokenAsync(AuthenticationRequest authenticationRequest)
+    {
+        var request = new RestRequest("/connect/token", Method.Post);
+        request.AddParameter("grant_type", "password");
+        request.AddParameter("client_id", authenticationRequest.ClientId);
+        request.AddParameter("client_secret", authenticationRequest.ClientSecret);
+        request.AddParameter("username", authenticationRequest.Username);
+        request.AddParameter("password", authenticationRequest.Password);
+
+        return await _client.ExecuteAsync<AccessTokenResponse>(request);
```
tnotheis, line 27: Did you test that? Because I wouldn't expect this to work. The /connect/token endpoint expects the content type `application/x-www-form-urlencoded`, and I don't think that the client uses that by default.

Modules/Challenges/src/Challenges.API.Client/API/ChallengesClient.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public async Task<RestResponse<ChallengeResponse>> CreateChallengeAsync(string accessToken = default)
```
tnotheis, line 30: As a user of the Client you shouldn't need to pass an access token. Instead, the client should get one if there is none or if the existing one is expired.
There are probably some http clients out there which have built-in support for OAuth, and which help you managing the access token.

tnotheis (changes requested):

tnotheis:
I was just trying to create a class diagram for the HTTP client, because I wanted to put some details into the Jira issue description. But the more I think about it, the more I drift away from the opinion that we actually should use such a client.
Maybe it's better to use a plain HttpClient. The reason is that, especially in the integration tests, we need a certain degree of flexibility in calling the requests, which a strongly typed client shouldn't even offer (it's the purpose of a strongly typed client to only allow valid requests). E.g. we could want to send invalid JSON or try to get a Challenge without a JWT, in order to test the authorization.
That was probably the reason why you implemented it the way you did, and not the way I expected you to implement it. 😉 You were looking at it in the context of using it in the tests, and I only thought about how to design a "normal" strongly typed client.
Conclusion:
* We should close the PR with the client and delete the client.
* We should use an HttpClient in the tests
* In the future I will try to think more about the tasks before I create them.

PR #19, opened 2023-02-20
Test(files)/add performance tests

JLSRKonk:


Modules/Files/test/Files.API.Tests.Performance.EndToEnd/run-all.ps1
```diff
@@ -0,0 +1,17 @@
...
+
+Write-Host "Running files performance tests..."
+
+npm install
+node ".\node_modules\webpack\bin\webpack.js"
+
+Get-ChildItem -Path "./dist" -Filter *.test.js |
+
+Foreach-Object {
+    k6 run -e HOST=$baseUrl -e SIZE=$size -e CLIENT_SECRET=test $_.FullName
```
tnotheis, line 14: Via chat I told you that the way to run composite tests, as described in the first comment of the GitHub issue you found (https://github.com/grafana/k6/issues/1342), looks promising. Why didn't you implement that instead of running one by one?

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/get-download.test.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export function getSizeIterations(size: string){
+    switch(size){
+        case "S": return 10;
+        case "M": return 50;
+        case "L": return 100;
+        default: throw new Error("Invalid 'Size' value");
+    }
+}
```
tnotheis, line 29: How about the following method to replace the two existing ones?
```ts
function translateSize(size: string) {
    switch(size){
        case "S": return { vus: 1, iterations: 10 };
        case "M": return { vus: 10, iterations: 50 };
        case "L": return { vus: 50, iterations: 100 };
        default: throw new Error("Invalid 'Size' value");
    }
}
```

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/get-download.test.ts
```diff
@@ -0,0 +1,103 @@
...
+                    "Authorization": `Bearer ${data.authToken}`,
+                }
+            }
+        );
+
+        expect(response.status, "response status").to.equal(200);
+    });
+};
+
+export function getDate(){
```
tnotheis, line 98: That's not a good name for this function, since it returns "tomorrow". Most of the time, when getting a date and not specify anything, you receive the current date. So the method name raises wrong expectations.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/get-download.test.ts
```diff
@@ -0,0 +1,103 @@
+import http from "k6/http";
```
tnotheis, line 1: For the whole file: only export stuff that is used in other files/modules.

tnotheis, line 1: I would change the filename to `download-file.test.ts`

tnotheis (commented):
This is only a partial review because I have to time to complete it right now. But I think that one of the comments is something that requires a bit more work, which is why I wanted to commit it right away.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/get-download.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        const response = http.get(
+            `${apiEndpoint}Files/${data.fileId}`,
+            {
+                headers: {
+                    "Authorization": `Bearer ${data.authToken}`,
+                }
+            }
+        );
+
+        expect(response.status, "response status").to.equal(200);
```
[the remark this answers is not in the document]
JLSRKonk: I tried without the "response status", and the test results became less clear. Instead of showing **✓ expected response status to equal 200** the result comes out as **✓ expected ${this} to equal 200**.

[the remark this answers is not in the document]
tnotheis: Okay, that's ugly. Then we should stay with your current solution of course.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "webpack": "^5.75.0",
+    "webpack-cli": "4.6.0",
+    "webpack-glob-entries": "^1.0.1"
+  },
+  "scripts": {
+    "test:get": "k6 run -e HOST=http://localhost:5000 -e SIZE=S -e CLIENT_SECRET=test dist/get.test.js",
+    "test:get-metadata": "k6 run -e HOST=http://localhost:5000 -e SIZE=S -e CLIENT_SECRET=test dist/get-metadata.test.js",
+    "test:download-file": "k6 run -e HOST=http://localhost:5000 -e SIZE=S -e CLIENT_SECRET=test dist/download-file.test.js",
+    "test:post-with-authentication": "k6 run -e HOST=http://localhost:5000 -e SIZE=S -e CLIENT_SECRET=test dist/post-with-auth.test.js",
+    "test": "webpack && npm run test:get && npm run test:get-metadata && npm run test:download-file && npm run test:post-with-authentication"
```
tnotheis, line 31: Remove these npm scripts. We have the (power)shell scripts for that.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "typescript",
```
tnotheis, line 2: Give it an appropriate name.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "1.0.0",
+  "main": "index.js",
+  "repository": "ssh://git@github.com/k6io/example-typescript.git",
+  "author": "Simon Aronsson <simme@k6.io>",
```
tnotheis, line 6: Remove these properties

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/download-file.test.ts
```diff
@@ -0,0 +1,94 @@
...
+const host = __ENV.HOST;
+const apiEndpoint = host+"/api/v1";
+
+interface Data {
+    authToken: string,
+    fileId: string
+}
+
+function translateSize() {
+    switch(__ENV.SIZE){
```
tnotheis, line 14: Did you run prettier? Because I would expect there to be a space between `)` and `{`.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/prettierrc.json
```diff
@@ -0,0 +1,15 @@
+{
+    "endOfLine": "lf",
+    "printWidth": 120,
+    "tabWidth": 4,
+    "trailingComma": "none",
+    "semi": false,
```
tnotheis, line 6: Set it to true (I forgot that they don't use semicolons for most of the enmeshed consumption projects)

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/download-file.test.ts
```diff
@@ -0,0 +1,94 @@
...
+                    "Authorization": `Bearer ${data.authToken}`,
+                }
+            }
+        );
+
+        expect(response.status, "response status").to.equal(200);
+    });
+};
+
+function getTomorrowDate(){
```
tnotheis, line 89: I think I would just call the function `tomorrow()`.

Then the usage would look like

```ts
expiresAt: tomorrow()
```
which in my opinion reads quite nicely.

Sometimes, and this is true especially for tests, it's a good idea to rethink common naming conventions (like prepend `get` to the name of a function if it gets something), in order to improve the readability.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/get.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        bodyConnectToken,
+        {
+            headers: {
+                "Content-Type": "application/x-www-form-urlencoded"
+            }
+        }
+    ).json("access_token")!
+    .toString();
+
+    return { authToken: authToken }
```
tnotheis, line 49: This code is duplicated among different files. You should extract it into a separate file which is then imported by the tests.
I think for now a `utils.ts` file with a getJwt() function is sufficient.

tnotheis (changes requested):

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/run-all.ps1
```diff
@@ -1,19 +1,19 @@
 Param(
-    [parameter(Mandatory)] $baseUrl,
-    [parameter(Mandatory)] $userName,
+    [parameter(Mandatory)] $baseurl,
```
tnotheis, line 2: I was just complaining about username, because it's common to have it in lowercase. You can leave baseUrl as it was before.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/download-file.test.ts
JLSRKonk: So I ran prettier, but I mistyped the command so It didn't do anything. I also installed the extension, but didn't set it up to change on save, so it didn't do anything either. :) It's fixed now.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/download-file.test.ts
```diff
@@ -0,0 +1,77 @@
+import http from "k6/http";
+import { Options } from "k6/options";
+import {
+  describe,
+  expect,
+} from "https://jslib.k6.io/k6chaijs/4.3.4.2/index.js";
+import { getJwt, assertEnvVarExists, Size, tomorrow } from "./utils";
+
+assertEnvVarExists();
```
tnotheis, line 9: I actually wanted you to have a method to which you pass the name of an env var, which checks if an env var with this name exists.

Since all the tests need the same env vars, it's maybe actually a good idea to have a method like your current implementation as well. But this method should have a more appropriate name then (something like `assertAllRequiredEnvVarsExist`).

Or maybe even better: a method called `getConfiguration()`, which validates that all required env vars exist, reads them and returns them in a strongly typed object.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/download-file.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    .toString();
+
+  return {
+    authToken: authToken,
+    fileId: fileId,
+  };
+}
+
+export default function (data: Data): void {
+  describe("Get a File download:", () => {
```
tnotheis, line 68:
```suggestion
  describe("Download a file:", () => {
```

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/get-metadata.test.ts
```diff
@@ -0,0 +1,77 @@
...
+export const options: Options = {
+  vus: size().vus,
+  thresholds: {
+    http_req_duration: ["p(90)<50", "p(98)<100"],
+  },
+  iterations: size().iterations,
+};
+
+export function setup(): Data {
+  const authToken = getJwt();
```
tnotheis, line 40: Maybe we should call the method `getAuthenticationHeader` and return `Bearer ${JWT}`, so that we don't have to concatenate it everywhere.

tnotheis (changes requested):
Looks good so far. I just have one minor comment and two new ideas.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/download-file.test.ts
JLSRKonk: After thinking about it for a bit, I went for the getConfiguration() method implementation. For the way we currently have the tests setup (including the run-all scripts), I think it makes sense to have the validations done all at once, and return a configuration object with the actual "options", I don't think we need a function that validates each env var separately, at least for now.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export function tomorrow(): Date {
+  const date = new Date();
+  date.setDate(date.getDate() + 1);
+
+  return date;
+}
+
+export function getAuthenticationHeader(configuration: Configuration): string {
```
tnotheis, line 24: My bad this time. It should be called `getAuthorizationHeader`, of course.

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+    exec.test.abort("Parameter 'USERNAME' cannot be null or empty");
+  }
+
+  if (!__ENV.PASSWORD) {
+    exec.test.abort("Parameter 'PASSWORD' cannot be null or empty");
+  }
+
+  if (!__ENV.SIZE) {
+    exec.test.abort("Parameter 'SIZE' cannot be null or empty");
+  }
```
tnotheis, line 64: I would probably still add a method assertEnvVarExists, which you can use in this method. Just to reduce duplication.
```ts
assertEnvVarExists("HOST");
assertEnvVarExists("CLIENT_SECRET");
assertEnvVarExists("USERNAME");
assertEnvVarExists("PASSWORD");
assertEnvVarExists("SIZE");
```

tnotheis (dismissed):

Modules/Files/test/Files.API.Tests.Performance.EndToEnd/src/utils.ts
JLSRKonk: Ok, I implemented this method now. While testing this I noticed that when the USERNAME variable wasn't set, k6 was using  my username for some reason. It seems that USERNAME is already used as an environment variable by default (I think), so when we pass it as a parameter, it is overwritten, so I had to change it to USER instead.

tnotheis (approved):

PR #24, opened 2023-03-02
Test(challenges)/implement tests with specflow

KONKdiogo-ferraz:


Modules/Challenges/test/SpecFlowChallenges.Specs/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,219 @@
...
+            challengeResponse.IsSuccessStatusCode.Should().BeTrue();
+
+            if (challengeResponse.Data != null)
+            {
+                _challengeId = challengeResponse.Data?.Result?.Id;
+                _challengeId.Should().NotBeNullOrEmpty(because: "Required value for 'Id' is missing.");
+            }
+            else
+            {
+                challengeResponse.Data.Should().NotBeNull();
```
tnotheis, line 84: This assertion will **always** fail. So you could as well write it before the `if` and remove the `if` afterwards. But as stated above, you don't need the `if` at all, because at this place we can assume that the POST-route does its job.

Modules/Challenges/test/SpecFlowChallenges.Specs/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        var requestConfiguration = table.CreateInstance<RequestConfiguration>();
+        _challengeResponse = await _challengeApi.CreateChallengeAsync(requestConfiguration);
+    }
+
+    [When(@"a POST request is sent to the Challenges endpoint")]
+    public async Task WhenAPOSTRequestIsSentToTheChallengesEndpoint()
+    {
+        // Check if the request was made by an authenticated user
+        _isAuthenticatedChallenge = !string.IsNullOrEmpty(_requestConfiguration.TokenResponse?.AccessToken);
```
tnotheis, line 103: Didn't I tell you that you shouldn't infer authentication from the content of the access token? 
Because there could be an expired access token in it.

Modules/Challenges/test/SpecFlowChallenges.Specs/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,219 @@
...
+    {
+        switch (id)
+        {
+            case "c.Id":
+                id = _challengeId!;
+                break;
+            case "a valid Id":
+                id = "CHLjVPS6h1082AuBVBaR";
+                break;
+        }
```
tnotheis, line 119: Maybe you should throw in any other case, in order to get a better error message in case someone passes something invalid..

Modules/Challenges/test/SpecFlowChallenges.Specs/API/ChallengesApi.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        foreach (var parameter in authenticationParams.Parameters)
+        {
+            request.AddParameter(parameter.Key, parameter.Value);
+        }
+
+        return await _client.ExecuteAsync<AccessTokenResponse>(request);
+    }
+
+    public async Task<RestResponse<ChallengeResponse>> CreateChallengeAsync(RequestConfiguration requestConfiguration)
```
tnotheis, line 33: You shouldn't return RestResponse here, in order to stay independent of the used rest client.

Modules/Challenges/test/SpecFlowChallenges.Specs/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,219 @@
...
+        if (_challengeResponse.Content != null)
+        {
+            var errorMessage = JsonConvert.DeserializeObject<ErrorResponse>(_challengeResponse.Content);
+            if (errorMessage != null)
+            {
+                errorMessage.Error.Code.Should().Be(errorCode);
+            }
+            errorMessage.Should().NotBeNull();
+        }
+        _challengeResponse.Content.Should().NotBeNull();
```
tnotheis, line 135: Again this weird "pattern" in you place an assertion after an if, which both check for the same condition. You can simplify this a lot:

```suggestion
        _challengeResponse.Content.Should().NotBeNull();
        
        var error = JsonConvert.DeserializeObject<ErrorResponse>(_challengeResponse.Content);        
        
        error.Should().NotBeNull();        
        error.Error.Code.Should().Be(errorCode);
```

I've seen this in some other places as well. Please make sure to refactor it everywhere.

tnotheis (changes requested):

Modules/Challenges/test/SpecFlowChallenges.Specs/Models/AuthenticationParams.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace SpecFlowChallenges.Specs.Models;
+public class AuthenticationParameters
+{
+    public Dictionary<string, string> Parameters { get; set; }
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: My intention here was to make the code more dynamic since we can later add all the parameters to the HTTP request in one go, instead of multiple lines of code to add each parameter.

Modules/Challenges/test/SpecFlowChallenges.Specs/Utils/JsonValidators.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Schema;
+using Newtonsoft.Json.Schema.Generation;
+using Newtonsoft.Json.Serialization;
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: > Is there a similar solution via System.Text.Json?

Not without any external libraries:
https://github.com/dotnet/runtime/issues/29887

Modules/Challenges/test/SpecFlowChallenges.Specs/API/ChallengesApi.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System.Net;
+using RestSharp;
+using SpecFlowChallenges.Specs.Models;
+
+namespace SpecFlowChallenges.Specs.API;
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: > Why does the project have such a weird name?

It's the naming convention Specflow uses in their "Get started" tutorial (https://docs.specflow.org/projects/getting-started/en/latest/GettingStarted/Step3.html). But I haven't found any docs saying that we should actually follow it, so I'll update it to the one you suggested.

Modules/Challenges/test/SpecFlowChallenges.Specs/Models/AuthenticationParams.cs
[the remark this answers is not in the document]
tnotheis: I'm generally a fan of generic/dynamic code. And I even tend to drive it too far in that direction. That's why I keep asking myself "Is it really necessary to be THAT generic?". Most of the time the answer is "No". In this case (and many other cases as well), being generic comes with the cost of loosing readability. In this case, we also loose type safety, because a user of the AuthenticationParameters could by accident add a "userName" parameter instead of "username".

Also, always think about the questions "How often do I need to change the code?" and "What are the cost of changing the code?". 
In your case: 
* How often do you need to add a authentication parameter? Since OAuth is a widely used standard, you can probably expect there to be no changes at all. 
* What are the cost of changing the code? Well, you need to add a property to the class. If you declare it as `required`, the compiler tells you all the places in which it is missing. All in all, it's probably a matter of a few minutes.

Modules/Challenges/test/SpecFlowChallenges.Specs/StepDefinitions/ChallengesApiStepDefinitions.cs
KONKdiogo-ferraz: We're also using this step definition for tests that proposedly pass an invalid id. In my opinion, if we validate the challenge Id here, we will not check the server validation properly.

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/CreateChallenge.feature
```diff
@@ -0,0 +1,36 @@
...
+	Given the user is unauthenticated
+	When a POST request is sent to the Challenges endpoint
+	Then the response status code is 201
+	And the response contains a Challenge
+
+Scenario: Creating a Challenge as an authenticated user
+	Given the user is authenticated
+	When a POST request is sent to the Challenges endpoint
+	Then the response status code is 201
+	And the response contains a Challenge
```
tnotheis, line 16: We should add `And the Challenge contains information about the creator`, which asserts that createdBy and createadByDevice are not null.

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/CreateChallenge.feature
```diff
@@ -0,0 +1,36 @@
...
+
+Scenario: Creating a Challenge with a JSON sent in the request content
+	When a POST request is sent to the Challenges endpoint with
+		| Key  | Value                              |
+		| Body | {"this": "is some arbitrary json"} |
+	Then the response status code is 201
+	And the response contains a Challenge
```
tnotheis, line 23: We should add `And the Challenge does not contain information about the creator`, which asserts that createdBy and createadByDevice are null.

tnotheis (changes requested):

Modules/Challenges/test/Challenges.API.Tests.Integration/Models/ChallengeResponse.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Newtonsoft.Json;
+
+namespace Challenges.API.Tests.Integration.Models;
+public class ChallengeResponse
+{
+    public Challenge? Result { get; set; }
+    [JsonIgnore]
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: A success Challenge response doesn't include an Error property. If we don't ignore that, the json schema validation will fail.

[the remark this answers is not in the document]
tnotheis: Are you sure? Because the property is marked as nullable. Doesn't the JSON schema validator recognize that?
If not, there is probably an attribute that can be used to mark it as optional.

Modules/Challenges/test/Challenges.API.Tests.Integration/API/ChallengesApi.cs
```diff
@@ -0,0 +1,92 @@
...
+            ContentType = response.ContentType,
+            Content = response.Content
+        };
+
+        return result;
+    }
+
+    public async Task<AccessTokenResponse> GetAccessToken(AuthenticationParameters authenticationParams)
+    {
+        if (_accessTokenResponse is { Expired: false })
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: I initialize _accessTokenResponse as null, so I would need to do
if(_accessTokenResponse  != null && !_accessTokenResponse.Expired)
But if you prefer, I can make the change.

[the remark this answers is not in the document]
tnotheis: Oh, right. Well, you could do something like `_accessTokenResponse?.Expired != null`. But it's fine to leave it as it is.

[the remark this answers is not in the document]
tnotheis: Nvm, that doesn't work, of course.

Modules/Challenges/test/Challenges.API.Tests.Integration/Models/AuthenticationParameters.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Challenges.API.Tests.Integration.Models;
+public class AuthenticationParameters
+{
+    [Required]
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: Now that I think about it, 2/5 of the properties are validated when we validate the HttpConfiguration (ClientId, ClientSecret) and the remaining ones are hardcoded in the step definitions (GrantType, UserName, Password). So I don't think we need to add these fields as [Required] here. What do you think?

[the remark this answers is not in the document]
tnotheis: Yep, I think so too

Modules/Challenges/test/Challenges.API.Tests.Integration/Models/ChallengeResponse.cs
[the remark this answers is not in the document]
KONKdiogo-ferraz: There's `[JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]` . The attribute is used to indicate that the JSON property should be ignored if its value is the default value for the data type.

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/GetChallengeById.feature
```diff
@@ -0,0 +1,34 @@
...
+	Then the response status code is 401
+
+Scenario: Requesting a nonexistent Challenge as an authenticated user
+	Given the user is authenticated
+	When a GET request is sent to the Challenges/{id} endpoint with "CHLthisisnonexisting"
+	Then the response status code is 404
+
+Scenario: Requesting a Challenge with an invalid id as an authenticated user
+	Given the user is authenticated
+	When a GET request is sent to the Challenges/{id} endpoint with "a123"
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: I added a scenario outline to consider all those invalid id options:

`Scenario Outline: Requesting a Challenge with an invalid id as an authenticated user
	Given the user is authenticated
	When a GET request is sent to the Challenges/{id} endpoint with <id>
	Then the response status code is 400
	And the response content includes an error with the error code "error.platform.invalidId"

Examples:
	| id                          |
	| a123                        |
	| CHLthishastoomanycharacters |
	| CHLnotenoughchars           |
	| !CHLdfhuwnjdfnjnjfnd        |
	| PHLfdjfdjflndjkfndjk        |
	| CHL_frfssd_fdfdsed#_        |`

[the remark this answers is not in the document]
tnotheis: That's fine 👍

Modules/Challenges/test/Challenges.API.Tests.Integration/Models/RequestConfiguration.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Challenges.API.Tests.Integration.Models;
+public class RequestConfiguration
+{
+    public AuthenticationParameters AuthenticationParameters { get; set; } = new AuthenticationParameters();
+    public bool IsAuthenticated { get; set; } = false;
```
tnotheis, line 5: The property name is not ideal. A RequestConfiguration cannot be authenticated. Let's rather use something like "Authenticate". Or maybe an "AuthenticationMode" property of type "AuthenticationMode", which is an enum with the values "Authenticated" and "Anonymous".

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/CreateChallenge.feature
```diff
@@ -0,0 +1,37 @@
...
+Scenario: Creating a Challenge with a JSON sent in the request content
+	When a POST request is sent to the Challenges endpoint with
+		| Key  | Value                              |
+		| Body | {"this": "is some arbitrary json"} |
+	Then the response status code is 415
+
+Scenario: Creating a Challenge with an invalid JSON sent in the request content
+	When a POST request is sent to the Challenges endpoint with
+		| Key  | Value                             |
+		| Body | { "thisJSON": "has an extra }" }} |
```
tnotheis, line 29: We once said that we want to use the word "content" instead of "body". In the title of this scenario you already use it, but not in the table. Please search in the whole project for the word "Body" and replace it accordingly.

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/CreateChallenge.feature
```diff
@@ -0,0 +1,37 @@
+﻿@Integration
+Feature: Create a Challenge
+
+User creates a Challenge
+
+Scenario: Creating a Challenge as an anonymous user
+	Given the user is unauthenticated
+	When a POST request is sent to the Challenges endpoint
```
tnotheis, line 8: Shouldn't it be called `When he sends a POST request to the Challenges endpoint`?

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/CreateChallenge.feature
```diff
@@ -0,0 +1,37 @@
...
+	Then the response status code is 201
+	And the response contains a Challenge
+	And the Challenge contains information about the creator
+
+Scenario: Creating a Challenge with a JSON sent in the request content
+	When a POST request is sent to the Challenges endpoint with
+		| Key  | Value                              |
+		| Body | {"this": "is some arbitrary json"} |
+	Then the response status code is 415
```
tnotheis, line 24: Maybe you should add the meaning of the status code, so that the reader doesn't have to look it up.

Example:
```suggestion
	Then the response status code is 415 (Unsupported Media Type)
```

The meaning of course doesn't have to be validated. It's just for information purposes.

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/GetChallengeById.feature
```diff
@@ -0,0 +1,42 @@
...
+	And the Accept header is 'application/xml'
+	When a GET request is sent to the Challenges/{id} endpoint with a valid Id
+	Then the response status code is 406
+
+Scenario Outline: Requesting a Challenge with an invalid id as an authenticated user
+	Given the user is authenticated
+	When a GET request is sent to the Challenges/{id} endpoint with <id>
+	Then the response status code is 400
+	And the response content includes an error with the error code "error.platform.invalidId"
+
```
tnotheis, line 35: Is the blank line required? If not I would say you should remove it, to show that the two paragraphs belong together

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/GetChallengeById.feature
```diff
@@ -0,0 +1,42 @@
...
+Examples:
+	| id                          | description                 |
+	| CHLthishastoomanycharacters | More than 20 characters     |
+	| CHLnotenoughchars           | Less than 20 characters     |
+	| !CHLdfhuwnjdfnjnjfnd        | Contains invalid characters |
+	| PHLfdjfdjflndjkfndjk        | Does not have CHL prefix    |
+	| CHL_frfssd_fdfdsed#_        | Contains invalid characters |
```
tnotheis, line 42: Did you add two lines with the same description on purpose? If so, can you put them next to each other please?

Modules/Challenges/test/Challenges.API.Tests.Integration/Models/AuthenticationParameters.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Challenges.API.Tests.Integration.Models;
+public class AuthenticationParameters
+{
+    public string GrantType { get; set; }
+    public string ClientId { get; set; }
+    public string ClientSecret { get; set; }
+    public string UserName { get; set; }
```
tnotheis, line 7:
```suggestion
    public string Username { get; set; }
```

Modules/Challenges/test/Challenges.API.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    [Given(@"a Challenge c")]
+    public async Task GivenAChallengeC()
+    {
+        var challengeResponse = await _challengeApi.CreateChallenge(_requestConfiguration);
+        challengeResponse.IsSuccessStatusCode.Should().BeTrue();
+
+        _challengeId = challengeResponse.Data!.Result!.Id;
+        _challengeId.Should().NotBeNullOrEmpty(because: "Required value for 'Id' is missing.");
```
tnotheis, line 61: I would not assert this here. That's not the responsibility of this step.

Modules/Challenges/test/Challenges.API.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    [When(@"a POST request is sent to the Challenges endpoint with")]
+    public async Task WhenAPOSTRequestIsSentToTheChallengesEndpointWith(Table table)
+    {
+        var requestConfiguration = table.CreateInstance<RequestConfiguration>();
```
tnotheis, line 67: Doesn't the requestConfiguration from the table have to be merged with `_requestConfiguration`?

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/GetChallengeById.feature
```diff
@@ -0,0 +1,42 @@
+﻿@Integration
+Feature: Get Challenge By Id
+
+User requests a Challenge
+
+Scenario: Requesting a Challenge as an authenticated user
+	Given the user is authenticated
+	And a Challenge c
+	When a GET request is sent to the Challenges/{id} endpoint with "c.Id"
```
tnotheis, line 9: Didn't we say that we want to remove the quotes around c.Id?

Modules/Challenges/test/Challenges.API.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,157 @@
...
+                id = _challengeId!;
+                break;
+            case "a valid Id":
+                id = "CHLjVPS6h1082AuBVBaR";
+                break;
+        }
+        _challengeResponse = await _challengeApi.GetChallengeById(_requestConfiguration, id);
+    }
+
+    [Then(@"the response content includes an error with the error code ""([^""]*)""")]
```
tnotheis, line 92:
```suggestion
    [Then(@"the response content includes an error with the error code ""([^""]+)""")]
```

Modules/Challenges/test/Challenges.API.Tests.Integration/Support/Dependencies.cs
```diff
@@ -0,0 +1,35 @@
...
+public static class Dependencies
+{
+    private const string APP_SETTINGS_FILE = "appsettings.json";
+
+    [ScenarioDependencies]
+    public static IServiceCollection CreateServices()
+    {
+        var services = new ServiceCollection();
+
+        IConfiguration config = new ConfigurationBuilder()
```
tnotheis, line 19: Can't you use `var` here?

Modules/Challenges/test/Challenges.API.Tests.Integration/Support/Dependencies.cs
```diff
@@ -0,0 +1,35 @@
...
+            .AddJsonFile(Path.Combine(Directory.GetCurrentDirectory(), APP_SETTINGS_FILE), optional: true, reloadOnChange: true)
```
tnotheis, line 20: It's not necessary to reload the settings on change. But the file is required, isn't it?
```suggestion
            .AddJsonFile(Path.Combine(Directory.GetCurrentDirectory(), APP_SETTINGS_FILE), optional: false, reloadOnChange: false)
```

tnotheis (changes requested):

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/CreateChallenge.feature
KONKdiogo-ferraz: I followed the examples they have in this specflow blog post:
https://specflow.org/challenges/test-apis-with-gherkin/

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/GetChallengeById.feature
KONKdiogo-ferraz: It's not required. However they include the blank line in the examples:
https://cucumber.io/docs/gherkin/reference/#scenario-outline

Modules/Challenges/test/Challenges.API.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
KONKdiogo-ferraz: Given the RequestConfiguration class:
```c#
public class RequestConfiguration
{
    public AuthenticationParameters AuthenticationParameters { get; set; } = new AuthenticationParameters();
    public bool Authenticate { get; set; } = false;
    public string ContentType { get; set; } = string.Empty;
    public string AcceptHeader { get; set; } = string.Empty;
    public string Content { get; set; } = string.Empty;
}
```
If the authentication is not relevant for the test (in this test we don't specify if the user is anonymous or authenticated), we don't really need to merge requestConfiguration from the table with _requestConfiguration. All the relevant properties are filled in the scenario table.

tnotheis: But you don't know what other properties of the RequestConfiguration defaults are created for in the future. So I think it would be better/safer to just merge them

Modules/Challenges/test/Challenges.API.Tests.Integration/Features/GetChallengeById.feature
tnotheis: I think I would still remove the blank line. It looks confusing, I think. As if the examples belonged to each of the scenarios

Modules/Challenges/test/Challenges.API.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
tnotheis: Alright. I think we even discussed this some days ago, and you already convinced me back then. Sorry for repeating this. :)

Modules/Challenges/test/Challenges.API.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -65,7 +68,9 @@ public async Task GivenAChallengeC()
     public async Task WhenAPOSTRequestIsSentToTheChallengesEndpointWith(Table table)
     {
         var requestConfiguration = table.CreateInstance<RequestConfiguration>();
-        _challengeResponse = await _challengeApi.CreateChallenge(requestConfiguration);
+        var mergedRequestConfiguration = _mapper.Map(_requestConfiguration, requestConfiguration);
```
tnotheis, line 71: I appreciate your effort to do some research on this topic. And in most cases this would probably be the best solution.
But in this case, the merge is pretty simple and contains only a few properties. Introducing AutoMapper for this sole purpose is a bit overkill. That's why I would have used a simple method on the RequestConfiguration:

```c#
public void SupplementWith(RequestConfiguration other) 
{
    this.Authenticate ?= other.Authenticate; 
    this.AcceptHeader ?= other.AcceptHeader;
    this.ContentTypeHeader ?= other.ContentTypeHeader;
    this.Content ?= other.Content;
}
``` 
You need to make all the properties nullable to be able to do this.

And to call it:

```c#
requestConfiguration.SupplementWith(_requestConfiguration);
```

tnotheis (changes requested):

Modules/Challenges/test/Challenges.API.Tests.Integration/Models/RequestConfiguration.cs
```diff
@@ -1,21 +1,16 @@
...
+    public void SupplementWith(RequestConfiguration other)
     {
-        configuration
-            .CreateMap<RequestConfiguration, RequestConfiguration>()
-            .ForMember(dest => dest.ContentType, opt => opt.Condition((src, dest) => string.IsNullOrEmpty(dest.ContentType)))
-            .ForMember(dest => dest.AcceptHeader, opt => opt.Condition((src, dest) => string.IsNullOrEmpty(dest.AcceptHeader)))
-            .ForMember(dest => dest.Content, opt => opt.Condition((src, dest) => string.IsNullOrEmpty(dest.Content)));
+        AcceptHeader = string.IsNullOrEmpty(AcceptHeader) ? other.AcceptHeader : AcceptHeader;
+        ContentType = string.IsNullOrEmpty(ContentType) ? other.ContentType : ContentType;
+        Content = string.IsNullOrEmpty(Content) ? other.Content : Content;
```
tnotheis, line 14: That's not good. If you now set an AcceptHeader to, let's say "application/json" in the defaults, and in a test you say "Given an empty Accept header", then this results in an Accept header of "application/json", and you wouldn't even notice it in the tests. That's why I used null in my comment.

tnotheis (changes requested):

Modules/Challenges/test/Challenges.API.Tests.Integration/Models/RequestConfiguration.cs
KONKdiogo-ferraz: Good call!

tnotheis (approved):

PR #28, opened 2023-03-07
Feature/database health checks support for postgresql

daniel-almeida-konkconsulting:


Backbone.API/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -114,6 +137,14 @@ public static class IServiceCollectionExtensions
         return services;
     }
 
+    public static object GetPropertyValue(object source, string propertyName)
+    {
+        foreach (var property in propertyName.Split('.').Select(s => source.GetType().GetProperty(s)))
+            source = property.GetValue(source, null);
```
tnotheis, line 143: Isn't this the same as 

```cs
var source = propertyName.Split('.').Select(n => source.GetType().GetProperty(n)).Select(p => p.GetValue(source, null))
```
?

A little bit of formatting to make it more readable:

```cs
var source = propertyName
                    .Split('.')
                    .Select(n => source.GetType().GetProperty(n))
                    .Select(p => p.GetValue(source, null))
```

And you should probably rename `propertyName` to `propertyPath`

Backbone.API/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -102,10 +102,33 @@ public static class IServiceCollectionExtensions
...
-            // TODO: M use separate Connection String
-            .AddSqlServer(configuration.Modules.Challenges.Infrastructure.SqlDatabase.ConnectionString);
+        var modules = configuration.Modules.GetType().GetProperties();
+        foreach (var moduleProperty in modules)
+        {
+            if (moduleProperty is null) continue;
+
+            var moduleName = moduleProperty.Name;
+            var module = configuration.Modules.GetType().GetProperty(moduleName).GetValue(configuration.Modules, null);
+            var provider = GetPropertyValue(module, "Infrastructure.SqlDatabase.Provider");
```
tnotheis, line 112: I would add a cast to string here as well, so it's clear what is inside

tnotheis (changes requested):

Backbone.API/Extensions/IServiceCollectionExtensions.cs
tnotheis: Nvm, that doesn't make any sense. ;-)

tnotheis (approved):

PR #31, opened 2023-03-13
Test(messages)/add performance tests to messages

KonkLuisCosta:
-performance tests added to messages service

### DIFF
--- a/Modules/Messages/test/Messages.API.Tests.Performance.EndToEnd/src/utils.ts
+++ b/Modules/Messages/test/Messages.API.Tests.Performance.EndToEnd/src/utils.ts
@@ -58,7 +58,7 @@ export function getConfiguration(): Configuration {
   };
 }
 
-function assertEnvVarExists(parameter: string) {
+function assertEnvVarExists(parameter: string): void {
   if (!__ENV[parameter]) {
     exec.test.abort(`Parameter '${parameter}' cannot be null or empty`);
   }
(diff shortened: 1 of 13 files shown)


Modules/Messages/test/Messages.API.Tests.Performance.EndToEnd/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  return {
+    host: simplifyHost(__ENV.HOST),
+    clientSecret: __ENV.CLIENT_SECRET,
+    user: __ENV.USER,
+    password: __ENV.PASSWORD,
+    size: __ENV.SIZE,
+  };
+}
+
+function assertEnvVarExists(parameter: string) {
```
daniel-almeida-konkconsulting, line 61: Please specify the return type with ": void"

Modules/Messages/test/Messages.API.Tests.Performance.EndToEnd/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
tnotheis, line 3: es5 is from 2009, so it's very old. Is it possible to use the latest version (es2022)?

tnotheis (changes requested):

PR #37, opened 2023-03-20
Test(relationships)/performance tests

KonkLuisCosta:


tnotheis:
It's not necessary for such small changes to create separate PRs for each of them. To avoid this we have introduced a monorepo.

Further: the names of the PRs do not describe what's in them. The goal is not to describe then part that was worked on, but rather what was done. So instead of "Test(relationships)/performance tests" the title should be "Test(relationships)/change ES version of performance tests".

If we combine the two suggestions, the result is a single PR with the title "Test/change ES version of performance tests".

Modules/Relationships/test/RelationshipTemplates.API.Tests.Performance.EndToEnd/external.module.declarations.ts
```diff
@@ -1 +0,0 @@
-declare module "https://jslib.k6.io/k6chaijs/4.3.4.2/index.js";
```
tnotheis, line 1: Just noticed that you removed this line. Why?

tnotheis:
Use a single PR for the changes of the ES version

PR #46, opened 2023-03-24
Admin API: List Identities

Dannyps:
Adds Identities list repo, mediatr and controller, as well as integration and unit tests for the controller and mediatr handler, respectively.


tnotheis (changes requested):
Why did you put the code into Backbone.API? It should be in a new `AdminAPI` project (hence the two bullet points "Create the project" and "Add it to docker-compose.yml" in the Jira task description).

Dannyps:
@tnotheis I mentally split the two tickets I had assigned and did not understand that this functionality should be part of the new Admin API. @daniel-almeida-konkconsulting has explained this to me.

Considering the creation of the project for the Admin API itself is part of another ticket, should I create another branch for the creation of the project, or can I do it on this one?

Also, where would you like for the new project to be located?
Thanks

tnotheis:
Oh, I see. Originally the only task regarding the Admin API in the current sprint was the one for creating Tiers. We then spontaneously decided to add the task for listing Identities. That's why it doesn't contain the step for creating the project. Sorry for the confusion.
You can create the AdminAPI project as part of this task. And just place into the root of the repository:
```
bkb-backbone/
├─ AdminAPI/
│  ├─ BackboneAPI.csproj
```

Dannyps:
The project name should be ~`BackboneAPI.csproj`~ `AdminAPI.csproj`, right?

tnotheis:
Oops. Yes, of course

Timo Notheisen
Senior Developer
+49 (6221) 4312127
***@***.***
j&s-soft GmbH
Max-Jarecki-Straße 21
69115Heidelberg
Germany
js-soft.com
Sitz der Gesellschaft/Registered Office: 
​Max-Jarecki-Straße 21, D-69115 Heidelberg
 Geschäftsführer: Dirk Jennerjahn
 Handelsregister Amtsgericht Mannheim HRB 337601

​​Diese E-Mail kann Betriebs- oder Geschäftsgeheimnisse oder sonstige vertrauliche Informationen enthalten. Sollten Sie diese E-Mail irrtümlich erhalten haben, ist Ihnen eine Kenntnisnahme des Inhalts, eine Vervielfältigung oder Weitergabe der E-Mail ausdrücklich untersagt. Bitte benachrichtigen Sie uns und vernichten Sie die empfangene E-Mail. Vielen Dank.
 
 This e-mail may contain trade secrets or privileged, undisclosed, or otherwise confidential information. If you have received this e-mail in error, you are hereby notified that any review, copying, or distribution of it is strictly prohibited. Please inform us immediately and destroy the original transmittal. Thank you for your cooperation.

tnotheis (changes requested):
The Admin API is to be unprotected for the time being. In the medium term, we will then introduce authentication via API key. In the long term, there will probably be something more sophisticated, but at least not in the first few months. So the authentication code is redundant until then.

Dannyps:
@tnotheis I did start without the Authentication part, as per the issue description. However, I came across difficulties when calling mediatr handlers due to the need of a `IUserContext`.

I have since changed the way the handler works and trimmed the authentication part.

tnotheis:
@Dannyps There shouldn't be any problem if you don't execute a command/query that actually uses the `IUserContext`. If you still observe this behavior, don't hesitate to call me.

Admin.API/Controllers/AuthorizationController.cs
tnotheis, line 1: You can remove this class

Admin.API/Admin.API.csproj
```diff
@@ -0,0 +1,36 @@
...
+    <PackageReference Include="Serilog.Enrichers.Environment" Version="2.2.0" />
+    <PackageReference Include="Serilog.Sinks.ApplicationInsights" Version="4.0.0" />
+    <PackageReference Include="Serilog.Sinks.Http" Version="8.0.0" />
+    <PackageReference Include="AspNetCore.HealthChecks.NpgSql" Version="6.0.2" />
+    <PackageReference Include="AspNetCore.HealthChecks.SqlServer" Version="6.0.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Backbone.Infrastructure\Backbone.Infrastructure.csproj" />
+    <ProjectReference Include="..\BuildingBlocks\src\BuildingBlocks.API\BuildingBlocks.API.csproj" />
```
tnotheis, line 31: You should remove this reference. The Admin API should be completely separated from the Consumer API

Modules/Devices/test/Devices.API.Tests.Integration/Models/RequestConfiguration.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Devices.API.Tests.Integration.Models;
+public class RequestConfiguration
+{
+    public bool Authenticate { get; set; } = false;
```
tnotheis, line 4: You can remove this

Modules/Devices/test/Devices.API.Tests.Integration/StepDefinitions/GetIdentityListStepDefinitions.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.Extensions.Options;
+    
+using static Challenges.API.Tests.Integration.Configuration.Settings;
+
+namespace Devices.API.Tests.Integration.StepDefinitions
+{
+    [Binding]
+    public class GetIdentityListStepDefinitions
+    {
+
```
tnotheis, line 12: Redundant blank line

tnotheis (changes requested):
I stop reviewing because of the reference of the Backbone.API project. Depending on what you're using it for, it's possible that you need to change a lot. I will continue after you removed the reference.

Modules/Devices/test/Devices.API.Tests.Integration/API/IdentitiesApi.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        _client = client;
+
+        ServicePointManager.ServerCertificateValidationCallback +=
+                (sender, cert, chain, sslPolicyErrors) => true;
+    }
+
+    public async Task<HttpResponse<List<IdentityDTO>>> GetIdentitiesList(RequestConfiguration requestConfiguration)
+    {
+        return await ExecuteIdentitiesRequest<List<IdentityDTO>>(Method.Get, new PathString(ROUTE_PREFIX).Add($"/identities").ToString(), requestConfiguration);
```
tnotheis, line 22:
```suggestion
        return await ExecuteIdentitiesRequest<List<IdentityDTO>>(Method.Get, new PathString(ROUTE_PREFIX).Add($"/Identities").ToString(), requestConfiguration);
```

Modules/Devices/test/Devices.API.Tests.Integration/Features/IdentityList.feature
```diff
@@ -0,0 +1,9 @@
+﻿@Integration
+Feature: Get Identity List
+
+User requests an Identity List
+
+Scenario: Requesting an Identity List as an authenticated user
```
tnotheis, line 6:
```suggestion
Scenario: Requesting an Identity List
```

Modules/Devices/test/Devices.API.Tests.Integration/Features/IdentityList.feature
```diff
@@ -0,0 +1,9 @@
...
+	When a GET request is sent to the Identities/ endpoint
+	Then the response status code is 200 (OK)
+	And the response contains a list
```
tnotheis, line 9:
```suggestion
	And the response contains a paginated list of Identities
```

Modules/Devices/test/Devices.API.Tests.Integration/API/IdentitiesApi.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Net;
+using Devices.API.Tests.Integration.Models;
+using Microsoft.AspNetCore.Http;
+using RestSharp;
+
+namespace Challenges.API.Tests.Integration.API;
```
tnotheis, line 6: Wrong namespace. There are some other files in this project which have the wrong namespace as well. Please check all of them.

Modules/Devices/test/Devices.API.Tests.Integration/StepDefinitions/GetIdentityListStepDefinitions.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Devices.API.Tests.Integration.StepDefinitions
+{
+    [Binding]
+    public class GetIdentityListStepDefinitions
+    {
+        private readonly RequestConfiguration _requestConfiguration;
+        private readonly IdentitiesApi _identitiesApi;
+        private HttpResponse<List<IdentityDTO>>? _identitiesResponse;
+
+        public GetIdentityListStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, IdentitiesApi identitiesApi)
```
tnotheis, line 16: No need to inject `httpConfiguration`. It's not used anyway, as far as I can see.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/ListIdentities/HandlerTests.cs
```diff
@@ -0,0 +1,104 @@
...
+    public HandlerTests()
+    {
+        AssertionScope.Current.FormattingOptions.MaxLines = 1000;
+
+        (_arrangeContext, _, _actContext) = FakeDbContextFactory.CreateDbContexts<DevicesDbContext>();
+        _handler = CreateHandler();
+    }
+
+    [Fact]
+    public async void Returns_empty_list_when_no_loaded_entities()
```
tnotheis, line 35: What do you mean with "loaded entities"? Wouldn't it be better like `Returns_an_empty_list_when_no_identities_exist`?

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/ListIdentities/HandlerTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        // Act
+        var result = await _handler.Handle(new ListIdentitiesQuery(request), CancellationToken.None);
+
+        // Assert
+        result.Identities.Should().HaveCount(0);
+    }
+
+    [Fact]
+    public async void Returns_valid_list_when_entities_loaded()
```
tnotheis, line 50: What is a "**valid** list"?

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/ListIdentities/HandlerTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        // Act
+        var result = await _handler.Handle(new ListIdentitiesQuery(request), CancellationToken.None);
+
+        // Assert
+        result.Identities.Should().HaveCount(2);
+    }
+
+    [Fact]
+    public async void Returns_valid_list_when_entities_loaded_and_attributes_match()
```
tnotheis, line 76: > Returns_valid_list_when_entities_loaded_and_attributes_match

So this means that the list is invalid when the attributes don't match? ;)

tnotheis (changes requested):

Admin.API/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Admin.API.Configuration;
+using Microsoft.OpenApi.Models;
+
+namespace Admin.API.Extensions;
+
+public static class IServiceCollectionExtensions
+{
+    public static IServiceCollection AddSwaggerWithCustomUi(this IServiceCollection services,
```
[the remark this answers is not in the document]
Dannyps: This was bugging me because it implies that we're adding a custom Swagger Ui, but the truth is that if you don't call this, Swagger is not generated/made available. I figured this new name made more sense, but I'll revert it.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/ListIdentities/HandlerTests.cs
Dannyps: I suppose by valid list I meant simply "a list"

Admin.API/Extensions/IServiceCollectionExtensions.cs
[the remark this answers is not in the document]
tnotheis: Okay, I get you. Never thought about it his way.

The reason for this naming is a naming convention I introduced. There are multiple methods like this, for example AddCustomIdentity, AddCustomAspNetCore, ...
The problem is that most of the time, there are methods like AddIdentity or AddAspNetCore. In order to avoid naming conflicts, I added the `Custom` prefix. 
Do you have a better idea for it? It would be great if the naming convention would be applicable to all cases.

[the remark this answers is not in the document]
Dannyps: Out of the top of my head, I'd say `AddCustomSwaggerWithUi`.
I think the Custom keyword makes sense, so this is a way of keeping it.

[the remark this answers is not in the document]
tnotheis: Okay, I aggree

Modules/Devices/src/Devices.Application/Infrastructure/Persistence/Repository/IIdentityRepository.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Backbone.Modules.Devices.Domain.Entities;
+using Enmeshed.BuildingBlocks.Application.Abstractions.Infrastructure.Persistence.Database;
+using Enmeshed.BuildingBlocks.Application.Pagination;
+
+namespace Backbone.Modules.Devices.Application.Infrastructure.Persistence.Repository;
+
+public interface IIdentityRepository
```
tnotheis, line 7: I wanted to name the repositories in plural (=> IIdentitiesRepository)

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/IdentityRepository.cs
```diff
@@ -0,0 +1,31 @@
...
+using Enmeshed.BuildingBlocks.Application.Abstractions.Infrastructure.Persistence.Database;
+using Enmeshed.BuildingBlocks.Application.Extensions;
+using Enmeshed.BuildingBlocks.Application.Pagination;
+using Microsoft.EntityFrameworkCore;
+
+namespace Backbone.Modules.Devices.Infrastructure.Persistence.Repository;
+public class IdentityRepository : IIdentityRepository
+{
+    private readonly DbSet<Identity> _identityDbSet;
+    private readonly IQueryable<Identity> _readonlyIdentityDbSet;
```
tnotheis, line 18: I usually like expressive naming, but in this case, I think we can keep it simple:

```suggestion
    private readonly DbSet<Identity> _identities;
    private readonly IQueryable<Identity> _readonlyIdentities;
```

Modules/Devices/test/Devices.API.Tests.Integration/Models/IdentityResponse.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Devices.API.Tests.Integration.Models;
+using Newtonsoft.Json;
+
+namespace Challenges.API.Tests.Integration.Models;
+public class IdentityResponse
```
tnotheis, line 5:
```suggestion
public class ListIdentitiesResponse
```

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/IdentityRepository.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    public IdentityRepository(DevicesDbContext dbContext)
+    {
+        _identityDbSet = dbContext.Identities;
+        _readonlyIdentityDbSet = dbContext.Identities.AsNoTracking();
```
tnotheis, line 23: You didn't test this, did you? Because this code should throw a NullReferenceException when accessing complex child properties, like "Devices`. You have to tell EF Core to actually read them from the database.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/IdentityRepository.cs
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public async Task<DbPaginationResult<Identity>> FindAll(PaginationFilter paginationFilter)
+    {
+        var paginationResult = await _readonlyIdentityDbSet.OrderAndPaginate(d => d.CreatedAt, paginationFilter);
```
tnotheis, line 28: Include complex child properties

tnotheis (commented):
Can you please execute the GET /Identities endpoint at least once, so we know that everything is set up correctly?

Modules/Devices/test/Devices.API.Tests.Integration/Models/IdentityResponse.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Devices.API.Tests.Integration.Models;
+using Newtonsoft.Json;
+
+namespace Challenges.API.Tests.Integration.Models;
```
tnotheis, line 4: Namespace... Please execute a find&replace on the whole project to find all occurences.

Modules/Devices/test/Devices.API.Tests.Integration/StepDefinitions/GetIdentityListStepDefinitions.cs
```diff
@@ -0,0 +1,45 @@
...
+        [When(@"a GET request is sent to the Identities/ endpoint")]
+        public async Task WhenAGETRequestIsSentToTheIdentitiesEndpointAsync()
+        {
+            _identitiesResponse = await _identitiesApi.GetIdentitiesList(_requestConfiguration);
+            _identitiesResponse.Should().NotBeNull();
+            _identitiesResponse!.Data.Should().NotBeNull();
+            _identitiesList = _identitiesResponse!.Data!.Result;
+        }
+
+        [Then(@"the response status code is (\d*) \((?:[a-z]|[A-Z]|\s)+\)")]
```
tnotheis, line 31: Could you change the regex as follows (for the other integration tests as well please)
```suggestion
        [Then(@"the response status code is (\d\d\d) \((?:[a-z]|[A-Z]|\s)+\)")]
```

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/ListIdentities/HandlerTests.cs
```diff
@@ -0,0 +1,105 @@
+﻿using FluentAssertions;
+using FluentAssertions.Execution;
+using Xunit;
+using Backbone.Modules.Devices.Application.Tests;
+using Backbone.Modules.Devices.Domain.Entities;
+using Backbone.Modules.Devices.Application.Identities.Queries.ListIdentities;
+using Handler = Backbone.Modules.Devices.Application.Identities.Queries.ListIdentities.Handler;
+using Enmeshed.BuildingBlocks.Application.Pagination;
+using Backbone.Modules.Devices.Application.Infrastructure.Persistence.Repository;
+
```
tnotheis, line 10: unnecessary blank line

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/ListIdentities/HandlerTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        // Act
+        var result = await _handler.Handle(new ListIdentitiesQuery(request), CancellationToken.None);
+
+        // Assert
+        result.Should().HaveCount(0);
+    }
+
+    [Fact]
+    public async void Returns_list_when_identities_loaded()
```
tnotheis, line 46:
```suggestion
    public async void Returns_a_list_of_all_existing_identities()
```

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/ListIdentities/HandlerTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        // Act
+        var result = await _handler.Handle(new ListIdentitiesQuery(request), CancellationToken.None);
+
+        // Assert
+        result.Should().HaveCount(2);
+    }
+
+    [Fact]
+    public async void Returns_list_when_identities_loaded_and_attributes_match()
```
tnotheis, line 72:
```suggestion
    public async void Returned_identities_have_all_properties_filled_as_expected()
```

tnotheis (changes requested):

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/IdentityRepository.cs
Dannyps: Well spotted. No NullReferenceException, but the counts were returning zero.

Dannyps:
I've been using swagger to test the `/api/v1/Identities` endpoint. Here's an example of the current output:
```json
{
  "result": [
    {
      "address": "id12Pbi7CgBHaFxge6uy1h6qUkedjQr8XHfm",
      "clientId": "test",
      "publicKey": "AQEBAQE=",
      "createdAt": "2023-03-27T09:58:52.6827474Z",
      "identityVersion": 1,
      "numberOfDevices": 1,
      "devices": [
        {
          "id": {
            "stringValue": "DVChfltTHVnATVrF5J6j"
          },
          "createdAt": "2023-03-27T09:58:52.684189Z",
          "createdByDevice": {
            "stringValue": "DVChfltTHVnATVrF5J6j"
          },
          "deletedAt": null,
          "deletedByDevice": null,
          "deletionCertificate": null,
          "lastLogin": {
            "time": null
          }
        }
      ]
    },
    {
      "address": "id1CsU6VwBMgBA4TLc61nj895ZD1mPjbddNW",
      "clientId": "test",
      "publicKey": "AgICAgI=",
      "createdAt": "2023-03-27T09:58:52.9513043Z",
      "identityVersion": 1,
      "numberOfDevices": 1,
      "devices": [
        {
          "id": {
            "stringValue": "DVCGd3QcUYyiJGb94nkt"
          },
          "createdAt": "2023-03-27T09:58:52.9513533Z",
          "createdByDevice": {
            "stringValue": "DVCGd3QcUYyiJGb94nkt"
          },
          "deletedAt": null,
          "deletedByDevice": null,
          "deletionCertificate": null,
          "lastLogin": {
            "time": null
          }
        }
      ]
    }
  ]
}
```

Admin.API/Program.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    services
+        .ConfigureAndValidate<AdminConfiguration>(configuration.Bind)
+        .ConfigureAndValidate<ApplicationOptions>(options => configuration.GetSection("Modules:Devices:Application").Bind(options));
+
+#pragma warning disable ASP0000 We retrieve the Configuration via IOptions here so that it is validated
+    var parsedConfiguration = services.BuildServiceProvider().GetRequiredService<IOptions<AdminConfiguration>>().Value;
+#pragma warning restore ASP0000
+
+    services.AddCustomSwaggerWithUi();
```
tnotheis, line 47: Just noticed that you are missing the following custom extension methods from Backbone.API:
* AddCustomFluentValidation
* AddCustomAspNetCore
* 
Is there a reason why you didn't add these?

Dannyps: I added those initially, but I tried to strip down as much as I could whithout breaking the API functionality.
Should I include those two calls?

tnotheis:
@Dannyps 
> I've been using swagger to test the /api/v1/Identities endpoint. Here's an example of the current output:

There is a problem. There are no json converters registered. This leads to the following:

```c#
     "devices": [
        {
          "id": {
            "stringValue": "DVCGd3QcUYyiJGb94nkt"
          }
          ...
       }
```
instead of
```c#
     "devices": [
        {
          "id": "DVCGd3QcUYyiJGb94nkt"
          ...
       }
```

The problem is that the value converters are defined in the Backbone.API project. I would suggest that you copy all of the value converters from there for now. Later on we will probably define a BuildingBlocks.API project where we can extract all common code between the two API projects.

Admin.API/Program.cs
tnotheis: Yep. The first one is for proper validation and the second one sets up AspNetCore in the "correct" way (you should remove the calls to AddAuthentication and AddAuthorization from the copied code, of course)

Dannyps:
I have added the Common converters. Should I add the other converters as well? They don't seem to be needed for now.

tnotheis:
> I have added the Common converters. Should I add the other converters as well? They don't seem to be needed for now.

No, it's fine. Thank you.

tnotheis:
Can you execute the route one last time to see if the value converters are working?

Dannyps:
I have confirmed that they are.

Admin.API/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -1,10 +1,134 @@
...
+                default:
+                    throw new Exception($"Unsupported database provider: {provider}");
+            }
+        }
+
+        services.AddHttpContextAccessor();
+
+        return services;
+    }
+    public static object GetPropertyValue(object source, string propertyPath)
```
tnotheis, line 124: Can you make this method private please? I already did it in Backbone.API

tnotheis (changes requested):
Just one tiny last thing :)

tnotheis (approved):

PR #47, opened 2023-03-27
Feature(devices)/create tier--tier aggregate

daniel-almeida-konkconsulting:


Modules/Devices/src/Devices.Domain/Aggregates/Tier/TierId.cs
```diff
@@ -0,0 +1,32 @@
...
+        var value = PREFIX + StringUtils.Generate(DefaultValidChars, DEFAULT_MAX_LENGTH_WITHOUT_PREFIX);
+        return new TierId(value);
+    }
+
+    public static Result<TierId, DomainError> Create(string value)
+    {
+        var validationResult = UTILS.Validate(value);
+        if (validationResult != null)
+            return Result.Failure<TierId, DomainError>(validationResult);
+        return Result.Success<TierId, DomainError>(new TierId(value));
```
tnotheis, line 29: Aren't there methods for "chaining" results in the CSharpFunctionalExtensions library? If so, can you try to use them please? I'm not sure if I'm gonna want to actually use them, but I would really like to see them in action. And since you said that you know them, it's probably no problem for you(?)

Modules/Devices/src/Devices.Infrastructure/Persistence/Database/ValueConverters/TierIdValueConverter.cs
tnotheis, line 1: The naming convention I used for the other value converters was `TierIdEntityFrameworkValueConverter`. 
But actually you didn't even need to add it in this PR.. Because this PR is only for the Aggregate related stuff, and the value converter is persistence.

Modules/Devices/test/Devices.Domain.Tests/TierNameTests.cs
```diff
@@ -0,0 +1,35 @@
...
+using Xunit;
+
+namespace Devices.Domain.Tests;
+public class TierNameTests
+{
+    [Fact]
+    public void Can_create_tier_name_with_valid_value()
+    {
+        var validLength = 30;
+        var validTierName = TestDataGenerator.GenerateString(validLength);
```
tnotheis, line 12: What I don't like about this approach is that you have no idea what the string looks like. So maybe we should use a hardcoded value here. In fact, even multiple values:
- "a-tier-name"
- "A Tier Name"

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Domain/StronglyTypedIds/Records/StronglyTypedIdHelpers.cs
```diff
@@ -20,11 +20,11 @@ public StronglyTypedIdHelpers(string prefix, char[] validChars, int maxLength)
     {
         var hasPrefix = stringValue.StartsWith(_prefix);
         if (!hasPrefix)
-            return GenericDomainErrors.InvalidIdPrefix($"Id starts with {stringValue.Take(3)} instead of {_prefix}");
+            return GenericDomainErrors.InvalidIdPrefix($"Id starts with {new string(stringValue.Take(3).ToArray())} instead of {_prefix}");
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: Without the change, the message would not print the actual values, but just the variable type.
With the change, it now correctly prints the prefix (e.g., TIR)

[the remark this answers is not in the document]
tnotheis: Good catch as well. :) I wonder why this was never noticed. :D

tnotheis (approved):

PR #59, opened 2023-04-03
Feature/create-tier--create-tier-command

daniel-almeida-konkconsulting:


Modules/Devices/src/Devices.Domain/Aggregates/Tier/TierName.cs
```diff
@@ -7,6 +7,7 @@ public record TierName
 {
     public string Value { get; }
     public const int MAX_LENGTH = 30;
+    public const int MINIMUM_LENGTH = 3;
```
tnotheis, line 10: Two minor things:

1. Can you move MINIMUM_LENGTH above MAX_LENGTH please? 
2. And please use either MIN/MAX or MINIMUM/MAXIMUM, but not a combination of both. :)

Modules/Devices/src/Devices.Domain/Aggregates/Tier/TierName.cs
```diff
@@ -23,9 +24,14 @@ private TierName(string value)
 
     public static DomainError? Validate(string value)
     {
-        if (value.Length > MAX_LENGTH)
-            return DomainErrors.InvalidTierName($"Tier Name is longer than the {MAX_LENGTH} characters");
+        if (value.Length > MAX_LENGTH || value.Length < MINIMUM_LENGTH)
+            return DomainErrors.InvalidTierName($"Tier Name length must be between {MINIMUM_LENGTH} and {MAX_LENGTH}");
```
tnotheis, line 28:
```suggestion
            return DomainErrors.InvalidTierName($"Tier Name length must have a length between {MINIMUM_LENGTH} and {MAX_LENGTH}");
```

tnotheis (changes requested):

Modules/Devices/test/Devices.Domain.Tests/TierNameTests.cs
```diff
@@ -30,6 +30,6 @@ public void Cannot_create_tier_name_with_invalid_name_length(int length)
 
         isTierNameInvalid.Should().BeTrue();
         errorCode.Should().Be("error.platform.validation.invalidTierName");
-        errorMessage.Should().Be($"Tier Name length must be between {TierName.MINIMUM_LENGTH} and {TierName.MAX_LENGTH}");
+        errorMessage.Should().Be($"Tier Name length must have a length between {TierName.MIN_LENGTH} and {TierName.MAX_LENGTH}");
```
tnotheis, line 33: Damn. I'm too tired. Didn't see that the word "length" was already contained in the message. So you should revert the change. Sorry :)

tnotheis (changes requested):

tnotheis (approved):

PR #61, opened 2023-04-04
Integration tests for tokens

KONKdiogo-ferraz:
Added the Integration tests for the Tokens service.


Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Token.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.ComponentModel.DataAnnotations;
+using Newtonsoft.Json;
+
+namespace Tokens.API.Tests.Integration.Models;
+public class Token
```
tnotheis, line 5: Why are there so many properties optional?

Modules/Tokens/test/Tokens.API.Tests.Integration/Features/GetTokenById.feature
```diff
@@ -0,0 +1,54 @@
...
+	Then the response status code is 200 (OK)
+	And the response contains a Token
+
+Scenario: Requesting a list of own Tokens
+	Given the user is authenticated
+	And the user created multiple Tokens
+	When a GET request is sent to the Tokens endpoint with the Tokens list
+	Then the response status code is 200 (OK)
+	And the response does not contain any peer Tokens
+
```
tnotheis, line 33: There should be a test where we have a Given that makes sure there is 1 own Token and 1 peer Token. Then we pass the id of both. As a result, we expect to only receive the own Token.

tnotheis (changes requested):

Modules/Tokens/test/Tokens.API.Tests.Integration/Features/CreateToken.feature
```diff
@@ -0,0 +1,88 @@
...
+Scenario: Creating a Token as an anonymous user
+	Given the user is unauthenticated
+	When a POST request is sent to the Tokens endpoint with
+		| Key       | Value      |
+		| Content   | QQ==       |
+		| ExpiresAt | <tomorrow> |
+	Then the response status code is 401 (Unauthorized)
+
+Scenario: Creating a Token without the "Content" request content property
+	Given the user is authenticated
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: The POST requests require authentication otherwise we will get a 401 Unauthorized status code

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Token.cs
KONKdiogo-ferraz: The response content received from a POST and GET requests are very different. A POST request only returns an Id and an expiresAt properties, while a get request return id, createdBy, createdByDevice, createdAt, expiresAt and a content porperties. To validate the json schema I had to turn those properties optional or I would require two different json schemas for a POST and GET requests.

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/TokenRequest.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Tokens.API.Tests.Integration.Models;
+public class TokenRequest
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: Yes, I agree. Good point!

Modules/Tokens/test/Tokens.API.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -0,0 +1,263 @@
...
+
+        AssertStatusCodeIsSuccess();
+
+        AssertResponseContentTypeIsJson();
+
+        AssertResponseContentCompliesWithSchema();
+
+        if (_tokenResponse.HttpMethod == Method.Get.ToString())
+        {
+            AssertExpirationDateIsInFuture();
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: I'm assuming you're asking about the 
````csharp
if (_tokenResponse.HttpMethod == Method.Get.ToString())
````

If so, please check my answer to a previous comment: https://github.com/nmshd/backbone/pull/61#discussion_r1157393643

File: Modules/Tokens/test/Tokens.API.Tests.Integration/Features/CreateToken.feature (hunk 3 above)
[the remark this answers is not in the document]
Comment (tnotheis) [reply Modules/Tokens/test/Tokens.API.Tests.Integration/Features/CreateToken.feature:24]: Right. :)

File: Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Token.cs (hunk 1 above)
Comment (tnotheis) [reply Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Token.cs:5]: But... then the existence of these properties is not validated correctly.

File: Modules/Tokens/test/Tokens.API.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs (hunk 5 above)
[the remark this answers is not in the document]
Comment (tnotheis) [reply Modules/Tokens/test/Tokens.API.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs:232]: No, I'm actually referring to the highlighted line (`AssertExpirationDateIsInFuture`). 
I assume you copied it from the Challenges tests. But for the Tokens, the expiration date is passed as a parameter during creation. And if it is in the past, the POST endpoint returns a 400.
So this assertion will never fail and can therefore be removed.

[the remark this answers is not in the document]
Comment (tnotheis) [reply Modules/Tokens/test/Tokens.API.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs:232]: To be precise: it CAN fail if too much time has passed since the creation of the Token and the Get. But since all of your Tokens expire "tomorrow", this will never be the case.

File: Modules/Tokens/test/Tokens.API.Tests.Integration/Models/CreateTokenResponse.cs [hunk 6]
```diff
@@ -0,0 +1,6 @@
+﻿namespace Tokens.API.Tests.Integration.Models;
+public class CreateTokenResponse
+{
+    public string? Id { get; set; }
+    public string? CreatedAt { get; set; }
```
tnotheis, line 5: Why optional?

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Token.cs
tnotheis: Why are the properties still optional, even though you have a separate CreateTokenResponse now?

tnotheis (changes requested):

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Token.cs
KONKdiogo-ferraz: This was all due to the warnings from VS

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/HttpResponse.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Net;
+
+namespace Tokens.API.Tests.Integration.Models;
+public class HttpResponse<T>
+{
+    public T? Data { get; set; }
```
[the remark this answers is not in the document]
KONKdiogo-ferraz: If we want to access a Token's ```Content``` property, then it would look like the following: ```httpResponse.Content.Result.Content```. By using ```Data``` it looks like the following: ```httpResponse.Data.Result.Content``` and I thought it looked better. Toughts?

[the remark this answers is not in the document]
tnotheis: Looks weird, I agree. But I think it's still better than living with the inconsistency.

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Token.cs
tnotheis: Which warnings?

KONKdiogo-ferraz: [CS8618](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings?f1url=%3FappId%3Droslyn%26k%3Dk(CS8618)#nonnullable-reference-not-initialized) - Non-nullable variable must contain a non-null value when exiting constructor. Consider declaring it as nullable.

tnotheis: You can leave them unassigned. We have this problem in several places all over the codebase. We will have to find a way to fix that in a pretty way some day.

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/HttpResponse.cs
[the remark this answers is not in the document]
tnotheis: I don't think that we need the raw Content property, do we?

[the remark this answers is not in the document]
KONKdiogo-ferraz: We do to validate the json schema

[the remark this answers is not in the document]
tnotheis: Well, actually not. In the validate method, you always parse the raw string before you do the JSON schema validation. So you could as well just pass in the parsed object (aka the current Data property)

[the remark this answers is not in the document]
daniel-almeida-konkconsulting: Removed the raw Content property and renamed the generic type Data property to Content

Modules/Tokens/test/Tokens.API.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -116,7 +116,7 @@ public async Task GivenTheUserCreatedMultipleTokens()
             var createTokenRequest = new CreateTokenRequest
             {
                 Content = "QQ==",
-                ExpiresAt = DateTime.Now.AddDays(1).ToString("dd-MM-yyyy")
+                ExpiresAt = DateTime.Now.AddDays(1).ToString("MM-dd-yyyy")
```
tnotheis, line 119: Can you try to use `yyyy-MM-dd` as format? That should do it as well and looks better imo

Modules/Tokens/test/Tokens.API.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -160,10 +160,10 @@ public async Task WhenAPOSTRequestIsSentToTheTokensEndpointWith(Table table)
             switch (requestConfiguration.Content)
             {
                 case var c when c.Contains("<tomorrow>"):
-                    requestConfiguration.Content = requestConfiguration.Content.Replace("<tomorrow>", DateTime.Now.AddDays(1).ToString("dd-MM-yyyy"));
+                    requestConfiguration.Content = requestConfiguration.Content.Replace("<tomorrow>", DateTime.Now.AddDays(1).ToString("MM-dd-yyyy"));
```
tnotheis, line 163: Maybe you should create two properties `TomorrowAsString` and `YesterdayAsString`, so that you don't need to duplicate the `DateTime.Now.AddDays(1).ToString("MM-dd-yyyy")` statement over and over again.

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/HttpResponse.cs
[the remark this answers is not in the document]
tnotheis: Looks good

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Response.cs
tnotheis, line 1: Rename the file to match the class name

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/HttpResponse.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Net;
+
+namespace Tokens.API.Tests.Integration.Models;
+public class HttpResponse<T>
+{
+    public T Content { get; set; }
```
tnotheis, line 6: I think you can use `ResponseContent<T>` here. You pass `ResponseContent` on every usage anyway, as far as I can see

```suggestion
    public ResponseContent<T> Content { get; set; }
```

Modules/Tokens/test/Tokens.API.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    private void AssertResponseContentTypeIsJson()
+    {
+        _tokenResponse.ContentType.Should().Be("application/json");
+    }
+
+    private void AssertResponseContentCompliesWithSchema<T>()
+    {
+        JsonValidators.ValidateJsonSchema<ResponseContent<T>>(JsonConvert.SerializeObject(_tokenResponse.Content!), out var errors)
```
tnotheis, line 298: Hmm, I think I see why Diogo had the raw content property. By deserializing and re-serializing, the JSON is distorted. 

So I suggest to add the property again - with a different name though. I'd suggest `RawContent`

Modules/Tokens/test/Tokens.API.Tests.Integration/Models/Response.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Newtonsoft.Json;
+
+namespace Tokens.API.Tests.Integration.Models;
+public class Response<T>
```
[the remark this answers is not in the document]
tnotheis: I think that's what I meant. :)

Modules/Tokens/test/Tokens.API.Tests.Integration/API/TokensApi.cs
```diff
@@ -18,23 +18,23 @@ public TokensApi(RestClient client)
...
 
-    public async Task<HttpResponse<ResponseContent<Token>>> GetTokenById(RequestConfiguration requestConfiguration, string id)
+    public async Task<HttpResponse<Token>> GetTokenById(RequestConfiguration requestConfiguration, string id)
     {
-        return await ExecuteTokenRequest<ResponseContent<Token>>(Method.Get, new PathString(ROUTE_PREFIX).Add($"/tokens/{id}").ToString(), requestConfiguration);
+        return await ExecuteTokenRequest<Token>(Method.Get, new PathString(ROUTE_PREFIX).Add($"/tokens/{id}").ToString(), requestConfiguration);
     }
 
-    public async Task<HttpResponse<ResponseContent<IEnumerable<Token>>>> GetTokenById(RequestConfiguration requestConfiguration, IEnumerable<string> ids)
+    public async Task<HttpResponse<IEnumerable<Token>>> GetTokenByIds(RequestConfiguration requestConfiguration, IEnumerable<string> ids)
```
tnotheis, line 31:
```suggestion
    public async Task<HttpResponse<IEnumerable<Token>>> GetTokensByIds(RequestConfiguration requestConfiguration, IEnumerable<string> ids)
```

tnotheis (approved):

PR #64, opened 2023-04-04
Auto create basic tier

Dannyps:


Backbone.API/ApplicationDbContextSeed.cs
```diff
@@ -58,4 +60,14 @@ private async Task SeedApplicationUsers(DevicesDbContext context)
 
         await context.SaveChangesAsync();
     }
+
+    private async Task SeedBasicTier(DevicesDbContext context)
+    {
+        if(await context.Tiers.AnyAsync())
+            return;
+
+        await context.Tiers.AddAsync(new Tier(TierName.Create("Basic Tier").Value));
```
tnotheis, line 69: The Tier name shouldn't contain the word "Tier". That's redundant.

tnotheis (changes requested):
There is something wrong with this PR. There are lots of changes in here that @daniel-almeida-konkconsulting made on another branch.

Dannyps:
This branch was based on `feature/create-tier--create-tier-command` but that branch does not seem to exist anymore, so I set `feature/create-tier` as base instead.

tnotheis:
Take a look at the diff. There is some pipeline stuff listed that shouldn't be there

Dannyps:
Sorry for the force, I couldn't revert a merge commit that was poisioning the PR. It should be good now.

tnotheis (changes requested):
Just noticed that there is a naming error on the APplicationDbContextSeed class, which originates from the microservice times.
The class should be called `DevicesDbContextSeed`. Could you fix that please?

tnotheis (approved):

PR #71, opened 2023-04-11
Update MediatR

tnotheis:
# Readiness checklist

- [x] I labeled the PR.


tnotheis:
I wanted to update MediatR in a separate PR because it required some changes to the codebase (not only to the csproj files).
There are two major changes we are affected from:
1. The package "MediatR.Extensions.Microsoft.DependencyInjection" does no longer exist. The code was moved to the main package (see https://github.com/jbogard/MediatR.Extensions.Microsoft.DependencyInjection)
2. When implement IRequest, there is no more need to return `Unit.Value`.

daniel-almeida-konkconsulting (approved):

PR #73, opened 2023-04-12
Automatically assign Basic Tier to Identities 

Dannyps:
# Readiness checklist

- [x] I labeled the PR.


tnotheis (changes requested):
Can you generate a unique database index for `Tier.Name` please? See https://github.com/nmshd/backbone/blob/main/Modules/Synchronization/src/Synchronization.Infrastructure/Persistence/Database/Configurations/DatawalletModificationEntityTypeConfiguration.cs#L11-L12 for an example

Modules/Devices/src/Devices.Domain/Entities/Identity.cs
```diff
@@ -25,6 +27,8 @@ public Identity(string? clientId, IdentityAddress address, byte[] publicKey, byt
 
     public byte IdentityVersion { get; set; }
 
+    public Tier? Tier { get; set; }
```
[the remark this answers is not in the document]
Dannyps: @tnotheis We have a chicken and egg problem here. The database may already have Identities, which do not have a Tier. When we add a TierId to the Identities table (via Migrations), we must set that collumn as nullable, hence that code. 
The alternative - having a default value - is inviable because the Basic Tier may not exist yet.
How should we address this?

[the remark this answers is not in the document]
Dannyps: A possible solution seems to be to make the attribute mandatory but tweak the migration to make the column nullable. I do not know how that can impact the runtime of the application though.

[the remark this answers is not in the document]
tnotheis: Okay, let's try to go with your suggested approach: make the property mandatory and make the column nullable. But don't edit the migration directly. Instead, you should edit the `IdentityEntityTypeConfiguration`:

```c#
builder.Property(x=>x.Tier).IsRequired(false)
```
I'm not sure though if that works in case of a navigation property. If not, I would suggest to leave it as it is for now, and try to remember to fix it in the future when the possibility that there are still people running a version without Tiers is almost 0.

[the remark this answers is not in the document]
Dannyps: We're using `.HasOne()`, which does not support `IsRequired` nor anything similar.

This SO answer seems to shed some light on the issue at hand: https://stackoverflow.com/a/35562728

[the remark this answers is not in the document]
tnotheis: Yeah, okay. Let's just ignore that for now. It's too much pain for too less benefit.

Modules/Devices/src/Devices.Application/Identities/Commands/CreateIdentity/Handler.cs
```diff
@@ -44,7 +45,9 @@ public async Task<CreateIdentityResponse> Handle(CreateIdentityCommand command,
         if (existingIdentity != null)
             throw new OperationFailedException(ApplicationErrors.Devices.AddressAlreadyExists());
 
-        var newIdentity = new Identity(command.ClientId, address, command.IdentityPublicKey, command.IdentityVersion);
+        var basicTier = await _dbContext.Set<Tier>().BasicTier(cancellationToken);
```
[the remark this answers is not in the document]
Dannyps: I have done this, but I'm not happy about the solution. See 02c60d2294debef148884b5519bfbb758564fde2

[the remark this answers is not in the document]
tnotheis: Why not? I don't see a problem with it.

Modules/Devices/src/Devices.Domain/Entities/Identity.cs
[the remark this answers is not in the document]
tnotheis: No worries. I would have done it myself. 😆

Modules/Devices/src/Devices.Application/Identities/Commands/CreateIdentity/Handler.cs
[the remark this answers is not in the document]
Dannyps: It feels that we just added redundant code, but I understand the logic. Ok, cool

Dannyps (commented):
The requested changes have been done.

Backbone.API/DevicesDbContextSeed.cs
```diff
@@ -32,15 +32,20 @@ private async Task SeedApplicationUsers(DevicesDbContext context)
         if (await context.Users.AnyAsync())
             return;
 
+        var basicTier = await context.Tiers.GetBasicTier(CancellationToken.None);
+        if (basicTier == null)
+            return;
```
tnotheis, line 37: You should throw an exception if it doesn't exist, so that we know that something's wrong. You can stick with the one you threw in the previous version:
```c#
throw new Exception("Basic Tier was not found.");
```

Backbone.API/DevicesDbContextSeed.cs
```diff
@@ -75,20 +81,16 @@ private async Task SeedBasicTier(DevicesDbContext context)
...
+
+        var identitiesWithoutTier = context.Identities.Where(i => i.Tier == null).ToList();
+        identitiesWithoutTier.ForEach(it =>
         {
-            throw new Exception("Basic Tier was not found.");
-        }
+            it.Tier = basicTier;
+        });
+
+        await context.SaveChangesAsync();
```
tnotheis: You should replace it with

```c#
await context.Identities.Where(i => i.Tier == null).ExecuteUpdateAsync(s => s.SetProperty(i => i.Tier, basicTier));
```
in order to reduce the number of database roundtrips. Depending on the number of Identities, this could make a huge difference.

I never used the `ExecuteUpdateAsync` method before, so I don't know if it works with navigation properties.

Modules/Devices/src/Devices.Domain/Entities/Identity.cs
```diff
@@ -6,15 +6,14 @@ namespace Backbone.Modules.Devices.Domain.Entities;
 
 public class Identity
 {
-    public Identity(string? clientId, IdentityAddress address, byte[] publicKey, Tier tier, byte identityVersion)
+    public Identity(string? clientId, IdentityAddress address, byte[] publicKey, byte identityVersion)
```
tnotheis: Why did you remove the Tier parameter from the constructor? 
It shouldn't be possible to create an Identity without a Tier.

Modules/Devices/src/Devices.Infrastructure.Database.SqlServer/Migrations/20230413132800_IdentityBasicTier.cs
```diff
@@ -29,7 +36,8 @@ protected override void Up(MigrationBuilder migrationBuilder)
                 table: "Identities",
                 column: "TierId",
                 principalTable: "Tier",
-                principalColumn: "Id");
+                principalColumn: "Id",
+                onDelete: ReferentialAction.Cascade);
```
tnotheis, line 40: This is evil. This means that if we delete a Tier, all Identities assigned to the Tier will be deleted.

tnotheis (changes requested):

Backbone.API/DevicesDbContextSeed.cs
Dannyps: I didn't pay a lot of attention to this because I supposed all changes would be commited at once on `await context.SaveChangesAsync();`.

It seems EF cannot handle it when we set a new navigation property. I'm trying to workaround the issue.

Modules/Devices/src/Devices.Domain/Entities/Identity.cs
Dannyps: EF needs to create instances of the object and cannot have a navigation property in the constructuor, as far as I could see. The error I got was the following:

```
System.InvalidOperationException: 'No suitable constructor was found for entity type 'Identity'. The following constructors had parameters that could not be bound to properties of the entity type: cannot bind 'tier' in 'Identity(..., Tier tier)'.'
```

Modules/Devices/src/Devices.Infrastructure.Database.SqlServer/Migrations/20230413132800_IdentityBasicTier.cs
Dannyps: Indeed! Should we set those as Null, or restrict the delete? We do not have a default value.

Backbone.API/DevicesDbContextSeed.cs
tnotheis: > I supposed all changes would be commited at once on `await context.SaveChangesAsync();`.

They are commited at once. But you still need 2 roundtrips: one for the update and one for the select (line 88).

If my suggested solution doesn't work for navigation properties, you should at least call `ToListAsync` instead of `ToList` in line 88.

Modules/Devices/src/Devices.Domain/Entities/Identity.cs
tnotheis: In that case, we should change it completely (I wanted to suggest that already a few days ago): instead of saving a reference to the Tier, we should only save the TierId. That's the solution DDD suggests anyway (when referencing other Aggregates, only do so via their Id and never the whole object)

Modules/Devices/src/Devices.Infrastructure.Database.SqlServer/Migrations/20230413132800_IdentityBasicTier.cs
tnotheis: Since we don't want to have Identities without a Tier, we should go with Restrict.

Modules/Devices/src/Devices.Domain/Entities/Identity.cs
Dannyps: In this case, will we have both a `Tier` and a `TierId` in the `Identity`?

tnotheis: Nope, only the TierId

Dannyps: The generated migration does not include a ForeignKey. Should I add it manually, or is there a way to specify that `TierId` must be a valid `Tier.Id`. I looked online but couldn't find anything.

Modules/Devices/src/Devices.Domain/Aggregates/Tier/TierId.cs
```diff
@@ -28,4 +28,12 @@ public static TierId Generate()
             return Result.Failure<TierId, DomainError>(validationResult);
         return Result.Success<TierId, DomainError>(new TierId(value));
     }
+
+    public static TierId New()
+    {
+
```
tnotheis, line 34: Redundant blank line

Modules/Devices/src/Devices.Domain/Aggregates/Tier/TierId.cs
```diff
@@ -28,4 +28,12 @@ public static TierId Generate()
             return Result.Failure<TierId, DomainError>(validationResult);
         return Result.Success<TierId, DomainError>(new TierId(value));
     }
+
+    public static TierId New()
+    {
+   
+        var tierIdAsString = StringUtils.Generate(DefaultValidChars, DEFAULT_MAX_LENGTH_WITHOUT_PREFIX);
```
tnotheis, line 35:
```suggestion
        var randomPart = StringUtils.Generate(DefaultValidChars, DEFAULT_MAX_LENGTH_WITHOUT_PREFIX);
```

Modules/Devices/src/Devices.Domain/Entities/Identity.cs
```diff
@@ -26,7 +28,8 @@ public Identity(string? clientId, IdentityAddress address, byte[] publicKey, byt
 
     public byte IdentityVersion { get; set; }
 
-    public Tier? Tier { get; set; }
+    [ForeignKey(nameof(Tier))]
```
tnotheis, line 31: Remember to remove this

tnotheis (changes requested):

tnotheis (approved):

PR #82, opened 2023-04-18
Docker support for Admin UI

Dannyps:
# Readiness checklist

- [x] I labeled the PR.

### DIFF
--- a/AdminUI/Dockerfile
+++ b/AdminUI/Dockerfile
@@ -1,3 +1,13 @@
-FROM nginx:alpine
-COPY /dist/admin-ui-app /usr/share/nginx/html
-EXPOSE 80
+# Build stage
+FROM node:lts-alpine as build-stage
+WORKDIR /app
+COPY package*.json /app/
+RUN npm install
+COPY ./ /app/
+ARG configuration=production
+RUN npm run build -- --output-path=./dist/out --configuration $configuration
+
+# Deploy stage
+FROM nginx:1.24
+COPY --from=build-stage /app/dist/out/ /usr/share/nginx/html
+COPY nginx.conf /etc/nginx/conf.d/default.conf
--- a/AdminUI/Dockerfile
+++ b/AdminUI/Dockerfile
@@ -1,5 +1,5 @@
 # Build stage
-FROM node:lts-alpine as build-stage
+FROM node:18.16-alpine as build-stage
 WORKDIR /app
 COPY package*.json /app/
 RUN npm install
--- a/AdminUI/Dockerfile
+++ b/AdminUI/Dockerfile
@@ -1,0 +1,3 @@
+FROM nginx:alpine
+COPY /dist/admin-ui-app /usr/share/nginx/html
+EXPOSE 80
--- a/AdminUI/Dockerfile
+++ b/AdminUI/Dockerfile
@@ -5,9 +5,9 @@ COPY package*.json /app/
 RUN npm install
 COPY ./ /app/
 ARG configuration=production
-RUN npm run build -- --output-path=./dist/out --configuration $configuration
+RUN npm run build -- --output-path=./dist --configuration $configuration
 
 # Deploy stage
 FROM nginx:1.24
-COPY --from=build-stage /app/dist/out/ /usr/share/nginx/html
+COPY --from=build-stage /app/dist/ /usr/share/nginx/html
 COPY nginx.conf /etc/nginx/conf.d/default.conf
--- a/AdminUI/Dockerfile.debug
+++ b/AdminUI/Dockerfile.debug
@@ -1,5 +1,5 @@
 # Build stage
-FROM node:18-alpine as build-stage
+FROM node:18.16-alpine as build-stage
 EXPOSE 4200
 
 WORKDIR /app
--- a/AdminUI/Dockerfile.debug
+++ b/AdminUI/Dockerfile.debug
@@ -1,0 +1,13 @@
+# Build stage
+FROM node:16-alpine as build-stage
+EXPOSE 4200
+
+WORKDIR /app
+
+# all configuration files are json
+COPY package*.json /app/
+COPY angular.json /app/
+COPY tsconfig*.json /app/
+RUN npm install
+RUN npm install -g @angular/cli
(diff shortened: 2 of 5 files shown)


AdminUI/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+COPY package*.json /app/
+RUN npm install
+COPY ./ /app/
+ARG configuration=production
+RUN npm run build -- --output-path=./dist/out --configuration $configuration
+
+# Deploy stage
+FROM nginx:1.24
+COPY --from=build-stage /app/dist/out/ /usr/share/nginx/html
+COPY nginx.conf /etc/nginx/conf.d/default.conf
```
tnotheis, line 13: Why no entrypoint?

AdminUI/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Build stage
+FROM node:lts-alpine as build-stage
+WORKDIR /app
+COPY package*.json /app/
+RUN npm install
+COPY ./ /app/
+ARG configuration=production
+RUN npm run build -- --output-path=./dist/out --configuration $configuration
```
tnotheis, line 8: I'm not familiar with angular, but `/dist/out` seems redundant to me. I would choose either `dist` or `out` (probably `dist`)

tnotheis (changes requested):
Can you add the Admin UI to the docker-compose.yml please? I changed the issue description accordingly.
And I would suggest to change the title of the PR to "Docker support for Admin UI"

Dannyps:
@tnotheis regarding the docker-compose, am I correct in assuming it'll be used merely for development? The way the app is build for production is radically different from the way it is built for development, which means we'll have to use a different approach/Dockerfile.

tnotheis:
Yes, you're right. It's only used for development. 
There is a separate Dockerfile for the Consumer API as well, so feel free do create a second one if necessary.

tnotheis:
Btw, the development Dockerfile for the Consumer API is called `Dockerfile.debug.`

AdminUI/Dockerfile
Dannyps: AS you can see in [this guide](https://www.docker.com/blog/how-to-use-the-official-nginx-docker-image/), the used image provides a Docker _underlying entrypoint_, which removes the need to provide one.

tnotheis: Mhh, okay. Seems to be the official way, so go for it. :)

AdminUI/Dockerfile.debug
```diff
@@ -0,0 +1,13 @@
+# Build stage
+FROM node:16-alpine as build-stage
+EXPOSE 4200
+
+WORKDIR /app
+
+# all configuration files are json
+COPY package*.json /app/
+COPY angular.json /app/
+COPY tsconfig*.json /app/
```
tnotheis, line 10: As far as I can see, these lines are unnecessary. Because either you mount all the files from "/AdminUI/src" into the container (in this case, you have these three files anyway) or you don't mount them at all (in this case, you are missing all the other files). 
I would just remove these lines.

docker-compose/docker-compose.yml
```diff
@@ -49,6 +49,21 @@ services:
...
+    build:
+      dockerfile: Dockerfile.debug
+      context: "../AdminUI/"
+    ports:
+      - "5020:4200"
+    depends_on:
+      - "adminAPI"
+    volumes:
+      - ../AdminUI/src:/app/src/:ro
+    command: "ng serve --host 0.0.0.0 --poll 1000"
```
tnotheis, line 66: Why don't you specify this as the entrypoint of the Dockerfile?

AdminUI/Dockerfile.debug
```diff
@@ -0,0 +1,13 @@
+# Build stage
+FROM node:16-alpine as build-stage
```
tnotheis, line 2: node 18 is the current LTS version. Why don't you use that one?

tnotheis (commented):
You are currently using `node:lts` for the build stage of the production Dockerfile. We should use a fix version instead (probably 18.16-alpine) in order to reduce the number of surprise caused by new features or breaking changes from the dynamic tag.

AdminUI/Dockerfile.debug
Dannyps: Those files are on the root of the `AdminUI` folder, which contains the `src` folder as well. I would share `angular.json` and `tsconfig*.json` whith the host system the same way we do with the `src` folder, but it turns out that you can't have a volume that is not a directory.
Sharing the whole `AdminUI` is not viable because we'd be poluting the host system with compilation artifacts such as the `node_modules` folder.
Moreover, the `package*.json` files are necessary to build the container, which makes it absolutely mandatory that they're copied.

AdminUI/Dockerfile.debug
Dannyps: Remnants of debugging, I'll fix it.

AdminUI/Dockerfile.debug
tnotheis: Oh, yeah, I didn't see that you only mount the `src` folder.

docker-compose/docker-compose.yml
tnotheis: Yeah, I think I would move it to the `Dockerfile`. The whole `Dockerfile.debug` doesn't work, when you don't mount the files. So why should we make an exception for the entrypoint? :)

AdminUI/Dockerfile.debug
```diff
@@ -1,5 +1,5 @@
 # Build stage
-FROM node:16-alpine as build-stage
+FROM node:18-alpine as build-stage
```
tnotheis, line 2: You should use 18.16 here as well, so that the behavior of the `Dockerfile` and `Dockerfile.debug` is similar, and there are not surprises in the deployed version.

tnotheis (changes requested):

docker-compose/docker-compose.yml
Dannyps: The entrypoint is not part of the building itself. Makes sense

tnotheis (approved):

PR #83, opened 2023-04-19
Repositories for Messages module

Dannyps:
# Readiness checklist

- [x] I labeled the PR.


Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+    private readonly IQueryable<Message> _readOnlyMessages;
+    private readonly IUserContext _userContext;
+
+    public MessagesRepository(MessagesDbContext dbContext, IUserContext userContext)
+    {
+        _messages = dbContext.Messages;
+        _readOnlyMessages = dbContext.Messages.AsNoTracking();
+        _userContext = userContext;
+    }
+    async Task<Message> IMessagesRepository.Find(MessageId id, CancellationToken cancellationToken)
```
tnotheis, line 27: Add the `public` modifier to each method please.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -0,0 +1,75 @@
...
+using Backbone.Modules.Messages.Application.Messages.Queries.ListMessages;
+
+namespace Backbone.Modules.Messages.Infrastructure.Persistence.Database.Repository;
+public class MessagesRepository : IMessagesRepository
+{
+    private readonly DbSet<Message> _messages;
+    private readonly IQueryable<Message> _readOnlyMessages;
+    private readonly IUserContext _userContext;
+
+    public MessagesRepository(MessagesDbContext dbContext, IUserContext userContext)
```
tnotheis, line 21: I'm not 100% convinced, but I think we shouldn't use the `IUserContext` in the repository. Instead, we should pass the active Identity into each method that needs it.
It's the responsibility of the Application layer to deal with users, and not the one of the Infrastructure layer.

Modules/Messages/src/Messages.Infrastructure/Persistence/IServiceCollectionExtensions.cs
```diff
@@ -22,6 +25,11 @@ public static void AddPersistence(this IServiceCollection services, PersistenceO
             blobOptions.RootFolder = options.BlobStorageOptions.Container);
         services.AddBlobStorage(options.BlobStorageOptions);
     }
+
+    public static void AddRepositories(this IServiceCollection services)
+    {
+        services.AddTransient<IMessagesRepository, MessagesRepository>();
+    }
```
tnotheis, line 32: I know you copied that from the Tokens module, but I guess we should remove this method and move the logic to the `AddPersistence` method. I don't know why I created it back then.

tnotheis (changes requested):
I forgot to mention that the responsibility of the repositories should also be to fill the content from the blob storage if required. 
If all methods of the repository fill the content, then the names can stay as they are. But if there is a method (let's say `FindAll`) that is used two scenarios, one where the content is required, and another one where it isn't, then you should add a parameter of type `FetchMode` (an enum with the two possible values `IncludeContent` and `SkipContent`) with a default value of `SkipContent`.

You didn't test your changes, did you? Because I think there will be an error on the `GET /Messages` and the `GET /Messages/{id}` routes. You forgot to remove the `IMessagesDbContext` interface. If you delete it, you will find another usage, which is in a class called `MessageService`, with the two methods `MarkMessageAsReceived` and `MarkMessagesAsReceived`. These two methods update the passed Messages and save them in the database. The methods are called from the query handlers for `GetMessageQuery` and `ListMessagesQuery`. And since these handlers now use the repository, which uses the `_readOnlyDbSet` to read them, they can't be saved, because they are untracked (`_readOnlyMessages = dbContext.Messages.AsNoTracking()`).

Another place that still uses `IMessagesDbContext` is the handler for the `SendMessageCommand`. There it is used to check whether a Relationship between the Sender and the Recipients of the Message exists. You should introduce a `RelationshipsRepository` for that.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
Dannyps: Agreed

Modules/Messages/src/Messages.Application/Messages/Commands/SendMessage/Handler.cs
```diff
@@ -59,14 +71,13 @@ public async Task<SendMessageResponse> Handle(SendMessageCommand request, Cancel
 
     private async Task SaveMessage(Message message, CancellationToken cancellationToken)
     {
-        await _dbContext.Set<Message>().AddAsync(message, cancellationToken);
-        _blobStorage.Add(_blobOptions.RootFolder, message.Id, message.Body);
+        await _messagesRepository.Add(message, cancellationToken);
 
+        _blobStorage.Add(_blobOptions.RootFolder, message.Id, message.Body);
```
tnotheis, line 76: The MessagesRepository should be responsible for saving the content. That's one of the main reasons for this whole refactoring. Because at the moment, the Command and Query handlers need to know way to much about how entities are persisted. 

So you should move the following line:

```
_blobStorage.Add(_blobOptions.RootFolder, message.Id, message.Body);
```

to the `MessagesRepository`. Also, remove the `IBlobStorage` dependency from the constructor of this class.

(I just realized that I wasn't explicit about that point in the task description and updated it accordingly)

Modules/Messages/src/Messages.Application/Infrastructure/Persistence/Repository/IMessagesRepository.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Backbone.Modules.Messages.Application.Infrastructure.Persistence.Repository;
+public interface IMessagesRepository
+{
+    Task<DbPaginationResult<Message>> FindAll(PaginationFilter paginationFilter);
+    Task<DbPaginationResult<Message>> FindMessagesOfIdentity(IdentityAddress identityAddress, ListMessagesQuery request, IdentityAddress addressOfActiveIdentity, CancellationToken cancellationToken);
+    Task<Message> Find(MessageId id, IdentityAddress address, CancellationToken cancellationToken);
+    Task<MessageId> Add(Message message, CancellationToken cancellationToken);
+    Task<int> CountUnreceivedMessagesFromActiveIdentity(IdentityAddress sender, SendMessageCommandRecipientInformation recipientDto, CancellationToken cancellationToken);
+    EntityEntry<Message> Update(Message message);
```
tnotheis, line 18: `EntityEntry` is a EF Core class. Having this in `Messages.Application` disables us to remove the reference to EF Core from this project. Just return `Task`.

Modules/Messages/src/Messages.Application/Messages/Queries/GetMessage/Handler.cs
```diff
@@ -1,38 +1,30 @@
...
     {
-        var addressOfActiveIdentity = _userContext.GetAddress();
-
-        var message = await _dbContext
-            .Set<Message>()
-            .IncludeAllReferences(addressOfActiveIdentity)
-            .WithSenderOrRecipient(_userContext.GetAddress())
-            .FirstWithId(request.Id, cancellationToken);
-
+        var message = await _messagesRepository.Find(request.Id, _userContext.GetAddress(), cancellationToken);
```
tnotheis, line 26: Here a reminder of the comment in my last review, because I think that this is still a problem:

> You didn't test your changes, did you? Because I think there will be an error on the GET /Messages and the GET /Messages/{id} routes. You forgot to remove the IMessagesDbContext interface. If you delete it, you will find another usage, which is in a class called MessageService, with the two methods MarkMessageAsReceived and MarkMessagesAsReceived. These two methods update the passed Messages and save them in the database. The methods are called from the query handlers for GetMessageQuery and ListMessagesQuery. And since these handlers now use the repository, which uses the _readOnlyDbSet to read them, they can't be saved, because they are untracked (_readOnlyMessages = dbContext.Messages.AsNoTracking()).

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -0,0 +1,81 @@
...
+            .AsQueryable()
+            .IncludeAllReferences();
+
+        if (request.Ids.Any())
+            query = query.WithIdsIn(request.Ids);
+
+        query = query.WithSenderOrRecipient(identityAddress);
+        query = query.DoNotSendBeforePropertyIsNotInTheFuture();
+
+        return query.OrderAndPaginate(d => d.CreatedAt, request.PaginationFilter);
```
tnotheis, line 72: I know that you copied this from my code, but I don't know why I did it that way. :)

The following is better:

```suggestion
        return query.WithSenderOrRecipient(identityAddress);
                      .DoNotSendBeforePropertyIsNotInTheFuture();
                      .OrderAndPaginate(d => d.CreatedAt, request.PaginationFilter);
```

Modules/Messages/src/Messages.Jobs.SanityCheck/Infrastructure/DataSource/DataSource.cs
tnotheis, line 1: The sanity check is infrastructure specific. This means that it's okay to use the DbContext here.

Just use `MessagesDbContext` instead of `IMessagesDbContext` and you should be fine.

tnotheis (changes requested):

Modules/Messages/src/Messages.Application/Messages/MessageService.cs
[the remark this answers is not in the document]
Dannyps: @tnotheis, the `MarkMessageAsReceived` method family have some logic, particularly checking if the message has a recipient, setting the `ReceivedAt` and the `ReceivedByDevice`. Should this be in the repository?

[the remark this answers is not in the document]
tnotheis: You're right, that's probably not ideal. Good catch!!

I thought about it again, and I came to the conclusion that we don't necessarily need the methods anymore after removing the triggering of the integration events. 

I suggest that we move the check for recipient.ReceivedAt.HasValue into the `RecipientInformation.ReceivedMessage` method. And maybe we should rename the method to `FetchedMessage`.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -72,10 +67,10 @@ public Task<DbPaginationResult<Message>> FindMessagesOfIdentity(IdentityAddress
         return query.OrderAndPaginate(d => d.CreatedAt, request.PaginationFilter);
     }
 
-    public EntityEntry<Message> Update(Message message)
+    public Message Update(Message message)
     {
         var m = _messages.Update(message);
         _dbContext.SaveChanges();
-        return m;
+        return m.Entity;
```
tnotheis, line 74: No need to return the Message. Whoever passes it to the method already has it - otherwise they couldn't have passed it. ;) 
It only causes confusion, because you would wonder if the returned Message is different than the one that was passed.

Modules/Messages/src/Messages.Application/Messages/MessageService.cs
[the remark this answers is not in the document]
Dannyps: Should the `ReceivedMessage` method return a boolean indicating if the message was marked as received?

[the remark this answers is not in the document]
tnotheis: Currently not. We do that as soon as we need it.

Modules/Messages/src/Messages.Application/Messages/Queries/GetMessage/Handler.cs
Dannyps: Using the debugger and checking on the database directly, I have confirmed that the way the message is fetched does not impact the write; possibly because it takes place with the appropriate dbContext, via the repo.
i.e., the expected behaviour is working.

tnotheis: Okay, maybe they changed this behavior. But I think we should still use the tracked queries here. Because 
1. it is recommended in the docs: https://learn.microsoft.com/en-us/ef/core/querying/tracking
2. if the Messages are untracked, I would bet that even though you only update a single property, the whole object is updated on the database, resulting in an unnecessary amount of data sent to the database.

I think we should add a parameter `track` of type `bool` that defaults to `false`, and that we set to `true` in case we want to edit the Message(s)

Modules/Messages/src/Messages.Application/Infrastructure/Persistence/Repository/IMessagesRepository.cs
```diff
@@ -1,19 +1,16 @@
...
 {
-    Task<DbPaginationResult<Message>> FindAll(PaginationFilter paginationFilter);
-    Task<DbPaginationResult<Message>> FindMessagesOfIdentity(IdentityAddress identityAddress, ListMessagesQuery request, IdentityAddress addressOfActiveIdentity, CancellationToken cancellationToken);
-    Task<Message> Find(MessageId id, IdentityAddress address, CancellationToken cancellationToken);
+    Task<DbPaginationResult<Message>> FindMessagesWithIds(IEnumerable<MessageId> ids, IdentityAddress requiredParticipant, PaginationFilter paginationFilter, bool track);
+    Task<Message> Find(MessageId id, IdentityAddress requiredParticipant, CancellationToken cancellationToken, bool track);
     Task<MessageId> Add(Message message, CancellationToken cancellationToken);
-    Task<int> CountUnreceivedMessagesFromActiveIdentity(IdentityAddress sender, SendMessageCommandRecipientInformation recipientDto, CancellationToken cancellationToken);
-    EntityEntry<Message> Update(Message message);
+    Task<int> CountUnreceivedMessagesFromSenderToReceiver(IdentityAddress sender, IdentityAddress recipient, CancellationToken cancellationToken);
```
tnotheis, line 13:
```suggestion
    Task<int> CountUnreceivedMessagesFromSenderToRecipient(IdentityAddress sender, IdentityAddress recipient, CancellationToken cancellationToken);
```

Modules/Messages/src/Messages.Application/Messages/Commands/SendMessage/Handler.cs
```diff
@@ -72,9 +66,6 @@ public async Task<SendMessageResponse> Handle(SendMessageCommand request, Cancel
     private async Task SaveMessage(Message message, CancellationToken cancellationToken)
     {
         await _messagesRepository.Add(message, cancellationToken);
-
-        _blobStorage.Add(_blobOptions.RootFolder, message.Id, message.Body);
-        await _blobStorage.SaveAsync();
     }
```
tnotheis: You can remove the whole method now.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -37,45 +41,49 @@ public async Task<Message> FindPlain(MessageId id, CancellationToken cancellatio
...
 
-    public EntityEntry<Message> Update(Message message)
+    public async Task FetchedMessage(Message message, IdentityAddress address, DeviceId deviceId)
     {
-        var m = _messages.Update(message);
-        _dbContext.SaveChanges();
-        return m;
+        var recipient = message.Recipients.FirstWithIdOrDefault(address);
+
+        recipient.ReceivedMessage(deviceId);
```
tnotheis, line 85: You misunderstood me here. I meant that you should rename the `recipient.ReceivedMessage` method to `recipient.FetchedMessage` and then call it directly from the Handlers. After that, you just need to call Update for the Message(s).
And you also should add an `Update` overload that accepts multiple Messages. Because currently, when you mark 1,000 Messages as received, you are making 1,000 database requests, which performs poorly.

Modules/Messages/src/Messages.Application/Messages/MessageService.cs
tnotheis, line 1: As I said in another comment, you can delete this whole class, because we don't need any of the methods anymore.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
tnotheis, line 1: You forgot to fill the content in the find methods

tnotheis (changes requested):

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
Dannyps: Due to other requested changes, I've had to change the `Find*` methods return types as follows:
```C#
Task<ListMessagesResponse> FindMessagesWithIds(...);
Task<MessageDTO> Find(...);
```

with these, I cannot call the `recipient.FetchedMessage` considering I do not have access to the message(s).

Am I missing something here, or is this indeed a problem? How should we proceed? Thanks.

tnotheis: Which requested changes do you mean? I definitely never intended your described change. The repositories should accept and return only aggregates - no DTOs, Commands, Queries or Responses.

Dannyps: For instance, let's focus on the `Find` method. The repository should fill the body from the BlobStorage. However, the `Message.Body` attribute is read only (`{ get; }`).

Considering the controller returns a `MessageDTO`, which has a writable `Body`, I assumed it would be the best approach. My bad.

Should the `Message.Body` attribute be made writable instead?

tnotheis: Oh, got it.

You should introduce I private setter for the `Body` property. And a method called `LoadBody`. This makes it explicit that it should only be used to load a body, and not to change it.

Inside of the method, you should additionally check if the Body is null. If it is not, throw an Exception.

Modules/Messages/src/Messages.Application/Infrastructure/Persistence/Repository/IMessagesRepository.cs
```diff
@@ -7,10 +7,10 @@
...
-    Task<DbPaginationResult<Message>> FindMessagesWithIds(IEnumerable<MessageId> ids, IdentityAddress requiredParticipant, PaginationFilter paginationFilter, bool track);
-    Task<Message> Find(MessageId id, IdentityAddress requiredParticipant, CancellationToken cancellationToken, bool track);
+    Task<DbPaginationResult<Message>> FindMessagesWithIds(IEnumerable<MessageId> ids, IdentityAddress requiredParticipant, PaginationFilter paginationFilter, bool track = false);
+    Task<Message> Find(MessageId id, IdentityAddress requiredParticipant, CancellationToken cancellationToken, bool track = false);
     Task<MessageId> Add(Message message, CancellationToken cancellationToken);
-    Task<int> CountUnreceivedMessagesFromSenderToReceiver(IdentityAddress sender, IdentityAddress recipient, CancellationToken cancellationToken);
+    Task<int> CountUnreceivedMessagesFromSenderToRecipient(IdentityAddress sender, IdentityAddress recipient, CancellationToken cancellationToken);
     Task Update(Message message);
-    Task FetchedMessage(Message message, IdentityAddress address, DeviceId deviceId);
+    Task Update(IEnumerable<Message> message);
```
tnotheis, line 15:
```suggestion
    Task Update(IEnumerable<Message> messages);
```

Modules/Messages/src/Messages.Application/Messages/Queries/GetMessage/Handler.cs
```diff
@@ -10,27 +11,24 @@ public class Handler : IRequestHandler<GetMessageQuery, MessageDTO>
...
-        await _messageService.MarkMessageAsReceived(message);
+        var message = await _messagesRepository.Find(request.Id, _userContext.GetAddress(), cancellationToken, track: true);
 
-        var response = _mapper.Map<MessageDTO>(message);
+        message.Recipients.FirstWithIdOrDefault(_userContext.GetAddress()).FetchedMessage(_userContext.GetDeviceId());
+
+        await _messagesRepository.Update(message);
 
-        if (!request.NoBody)
-            await _messageService.FillBody(response);
```
tnotheis, line 41: You just deleted this part of the code. So now, if someone passes `NoBody = true`, they still receive the body. 

The `MessagesRepository.Find` method needs a parameter for defining whether the body should be returned or not.

Modules/Messages/src/Messages.Domain/Entities/Message.cs
```diff
@@ -30,8 +30,17 @@ public Message(IdentityAddress createdBy, DeviceId createdByDevice, DateTime? do
...
+    public byte[] Body { get; private set; }
 
     public IReadOnlyCollection<Attachment> Attachments { get; }
     public IReadOnlyCollection<RecipientInformation> Recipients { get; }
+
+    public void LoadBody(byte[] bytes)
+    {
+        if(Body != null)
+        {
+            throw new InvalidOperationException($"The Body of the message {Id} is already filled.");
```
tnotheis, line 42:
```suggestion
            throw new InvalidOperationException($"The Body of the message {Id} is already filled. It is not possible to change it.");
```

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -19,21 +22,27 @@ public class MessagesRepository : IMessagesRepository
...
         _readOnlyMessages = dbContext.Messages.AsNoTracking();
         _dbContext = dbContext;
         _blobStorage = blobStorage;
         _blobOptions = blobOptions.Value;
+        _mapper = mapper;
     }
     public async Task<Message> Find(MessageId id, IdentityAddress address, CancellationToken cancellationToken, bool track = false)
     {
-        return await (track ? _messages : _readOnlyMessages)
+        var msg = await (track ? _messages : _readOnlyMessages)
```
tnotheis, line 38:
```suggestion
        var message = await (track ? _messages : _readOnlyMessages)
```

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -50,15 +59,15 @@ public async Task<MessageId> Add(Message message, CancellationToken cancellation
         return add.Entity.Id;
     }
 
-    public Task<int> CountUnreceivedMessagesFromSenderToReceiver(IdentityAddress sender, IdentityAddress recipient, CancellationToken cancellationToken)
+    public Task<int> CountUnreceivedMessagesFromSenderToRecipient(IdentityAddress sender, IdentityAddress recipient, CancellationToken cancellationToken)
     {
         return _readOnlyMessages
             .FromASpecificSender(sender)
             .WithASpecificRecipientWhoDidNotReceiveTheMessage(recipient)
             .CountAsync(cancellationToken);
```
tnotheis: Has to be awaited

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -50,15 +59,15 @@ public async Task<MessageId> Add(Message message, CancellationToken cancellation
...
     }
 
-    public Task<DbPaginationResult<Message>> FindMessagesWithIds(IEnumerable<MessageId> ids, IdentityAddress requiredParticipant, PaginationFilter paginationFilter, bool track = false)
+    public async Task<DbPaginationResult<Message>> FindMessagesWithIds(IEnumerable<MessageId> ids, IdentityAddress requiredParticipant, PaginationFilter paginationFilter, bool track = false)
```
tnotheis, line 70: You forgot to fill the Body's of the Messages...

tnotheis (changes requested):

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -22,27 +20,28 @@ public class MessagesRepository : IMessagesRepository
...
+    public async Task<Message> Find(MessageId id, IdentityAddress address, CancellationToken cancellationToken, bool track = false, bool noBody = false)
     {
-        var msg = await (track ? _messages : _readOnlyMessages)
+        var message = await (track ? _messages : _readOnlyMessages)
             .IncludeAllReferences()
             .WithSenderOrRecipient(address)
             .FirstWithId(id, cancellationToken);
 
-        msg.LoadBody(await _blobStorage.FindAsync(_blobOptions.RootFolder, msg.Id));
+        if (noBody == false)
```
tnotheis, line 39: The condition has to inverted. Further, you should simplify it to `if(!noBody)`

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -76,9 +75,16 @@ public async Task<DbPaginationResult<Message>> FindMessagesWithIds(IEnumerable<M
...
 
-        return await query.WithSenderOrRecipient(requiredParticipant)
+        var messages = await query.WithSenderOrRecipient(requiredParticipant)
             .DoNotSendBeforePropertyIsNotInTheFuture()
             .OrderAndPaginate(d => d.CreatedAt, paginationFilter);
+
+        foreach (var msg in messages.ItemsOnPage)
+        {
+            msg.LoadBody(await _blobStorage.FindAsync(_blobOptions.RootFolder, msg.Id));
+        }
```
tnotheis, line 85: This performs poorly. You should load all body in parallel.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/Repository/MessagesRepository.cs
```diff
@@ -76,9 +75,16 @@ public async Task<DbPaginationResult<Message>> FindMessagesWithIds(IEnumerable<M
         if (ids.Any())
             query = query.WithIdsIn(ids);
 
-        return await query.WithSenderOrRecipient(requiredParticipant)
+        var messages = await query.WithSenderOrRecipient(requiredParticipant)
             .DoNotSendBeforePropertyIsNotInTheFuture()
             .OrderAndPaginate(d => d.CreatedAt, paginationFilter);
+
+        foreach (var msg in messages.ItemsOnPage)
```
tnotheis, line 82:
```suggestion
        foreach (var message in messages.ItemsOnPage)
```

tnotheis (changes requested):

tnotheis (approved):

PR #85, opened 2023-04-20
Each Consumer API call returns 401 (Unauthorized)

tnotheis:
# Readiness checklist

- [x] I labeled the PR.


jkoenig134 (approved):

tnotheis:
I have narrowed the problem down to the Microsoft.IdentityModel.Tokens package. After upgrading to a version >6.27.0 the error occurs.

tnotheis:
In order to avoid accidental updates to the corrupted version, I have pinned it to 6.27.0. We need to check for new versions of this package regularly. I registered for notifications on the corresponding repository (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet).

PR #87, opened 2023-04-21
Tier name database column has fixed length

daniel-almeida-konkconsulting:
# Readiness checklist

- [X] I labeled the PR.


Modules/Devices/src/Devices.Infrastructure/Persistence/Database/DevicesDbContext.cs
```diff
@@ -95,7 +95,7 @@ protected override void ConfigureConventions(ModelConfigurationBuilder configura
             .HaveMaxLength(Username.MAX_LENGTH).HaveConversion<UsernameValueConverter>();
         configurationBuilder.Properties<TierId>().AreUnicode(false).AreFixedLength()
             .HaveMaxLength(TierId.MAX_LENGTH).HaveConversion<TierIdEntityFrameworkValueConverter>();
-        configurationBuilder.Properties<TierName>().AreUnicode(false).AreFixedLength()
+        configurationBuilder.Properties<TierName>().AreUnicode()
```
tnotheis, line 98: I actually would have made it explicit that the `TierName` is NOT fixed length by setting `AreFixedLength(false)`. Just to prevent that someone thinks "oh, somebody forgot to add the `AreFixedLengthPart`, let me better do that now". :)

tnotheis (changes requested):

PR #96, opened 2023-04-25
Make Admin CLI atomic

tnotheis:
# Readiness checklist

- [x] I labeled the PR.


tnotheis:
This is a total revamp of the Admin CLI. Instead of having an endless loop after running the CLI, you can now run commands like

```
backbone client create --clientId <client-id> --displayName <display-name> --clientSecret <client-secret>
```

which enables it to be run as part of an automation script.

PR #97, opened 2023-04-26
Invalid dev configuration

tnotheis:
# Readiness checklist

- [x] I labeled the PR.

### DIFF
--- a/Backbone.API/appsettings.override.json
+++ b/Backbone.API/appsettings.override.json
@@ -51,10 +51,7 @@
           "ConnectionString": "Server=ms-sql-server;Database=enmeshed;User Id=devices;Password=Passw0rd;TrustServerCertificate=True"
         },
         "PushNotifications": {
-          "Provider":  "Dummy",
-          "AzureNotificationHub": {
-            "HubName": "nh-dev"
-          }
+          "Provider":  "Dummy"
         }
       }
     },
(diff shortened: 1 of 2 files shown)


Backbone.API/appsettings.override.json
```diff
@@ -51,6 +51,7 @@
           "ConnectionString": "Server=ms-sql-server;Database=enmeshed;User Id=devices;Password=Passw0rd;TrustServerCertificate=True"
         },
         "PushNotifications": {
+          "Provider":  "Dummy",
           "AzureNotificationHub": {
```
jkoenig134, line 55: shouldn't this be removed if you are using the dummy provider?

Dannyps (dismissed):
As far as I can see, the Devices modules makes no use of the BlobStorage. Hence, this looks good. Please mind @jkoenig134's note. I think I understant what's going on, but it would be best if you could leave an explanation. Thanks

tnotheis:
@Dannyps I implemented a DummyPushService two days ago, which is mainly for local development, where we don't need push notifications. And I forgot to change the appsettings.override.json file to actually use it - even though I'm pretty sure I didn't forget it. This PR should fix it. But according to @jkoenig134's comment, I messed it up once again. :D I guess I'm loosing my mind.

PR #106, opened 2023-05-02
Database Seed makes use of Commands for Tiers and ApplicationUsers creation

Dannyps:
# Readiness checklist

- [x] I labeled the PR.

### DIFF
--- a/Backbone.API/DevicesDbContextSeed.cs
+++ b/Backbone.API/DevicesDbContextSeed.cs
@@ -1,9 +1,14 @@
 ﻿using Backbone.Modules.Devices.Application.Extensions;
+using Backbone.Modules.Devices.Application.Identities.Queries.ListIdentities;
+using Backbone.Modules.Devices.Application.Infrastructure.Persistence.Repository;
+using Backbone.Modules.Devices.Application.Tiers.Commands.CreateTier;
+using Backbone.Modules.Devices.Application.Users.Commands.SeedTestUsers;
 using Backbone.Modules.Devices.Domain.Aggregates.Tier;
 using Backbone.Modules.Devices.Domain.Entities;
 using Backbone.Modules.Devices.Infrastructure.Persistence.Database;
 using Enmeshed.DevelopmentKit.Identity.ValueObjects;
 using Enmeshed.Tooling;
+using MediatR;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.EntityFrameworkCore;
 
@@ -12,6 +17,12 @@ namespace Backbone.API;
 public class DevicesDbContextSeed
 {
     private readonly IPasswordHasher<ApplicationUser> _passwordHasher = new PasswordHasher<ApplicationUser>();
+    private readonly IMediator _mediator;
+
+    public DevicesDbContextSeed(IMediator mediator)
+    {
+        _mediator = mediator;
+    }
 
     public async Task SeedAsync(DevicesDbContext context)
     {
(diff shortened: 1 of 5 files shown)


Backbone.API/DevicesDbContextSeed.cs
```diff
@@ -36,47 +42,20 @@ private async Task SeedApplicationUsers(DevicesDbContext context)
...
+        try
+        {
+            var basicTier = await GetBasicTier(context);
             return;
-
-        await context.Tiers.AddAsync(new Tier(TierName.Create(TierName.BASIC_DEFAULT_NAME).Value));
-
-        await context.SaveChangesAsync();
+        }
+        catch (Exception)
```
tnotheis, line 55: Catching all exceptions is generally a bad idea. You should only catch the exception you expect (in this case it should be an ApplicationException, I think. And then you should even refine for the specific error code (in this case it should be `error.platform.recordNotFound`, if I remember i right).

Modules/Devices/src/Devices.Application/Users/Commands/SeedTestUsers/Handler.cs
```diff
@@ -0,0 +1,61 @@
...
+using Enmeshed.DevelopmentKit.Identity.ValueObjects;
+using Enmeshed.Tooling;
+using MediatR;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.EntityFrameworkCore;
+
+namespace Backbone.Modules.Devices.Application.Users.Commands.SeedTestUsers;
+public class Handler : IRequestHandler<SeedTestUsersCommand, SeedTestUsersResponse>
+{
+    private readonly IPasswordHasher<ApplicationUser> _passwordHasher = new PasswordHasher<ApplicationUser>();
```
tnotheis, line 13: Can you try to let the DI container inject the `IPasswordHasher<ApplicationUser>`? I would expect it to work.

Modules/Devices/src/Devices.Application/Users/Commands/SeedTestUsers/Handler.cs
```diff
@@ -0,0 +1,61 @@
...
+    private readonly DbSet<ApplicationUser> _usersDb;
+    private readonly IDevicesDbContext _context;
```
tnotheis, line 15: `_context` is confusing, because there is a `IUserContext` as well.

```suggestion
    private readonly IDevicesDbContext _dbContext;
```

Modules/Devices/src/Devices.Application/Users/Commands/SeedTestUsers/Handler.cs
```diff
@@ -0,0 +1,61 @@
...
+            UserName = "USRa",
+            NormalizedUserName = "USRA",
+            Device = new Device(new Identity("test",
+                IdentityAddress.Create(new byte[] { 1, 1, 1, 1, 1 }, "id1"),
+                new byte[] { 1, 1, 1, 1, 1 }, basicTier.Id, 1
+            )),
+            CreatedAt = SystemTime.UtcNow
+        };
+        user.PasswordHash = _passwordHasher.HashPassword(user, "a");
+        await _usersDb.AddAsync(user);
```
tnotheis, line 41: I would inline the `_usersDb` field:

```suggestion
        await _dbContext.Set<ApplicationUser>().AddAsync(user);
```

Modules/Devices/src/Devices.Application/Users/Commands/SeedTestUsers/SeedTestUsersCommand.cs
```diff
@@ -0,0 +1,6 @@
+﻿using MediatR;
+
+namespace Backbone.Modules.Devices.Application.Users.Commands.SeedTestUsers;
+public class SeedTestUsersCommand : IRequest<SeedTestUsersResponse>
```
tnotheis, line 4: There is no need to define a Response when it doesn't have any properties.

```suggestion
public class SeedTestUsersCommand : IRequest
```

tnotheis (changes requested):

tnotheis (approved):

PR #110, opened 2023-05-03
Configuration files for database migrations

daniel-almeida-konkconsulting:
# Readiness checklist

- [X] I labeled the PR.

### DIFF
--- a/scripts/windows/efcore/add_migration.ps1
+++ b/scripts/windows/efcore/add_migration.ps1
@@ -4,8 +4,7 @@ Param(
     [parameter(Mandatory)][ValidateSet("SqlServer", "Postgres", "")] $provider
 )
 
-$env:ASPNETCORE_ENVIRONMENT = 'Local'
-
+$environment="dbmigrations-" + $provider.ToLower()
 $repoRoot = git rev-parse --show-toplevel
 $dbContextName = "${moduleName}DbContext"
 $startupProject = "$repoRoot\Backbone.API"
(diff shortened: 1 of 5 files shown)


tnotheis:
Enhancements are only things that deliver value to the user. This task though did only simplify the life of developers.

scripts/windows/efcore/add_migration.ps1
```diff
@@ -19,7 +18,7 @@ function AddMigration {
 
     $migrationProject = "$repoRoot\Modules\$moduleName\src\$moduleName.Infrastructure.Database.$provider"
 
-    $cmd = "dotnet ef migrations add --startup-project '$startupProject' --project '$migrationProject' --context $dbContextName --output-dir Migrations --verbose $migrationName"
+    $cmd = "dotnet ef migrations add --startup-project '$startupProject' --project '$migrationProject' --context $dbContextName --output-dir Migrations --verbose $migrationName -- --environment $environment"
```
tnotheis, line 21: Why don't you set ASPNETCORE_ENVIRONMENT anymore? I like your new approach more - but only if it actually works. :) I guess you tried it out?

daniel-almeida-konkconsulting: I tried it out for all scripts and both databases, everything works properly.
Regarding ASPNETCORE_ENVIRONMENT, I found the success rate a bit wonky, which was really odd.
I was even able to find multiple threads with people reporting the same results (sometimes the dotnet ef migrations command would default to the appsetting.json file).

tnotheis (approved):

PR #111, opened 2023-05-05
Repository for Challenges module

filipemmagalhaes:
# Readiness checklist

- [x] I labeled the PR.


Modules/Challenges/src/Challenges.Infrastructure/Persistence/Database/Repository/ChallengesRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+            .ToListAsync(cancellationToken);
+
+        return idsOfExpiredChallenges;
+    }
+
+    public async Task DeleteExpiredChallenges(List<ChallengeId> idsOfExpiredChallenges, CancellationToken cancellationToken)
+    {
+        _challenges.RemoveRange(idsOfExpiredChallenges.Select(id => new Challenge { Id = id }));
+        await _dbContext.SaveChangesAsync(cancellationToken);
+    }
```
tnotheis, line 47:
```suggestion
    public async Task DeleteExpiredChallenges(CancellationToken cancellationToken)
    {
        await Challenges.Where(Challenge.CanBeCleanedUp).ExecuteDeleteAsync();
    }
```
You can also delete the `FindExpiredChallenges` method, since it is not needed anymore with the new implementation of the Delete method.

Modules/Challenges/src/Challenges.Infrastructure/Persistence/Database/Repository/ChallengesRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+    
+    public ChallengesRepository(ChallengesDbContext dbContext)
+    {
+        _challenges = dbContext.Challenges;
+        _dbContext = dbContext;
+    }
+    public async Task<Challenge> Find(ChallengeId id, CancellationToken cancellationToken)
+    {
+        return await _challenges
+            .NotExpired()
```
tnotheis, line 21: I think you should always return the Challenge, whether it is expired or not. In the GetChallengeByIdCommand Handler, you should then check if the Challenge is expired. Do this by adding an `IsExpired` method to the `Challenge` class. If the Challenge is expired, throw a NotFoundException.
The disadvantage of this approach is, that the database needs to return the data of the Challenge even though it isn't needed. But the advantage is, that the logic of "don't return the Challenge if it is expired" stays in the Application logic, and is not moved to the Infrastructure part.

Modules/Challenges/src/Challenges.Infrastructure/Persistence/Database/Repository/ChallengesRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+            .FirstWithId(id, cancellationToken);
+    }
+
+    public async Task<Challenge> Add(Challenge challenge, CancellationToken cancellationToken)
```
tnotheis, line 25: Don't return the Challenge. This is not needed.

(I know you copied that from the `MessagesRepository`, but it was wrong there as well and will be fixed with the next PR)

Modules/Challenges/src/Challenges.Infrastructure/Persistence/IServiceCollectionExtensions.cs
```diff
@@ -46,6 +48,7 @@ public static void AddDatabase(this IServiceCollection services, Action<DbOption
         }
 
         services.AddScoped<IChallengesDbContext, ChallengesDbContext>();
```
tnotheis, line 50: Delete the now unused `IChallengesDbContext` interface.

tnotheis (changes requested):

Modules/Challenges/src/Challenges.Application/Challenges/Commands/DeleteExpiredChallenges/Handler.cs
```diff
@@ -17,25 +17,19 @@ public Handler(ILogger<DeleteExpiredChallengesCommand> logger, IChallengesReposi
 
     public async Task<DeleteExpiredChallengesResponse> Handle(DeleteExpiredChallengesCommand request, CancellationToken cancellationToken)
     {
-        var idsOfExpiredChallenges = await _challengesRepository.FindExpiredChallenges(cancellationToken);
+        var challenges = await _challengesRepository.FindAll(cancellationToken);
```
tnotheis, line 20: This reads **all** Challanges from the database, which can potentially be a huge amount of data, which could lead to a dump of the application.

I guess you did this because I said that you can delete the `FindExpiredChallenges` method, but you still needed the Challanges to build the Response. But as a developer all your alarm bells should go off when you see a statement like this. And if that happens (and I really hope it **did** happen), please tell me and don't just implement it.
When I said that you can delete the `FindExpiredChallenges` method, I didn't have in mind that the response needed the Challenges after deletion.
I'd suggest to change the `ChallengesRepository.DeleteExpiredChallenges` method so that it returns the number of deleted Challenges. Then instead of returning the ids from the Handler, we just return the number of deleted items. I don't think it's necessary to have the ids anyway.

tnotheis (changes requested):

Modules/Challenges/src/Challenges.Infrastructure/Persistence/IServiceCollectionExtensions.cs
tnotheis: Not just the registration of the the interface, but also the interface itself. 😑

Modules/Challenges/src/Challenges.Application/Challenges/Commands/DeleteExpiredChallenges/DeleteExpiredChallengesResponse.cs
```diff
@@ -1,18 +1,16 @@
...
 public class DeleteExpiredChallengesResponse
 {
-    public DeleteExpiredChallengesResponse(IEnumerable<ChallengeId> deletedChallenges)
+    public DeleteExpiredChallengesResponse(int deletedChallenges)
     {
         DeletedChallenges = deletedChallenges;
     }
 
-    public IEnumerable<ChallengeId> DeletedChallenges { get; set; }
+    public int DeletedChallenges { get; set; }
```
tnotheis, line 10:
```suggestion
    public int NumberOfDeletedChallenges { get; set; }
```

tnotheis (changes requested):

tnotheis (approved):

PR #119, opened 2023-05-08
Restructure integration tests

daniel-almeida-konkconsulting:
# Readiness checklist

- [X] I added/updated integration tests.
- [X] I labeled the PR.


AdminApi.Tests.Integration/AdminApi.Tests.Integration.csproj
tnotheis, line 1: This file looks messed up. Is all this `*.feature` stuff necessary? The old projects didn't have something like that.
(maybe this code enables some features of the IDE, so do some research before you delete them)

AdminApi.Tests.Integration/Utils/Support/Dependencies.cs
tnotheis, line 1: Isn't it sufficient to have one Dependencies.cs file?

ConsumerApi.Tests.Integration/Challenges/StepDefinitions/ChallengesApiStepDefinitions.cs
tnotheis, line 1: It's confusing that the step definitions for the Admin API are grouped by endpoint (e.g. GetIdentityListStepDefinitions.cs), while the step definitions for the Consumer API are grouped by "route group" (e.g. ChallengesApiStepDefinitions.cs). You should decide for one. I would probably go with the latter.

AdminApi.Tests.Integration/Utils/StepDefinitions/BaseStepDefinitions.cs
tnotheis, line 1: I don't understand why the base class needs those step definitions. Isn't it easier to have a "CommonStepDefinitions" file, which simply doesn't define a scope, so that the step definitions are available in all feature files?

ConsumerApi.Tests.Integration/Challenges/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    [Given(@"a Challenge c")]
+    public async Task GivenAChallengeC()
+    {
+        var challengeResponse = await _challengeApi.CreateChallenge(_requestConfiguration);
+        challengeResponse.IsSuccessStatusCode.Should().BeTrue();
+
+        _challengeId = challengeResponse.Content!.Result!.Id;
+        _challengeId.Should().NotBeNullOrEmpty(because: "Required value for 'Id' is missing.");
```
tnotheis, line 32: This is the only line where a `because` is defined. You should remove it to stay consistent

ConsumerApi.Tests.Integration/Utils/Models/Error.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace ConsumerApi.Tests.Integration.Utils.Models;
+
+public class Error
+{
+    public string Id { get; set; } = string.Empty;
+    public string Code { get; set; } = string.Empty;
+    public string Message { get; set; } = string.Empty;
+    public string Docs { get; set; } = string.Empty;
```
tnotheis, line 8: Remove default values

ConsumerApi.Tests.Integration/Utils/Support/Dependencies.cs
tnotheis, line 1: Isn't it sufficient to have one Dependencies.cs file?

tnotheis (changes requested):
I don't think that I would split the tests per module. Admin API and Consumer API should imo be considered as two independent products. They shouldn't care about modules. And if you go with the approach I suggested via chat yesterday (making subfolders representing the routes), a developer usually knows what module is actually tested in the feature file.

ConsumerApi.Tests.Integration/Challenges/StepDefinitions/ChallengesApiStepDefinitions.cs
daniel-almeida-konkconsulting: Grouped them by "route group".

tnotheis (changes requested):

ConsumerApi.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -140,9 +109,9 @@ public async Task WhenAGETRequestIsSentToTheTokensEndpointWithAListOfIdsOfOwnTok
     {
         var tokenIds = _givenOwnTokens.Select(t => t.Id);
 
-        var response = await _tokensApi.GetTokenByIds(_requestConfiguration, tokenIds);
+        var response = await _tokensApi.GetTokensByIds(_requestConfiguration, tokenIds);
 
-        _tokenResponse.StatusCode = response.StatusCode;
+        _response.StatusCode = response.StatusCode;
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: I didn't change any test logic, all the code you see was previously created and submitted in separate PRs.
That being said, I do agree it makes no sense.
I checked other instances of this and fixed them.

ConsumerApi.Tests.Integration/API/BaseApi.cs
```diff
@@ -18,9 +19,36 @@ protected BaseApi(RestClient client)
                 (sender, cert, chain, sslPolicyErrors) => true;
     }
 
-    protected async Task<HttpResponse<T>> ExecuteRequest<T>(Method method, string endpoint, RequestConfiguration requestConfiguration)
+    protected async Task<HttpResponse<T>> Get<T>(string endpoint, RequestConfiguration requestConfiguration)
```
tnotheis, line 22: I would have kept the ExecuteRequest method, made it private, and called it from the `Get` and `Post` methods.

ConsumerApi.Tests.Integration/API/ChallengesApi.cs
```diff
@@ -10,11 +9,11 @@ public class ChallengesApi : BaseApi
 
     public async Task<HttpResponse<Challenge>> CreateChallenge(RequestConfiguration requestConfiguration)
     {
-        return await ExecuteRequest<Challenge>(Method.Post, new PathString(ROUTE_PREFIX).Add("/challenges").ToString(), requestConfiguration);
+        return await Post<Challenge>("/challenges", requestConfiguration);
```
tnotheis, line 12: Just noticed that the route parts are lowercase in the tests. But in the Backbone, they have an uppercase first latter (=> "/Challenges"). Can you do a quick scan of the tests and fix that please?

ConsumerApi.Tests.Integration/StepDefinitions/BaseStepDefinitions.cs
```diff
@@ -28,56 +28,56 @@ protected BaseStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, Htt
...
     [Then(@"the response content includes an error with the error code ""([^""]+)""")]
-    protected void ThenTheResponseContentIncludesAnErrorWithTheErrorCode(string errorCode)
+    public void ThenTheResponseContentIncludesAnErrorWithTheErrorCode(string errorCode)
     {
         _response.Content!.Error.Should().NotBeNull();
         _response.Content!.Error!.Code.Should().Be(errorCode);
     }
 
-    protected void AssertResponseIntegrity()
+    public void AssertResponseIntegrity()
```
tnotheis, line 62: Now you made the helper methods public as well. 😆 We only talked about the step definitions. :) But it's okay

ConsumerApi.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -111,12 +111,11 @@ public async Task WhenAGETRequestIsSentToTheTokensEndpointWithAListOfIdsOfOwnTok
 
         var response = await _tokensApi.GetTokensByIds(_requestConfiguration, tokenIds);
 
-        _response.StatusCode = response.StatusCode;
-
         var tokens = response.Content!.Result!;
         tokens.Should().NotBeNull();
         tokens.Should().HaveCount(_givenOwnTokens.Count);
 
+        _response.StatusCode = response.StatusCode;
```
tnotheis, line 118: Save the entire response if possible

ConsumerApi.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -216,17 +184,16 @@ public async Task WhenAPOSTRequestIsSentToTheTokensEndpointWith(string content,
...
 
     [When(@"a GET request is sent to the Tokens endpoint with a list containing t\.Id, p\.Id")]
     public async Task WhenAGETRequestIsSentToTheTokensEndpointWithAListContainingT_IdP_Id()
     {
         var tokenIds = new List<string> { _tokenId, _peerTokenId };
-        var response = await _tokensApi.GetTokenByIds(_requestConfiguration, tokenIds);
+        var response = await _tokensApi.GetTokensByIds(_requestConfiguration, tokenIds);
 
-        _tokenResponse.StatusCode = response.StatusCode;
+        _response.StatusCode = response.StatusCode;
```
tnotheis, line 196: Save the entire response if possible

tnotheis (changes requested):

AdminApi.Tests.Integration/Extensions/HttpResponseExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        response.IsSuccessStatusCode.Should().BeTrue();
+    }
+
+    public static void AssertResponseContentTypeIs<T>(this HttpResponse<T> response, string contentType)
+    {
+        response.ContentType.Should().Be(contentType);
+    }
+
+    public static void AssertResponseContentCompliesWithSchema<T>(this HttpResponse<T> response)
```
tnotheis, line 23:
```suggestion
    public static void AssertContentCompliesWithSchema<T>(this HttpResponse<T> response)
```

AdminApi.Tests.Integration/StepDefinitions/IdentitiesApiStepDefinitions.cs
```diff
@@ -5,16 +5,17 @@ namespace AdminApi.Tests.Integration.StepDefinitions;
...
+public class IdentitiesApiStepDefinitions : BaseStepDefinitions
 {
     private readonly IdentitiesApi _identitiesApi;
+    private HttpResponse<List<IdentitySummaryDTO>> _response;
 
-    public IdentitiesApiStepDefinitions(IdentitiesApi identitiesApi) : base(new HttpResponse<List<IdentitySummaryDTO>>())
+    public IdentitiesApiStepDefinitions(IdentitiesApi identitiesApi) : base()
     {
         _identitiesApi = identitiesApi;
+        _response = new HttpResponse<List<IdentitySummaryDTO>>();
```
tnotheis, line 16: We shouldn't initialize the `_response` from the constructor. Doesn't make sense.
Diogo probably did it to avoid the warning the compiler gives about an uninitialized field. But just setting it to a plain object is a lie. 
You should rather make `_response` optional and use `?`/`!` wherever it is used.

ConsumerApi.Tests.Integration/Models/Token.cs
```diff
@@ -13,4 +13,24 @@ public class Token
...
+
+    public bool ShouldSerializeCreatedByDevice()
+    {
+        return CreatedByDevice != null;
+    }
+
+    public bool ShouldSerializeContent()
+    {
+        return Content != null;
+    }
```
tnotheis, line 35: These aren't used anywhere. And I don't understand what they should be used for anyway.

tnotheis (changes requested):

ConsumerApi.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -221,14 +221,40 @@ public void ThenTheResponseContainsAllTokensWithTheGivenIds()
...
+
+        //if (_tokenResponse.HttpMethod == Method.Get.ToString())
+        //{
+        //_tokenResponse.AssertResponseContentCompliesWithSchema<Token>();
+        //}
+        //else if (_tokenResponse.HttpMethod == Method.Post.ToString())
+        //{
+        //    _tokenResponse.AssertResponseContentCompliesWithSchema<CreateTokenResponse>();
+        //    AssertResponseContentCompliesWithSchema<CreateTokenResponse>();
+        //}
```
tnotheis, line 237: ?

ConsumerApi.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -1,34 +1,39 @@
...
-    public TokensApiStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, TokensApi tokensApi) : base(httpConfiguration, new HttpResponse<Token>())
+    public TokensApiStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, TokensApi tokensApi) : base(httpConfiguration)
     {
         _tokensApi = tokensApi;
         _tokenId = string.Empty;
         _peerTokenId = string.Empty;
         _givenOwnTokens = new List<Token>();
         _responseTokens = new List<Token>();
+        _tokenResponse = new HttpResponse<Token>();
+        _tokensResponse = new HttpResponse<IEnumerable<Token>>();
```
tnotheis: See above (don't initialize them as empty responses)

ConsumerApi.Tests.Integration/Models/Token.cs
daniel-almeida-konkconsulting: These are used behind the scenes for the Schema validations, they are a pattern of Newtonsoft.JSON.
This way, I can use the Token class to validate both POST and GET responses, as opposed to having the CreateTokenResponse that we had before.

tnotheis (approved):

tnotheis (approved):

PR #120, opened 2023-05-08
Repository for files module

Dannyps:
# Readiness checklist

- [x] I labeled the PR.


Modules/Files/src/Files.Application/Extensions/FileMetadataQueryableExtensions.cs
tnotheis, line 1: You should move this file to the Infrastructure project, because it's only needed there. Could you tell the others to do the same on their tasks, please? And could you do it for the already refactored modules as well?

Modules/Files/src/Files.Domain/Entities/FileMetadata.cs
```diff
@@ -45,6 +45,17 @@ public FileMetadata(IdentityAddress createdBy, DeviceId createdByDevice, Identit
     public long CipherSize { get; set; }
     public byte[] CipherHash { get; set; }
 
+    public byte[]? Content { get; private set; }
+
+    public void LoadContent(byte[] bytes)
```
tnotheis, line 50: You should call the parameter according to the property (=> `content`)

Modules/Files/src/Files.Domain/Entities/FileMetadata.cs
tnotheis, line 1: Since the class now contains the content, it's not just `FileMetadata` anymore. Instead, it's just a `File`.

Modules/Files/src/Files.Domain/Entities/FileMetadata.cs
```diff
@@ -45,6 +45,17 @@ public FileMetadata(IdentityAddress createdBy, DeviceId createdByDevice, Identit
     public long CipherSize { get; set; }
     public byte[] CipherHash { get; set; }
 
+    public byte[]? Content { get; private set; }
```
tnotheis, line 48: The `Content` is theoretically not nullable. The only case it could be null is when it's not loaded. But I think we shouldn't leak this persistence detail to the domain model. 
So better remove the `?`. I don't like that, but I think I'd still prefer it compared to the alternative.

Modules/Files/src/Files.Infrastructure/Persistence/Database/EntityTypeConfigurations/FileMetadataEntityTypeConfiguration.cs
```diff
@@ -11,5 +11,7 @@ public void Configure(EntityTypeBuilder<FileMetadata> builder)
         builder.HasIndex(m => m.CreatedBy);
 
         builder.Property(m => m.CipherHash).IsRequired();
+
+        builder.Ignore(m => m.Content);
```
tnotheis, line 15: When you modify an EntityTypeConfiguration, you should always create a migration. This just creates empty `Up` and `Down` methods, but the DbContextModelSnapshot should update.

Modules/Files/src/Files.Infrastructure/Persistence/Database/Repository/FilesRepository.cs
```diff
@@ -0,0 +1,70 @@
...
+    public FilesRepository(FilesDbContext dbContext, IBlobStorage blobStorage, IOptions<BlobOptions> blobOptions)
+    {
+        _files = dbContext.FileMetadata;
+        _readOnlyFiles = dbContext.FileMetadata.AsNoTracking();
+        _dbContext = dbContext;
+        _blobStorage = blobStorage;
+        _blobOptions = blobOptions.Value;
+    }
+
+    public async Task Add(FileMetadata fileMetadata, byte[] body, CancellationToken cancellationToken)
```
tnotheis, line 31: Can you search for occurrences of `body` in the Files module and replace them with `content` please?

Modules/Files/src/Files.Infrastructure/Persistence/Database/Repository/FilesRepository.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        _blobStorage.Add(_blobOptions.RootFolder, fileMetadata.Id, body);
+        await _files.AddAsync(fileMetadata, cancellationToken);
+        await _blobStorage.SaveAsync();
+        await _dbContext.SaveChangesAsync(cancellationToken);
+
+    }
+
+    public async Task<FileMetadata> Find(FileId fileId, bool track = false, bool fillBody = true)
```
tnotheis, line 40:
```suggestion
    public async Task<FileMetadata> Find(FileId fileId, bool track = false, bool fillContent = true)
```

tnotheis (changes requested):

Modules/Challenges/src/Challenges.Infrastructure/Persistence/Database/QueryableExtensions/ApplicationServiceCollectionExtensions.cs
tnotheis, line 1: Why did you move this file to Infrastructure?

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/QueryableExtensions/MessagesQueryableExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+using Enmeshed.BuildingBlocks.Application.Abstractions.Exceptions;
+using Enmeshed.DevelopmentKit.Identity.ValueObjects;
+using Enmeshed.Tooling;
+using Microsoft.EntityFrameworkCore;
+
+namespace Backbone.Modules.Messages.Infrastructure.Persistence.Database.QueryableExtensions;
+
+public static class MessagesQueryableExtensions
+{
+    public static IQueryable<Message> IncludeAllReferences(this IQueryable<Message> messages)
```
tnotheis, line 12: There is an extension method in the BuildingBlocks.Infrastructure project called `IncludeAll` which includes all references generically. Can you use that and remove this one please (I know it isn't part of your current task)

tnotheis (changes requested):

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/QueryableExtensions/MessagesQueryableExtensions.cs
tnotheis, line 1: Now you've added MessagesQueryableExtensions and RelationshipsQueryableExtensions to the Infrastructure project, but you didn't remove the ones in the Application project. On the contrary: you even still use them, right?

tnotheis (changes requested):

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/QueryableExtensions/MessagesQueryableExtensions.cs
Dannyps: I believe those may be due to merges coming from main. I'll investigate.

tnotheis (approved):

PR #125, opened 2023-05-11
Improve Relationships repository

tnotheis:
# Readiness checklist

- [x] I labeled the PR.


Modules/Relationships/src/Relationships.Infrastructure/Persistence/Database/Repository/RelationshipsRepository.cs
```diff
@@ -110,42 +123,50 @@ public async Task<DbPaginationResult<Relationship>> FindRelationshipsWithIds(IEn
...
+        else if(latestChange.Response?.Content != null)
+            _blobStorage.Add(_blobOptions.RootFolder, $"{latestChange.Id}_Res", latestChange.Response.Content);
 
-    public async Task SaveContentOfChangeResponse(RelationshipChangeResponse changeResponse)
-    {
-        if (changeResponse.Content == null)
-            return;
-
-        _blobStorage.Add(_blobOptions.RootFolder, $"{changeResponse.Id}_Res", changeResponse.Content);
         await _blobStorage.SaveAsync();
```
daniel-almeida-konkconsulting, line 167: Are we no longer catching the exception?
```
try
{
    await _blobStorage.SaveAsync();
}
catch (BlobAlreadyExistsException)
{
    throw new DomainException(DomainErrors.ChangeRequestIsAlreadyCompleted(latestChange.Status));
}
```

daniel-almeida-konkconsulting (changes requested):

Modules/Relationships/src/Relationships.Infrastructure/Persistence/Database/Repository/RelationshipsRepository.cs
```diff
@@ -110,42 +123,50 @@ public async Task<DbPaginationResult<Relationship>> FindRelationshipsWithIds(IEn
...
+                        r.Status != RelationshipStatus.Revoked)
+            .AnyAsync(cancellationToken);
     }
 
-    public async Task SaveContentOfChangeRequest(RelationshipChangeRequest changeRequest)
+    public async Task SaveContentOfLatestChange(Relationship relationship)
     {
-        if (changeRequest.Content == null)
-            return;
+        var latestChange = relationship.Changes.OrderByDescending(c => c.CreatedAt).First();
```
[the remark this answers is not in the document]
tnotheis: 1. Right, I forgot that this method exists. Thanks for the hint.
2. There's always at least one change, yes. The same request that creates a Relationship creates a RelationshipChange of type "Creation". So if this method is called without any change, then it is a bug, and a 500 status code is correct.

Modules/Relationships/src/Relationships.Infrastructure/Persistence/Database/Repository/RelationshipsRepository.cs
tnotheis: Shouldn't be necessary, and actually never was before, because the checks before should avoid this.

But I still catched it to be on the safe side.

daniel-almeida-konkconsulting (approved):

PR #139, opened 2023-05-22
Quotas for tiers domain

daniel-almeida-konkconsulting:
# Readiness checklist

- [X] I added/updated unit tests.
- [X] I labeled the PR.


Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -9,6 +11,12 @@ public Identity(string address, string tierId)
     }
 
     public string Address { get; }
-    
     public string TierId { get; }
+    private readonly List<TierQuota> _tierQuotas = new();
```
tnotheis, line 15: Move the field to the top

Modules/Quotas/test/Quotas.Domain.Tests/IdentityTests.cs
```diff
@@ -0,0 +1,36 @@
...
+using Xunit;
+
+namespace Backbone.Modules.Quotas.Domain.Tests;
+
+public class IdentityTests
+{
+    [Fact]
+    public void Can_create_identity_with_valid_properties()
+    {
+        var address = TestDataGenerator.CreateRandomBytes().ToString();
```
tnotheis, line 14: Using a hardcoded value makes it easier to read, in my opinion. 

```suggestion
        var address = "someAddress"
```

Modules/Quotas/test/Quotas.Domain.Tests/IdentityTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        var identity = new Identity(address, tierId);
+
+        identity.Address.Should().Be(address);
+        identity.TierId.Should().Be(tierId);
+    }
+
+    [Fact]
+    public void Can_assign_tier_quota_from_definition_to_identity()
+    {
+        var address = TestDataGenerator.CreateRandomBytes().ToString();
```
tnotheis, line 25: See above

Modules/Quotas/test/Quotas.Domain.Tests/IdentityTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Backbone.Modules.Quotas.Domain.Tests;
+
+public class IdentityTests
+{
+    [Fact]
+    public void Can_create_identity_with_valid_properties()
+    {
+        var address = TestDataGenerator.CreateRandomBytes().ToString();
+        var tierId = "TIREYSCQI6XaMygco7Bw";
```
tnotheis, line 15: You should always make it obvious that your test data are actually just test data.
In this example:
```suggestion
        var tierId = "TIRsomeTierId1111111";
```

Modules/Quotas/test/Quotas.Domain.Tests/IdentityTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        identity.Address.Should().Be(address);
+        identity.TierId.Should().Be(tierId);
+    }
+
+    [Fact]
+    public void Can_assign_tier_quota_from_definition_to_identity()
+    {
+        var address = TestDataGenerator.CreateRandomBytes().ToString();
+        var tierId = "TIREYSCQI6XaMygco7Bw";
```
tnotheis, line 26: See above

Modules/Quotas/test/Quotas.Domain.Tests/IdentityTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        var identity = new Identity(address, tierId);
+
+        var metric = new Metric();
+        var max = 5;
+        var period = QuotaPeriod.Month;
+        var tierQuotaDefinition = new TierQuotaDefinition(metric, max, period);
+
+        identity.AssignTierQuotaFromDefinition(tierQuotaDefinition);
```
tnotheis, line 34: Maybe you should add a comment to the end of the test that says that no exception means that the test succeeded.

Modules/Quotas/test/Quotas.Domain.Tests/QuotaIdTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        var isQuotaIdValid = quotaId.IsSuccess;
+
+        isQuotaIdValid.Should().BeTrue();
+    }
+
+    [Fact]
+    public void Can_generate_quota_id()
+    {
+        var quotaId = QuotaId.Generate();
+        quotaId.Should().NotBeNull();
```
tnotheis, line 25: It also should have a length of 20 and start with "QUO"

Modules/Quotas/test/Quotas.Domain.Tests/QuotaIdTests.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Backbone.Modules.Quotas.Domain.Tests;
+public class QuotaIdTests
+{
+    [Fact]
+    public void Can_create_quota_id_with_valid_value()
+    {
+        var validQuotaIdPrefix = "QUO";
+        var validIdLengthWithoutPrefix = 17;
+        var validIdValue = validQuotaIdPrefix + TestDataGenerator.GenerateString(validIdLengthWithoutPrefix);
```
tnotheis, line 13: Use a hardcoded value instead

Modules/Quotas/test/Quotas.Domain.Tests/QuotaIdTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        var quotaId = QuotaId.Generate();
+        quotaId.Should().NotBeNull();
+    }
+
+    [Fact]
+    public void Cannot_create_quota_id_with_invalid_id_prefix()
+    {
+        var invalidQuotaIdPrefix = "QQQ";
+        var quotaId = QuotaId.Create(invalidQuotaIdPrefix + TestDataGenerator.GenerateString(17));
```
tnotheis, line 32: Use a hardcoded value instead

Modules/Quotas/test/Quotas.Domain.Tests/QuotaIdTests.cs
```diff
@@ -0,0 +1,75 @@
...
+public class QuotaIdTests
+{
+    [Fact]
+    public void Can_create_quota_id_with_valid_value()
+    {
+        var validQuotaIdPrefix = "QUO";
+        var validIdLengthWithoutPrefix = 17;
+        var validIdValue = validQuotaIdPrefix + TestDataGenerator.GenerateString(validIdLengthWithoutPrefix);
+
+        var quotaId = QuotaId.Create(validIdValue);
```
tnotheis, line 15: `quotaId` is a misleading name, because it actually contains a Result, and not the id

Modules/Quotas/test/Quotas.Domain.Tests/QuotaIdTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        quotaId.Should().NotBeNull();
+    }
+
+    [Fact]
+    public void Cannot_create_quota_id_with_invalid_id_prefix()
+    {
+        var invalidQuotaIdPrefix = "QQQ";
+        var quotaId = QuotaId.Create(invalidQuotaIdPrefix + TestDataGenerator.GenerateString(17));
+
+        var isQuotaIdValid = quotaId.IsFailure;
```
tnotheis, line 34: I once told you that I don't like that you create variables for everything that you assert on. This is an example of a problem that comes as a result. 😉 

```suggestion
        var isQuotaIdInvalid = quotaId.IsFailure;
```

Naming the variables wrong could cause confusion. Further you have to trace back where the variable is coming from when you want to debug a failing test. Whereas

```c#
quotaId.IsSuccess.Should().BeFalse();
```

is self explanatory.

Modules/Quotas/test/Quotas.Domain.Tests/QuotaIdTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        errorCode.Should().Be("error.platform.validation.invalidId");
+        errorMessage.Should().Contain("Id has a length of");
+    }
+
+    [Fact]
+    public void Cannot_create_quota_id_with_invalid_id_characters()
+    {
+        char[] invalidCharacters = { '|', '-', '!' };
+        var validQuotaIdPrefix = "QUO";
+        var quotaIdValue = TestDataGenerator.GenerateString(QuotaId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX, invalidCharacters);
```
tnotheis, line 64: I don't know if we had it like that in the already existing Id tests as well, but I don't like that this test is unpredictable in some cases.

Let's say we accidentally add "|" to the allowed characters of a QuotaId. When the randomly generated Id coincidentally generates an id that doesn't contain "|", the test will succeed. In all other cases, it will fail.

So you should rather create a `Theory`, which accepts an invalid character as a parameter, appends it to a string of valid characters, and tries to create a QuotaId with the result.
Define some invalid characters as `TestData`.

Modules/Quotas/test/Quotas.Domain.Tests/TestDataGenerator.cs
tnotheis, line 1: The TestDataGenerator can be move to the UnitTestTools library, so that we don't need to create it in each test project

Modules/Quotas/test/Quotas.Domain.Tests/TierQuotaDefinitionIdTests.cs
tnotheis, line 1: See comments in QuotaIdTests

Modules/Quotas/test/Quotas.Domain.Tests/TierTests.cs
```diff
@@ -1,20 +1,36 @@
...
-namespace Backbone.Modules.Quotas.Domain.Tests
+namespace Backbone.Modules.Quotas.Domain.Tests;
+
+public class TierTests
 {
-    public class TierTests
+    [Fact]
+    public void Can_create_tier_with_valid_properties()
+    {
+        var tierId = "TIREYSCQI6XaMygco7Bw";
```
tnotheis, line 14: You should always make it obvious that your test data are actually just test data.
In this example:
```suggestion
        var tierId = "TIRsomeTierId1111111";
```

Modules/Quotas/test/Quotas.Domain.Tests/QuotaIdTests.cs
```diff
@@ -8,68 +8,55 @@ public class QuotaIdTests
...
     {
-        char[] invalidCharacters = { '|', '-', '!' };
         var validQuotaIdPrefix = "QUO";
-        var quotaIdValue = TestDataGenerator.GenerateString(QuotaId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX, invalidCharacters);
-        var quotaId = QuotaId.Create(validQuotaIdPrefix + quotaIdValue);
-
-        var isQuotaIdValid = quotaId.IsFailure;
-        var errorCode = quotaId.Error.Code;
-        var errorMessage = quotaId.Error.Message;
+        var quotaIdValue = TestDataGenerator.GenerateString(QuotaId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX - 1);
```
tnotheis, line 55: Here I'd rather use a hardcoded value like "QUO1111111111111111" as well.

Modules/Quotas/test/Quotas.Domain.Tests/TierQuotaDefinitionIdTests.cs
```diff
@@ -13,9 +14,8 @@ public void Can_create_tier_quota_definition_id_with_valid_value()
         var validIdValue = validTierQuotaDefinitionIdPrefix + TestDataGenerator.GenerateString(validIdLengthWithoutPrefix);
 
         var tierQuotaDefinitionId = TierQuotaDefinitionId.Create(validIdValue);
-        var isTierQuotaDefinitionIdValid = tierQuotaDefinitionId.IsSuccess;
 
-        isTierQuotaDefinitionIdValid.Should().BeTrue();
+        tierQuotaDefinitionId.IsSuccess.Should().BeTrue();
```
tnotheis, line 18:
```suggestion
        result.IsSuccess.Should().BeTrue();
```

Check the other occurences as well please (don't forget the QuotaIdTests).

Modules/Quotas/test/Quotas.Domain.Tests/TierQuotaDefinitionIdTests.cs
```diff
@@ -47,29 +43,23 @@ public void Cannot_create_tier_quota_definition_id_with_invalid_id_length()
...
     {
-        char[] invalidCharacters = { '|', '-', '!' };
         var validTierQuotaDefinitionIdPrefix = "TQD";
-        var tierQuotaDefinitionIdValue = TestDataGenerator.GenerateString(TierQuotaDefinitionId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX, invalidCharacters);
-        var tierQuotaDefinitionId = TierQuotaDefinitionId.Create(validTierQuotaDefinitionIdPrefix + tierQuotaDefinitionIdValue);
-
-        var isTierQuotaDefinitionIdValid = tierQuotaDefinitionId.IsFailure;
-        var errorCode = tierQuotaDefinitionId.Error.Code;
-        var errorMessage = tierQuotaDefinitionId.Error.Message;
+        var tierQuotaDefinitionIdValue = TestDataGenerator.GenerateString(TierQuotaDefinitionId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX - 1);
```
tnotheis, line 58: See above

tnotheis (changes requested):

tnotheis (approved):

PR #146, opened 2023-05-26
Number of sent Messages quota: Building Blocks

Dannyps:
# Readiness checklist

- [x] I labeled the PR.


BuildingBlocks/src/BuildingBlocks.Application/Attributes/ApplyQuotasForMetricsAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Enmeshed.BuildingBlocks.Domain.StronglyTypedIds;
+
+namespace Enmeshed.BuildingBlocks.Application.Attributes;
+public class ApplyQuotasForMetricsAttribute : Attribute
```
tnotheis, line 4: Inherit from `ValidationAttribute` instead, override `IsValid` and make sure that the Attribute is applied to a class that inherits from `ICommand`.

tnotheis, line 4: Make sure the Attribute can only be added once per class

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
```diff
@@ -0,0 +1,51 @@
...
+        var attrs = request.GetType().CustomAttributes;  // Reflection.
+
+        // Displaying output.
+        foreach (var attr in attrs)
+        {
+            if (attr.AttributeType == typeof(ApplyQuotasForMetricsAttribute))
+            {
+                var args = attr.ConstructorArguments.Select(it => new MetricKey(it.Value as string)).ToList();
+                var statuses = await _metricStatusesRepository.GetMetricStatuses(_userContext.GetAddress(), args);
+                var expiredStatuses = statuses.Where(it=> it.IsExhaustedUntil < DateTime.Now).ToList();
```
tnotheis, line 33: Use `SystemTime` instead of `DateTime` please. `SystemTime` is a wrapper that allows setting the current time in order to make code that uses time testable.

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        // Using reflection.
+        var attrs = request.GetType().CustomAttributes;  // Reflection.
+
+        // Displaying output.
+        foreach (var attr in attrs)
+        {
+            if (attr.AttributeType == typeof(ApplyQuotasForMetricsAttribute))
+            {
+                var args = attr.ConstructorArguments.Select(it => new MetricKey(it.Value as string)).ToList();
```
tnotheis, line 31: `args` is not a good name

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
```diff
@@ -0,0 +1,51 @@
...
+                var statuses = await _metricStatusesRepository.GetMetricStatuses(_userContext.GetAddress(), args);
+                var expiredStatuses = statuses.Where(it=> it.IsExhaustedUntil < DateTime.Now).ToList();
+                // for each of the arguments, determine if they're exhausted.
+                // if one or more are exhausted, an exception should be thrown.
+                // We're to determine which MetricKey to pass to the expection
+                // DPS: suggestion that we pass the one with the bigger `IsExhaustedUntil` (so that actions submited past that point are more likely to succeed).
```
tnotheis, line 37: Your code should be self explanatory. In other words: If you need a comment to explain what your code does, then your code is not good enough. :)

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Enmeshed.BuildingBlocks.Application.Attributes;
+using Enmeshed.BuildingBlocks.Domain.StronglyTypedIds;
+using Enmeshed.Common.Infrastructure.Persistence.Repository;
+using MediatR;
+using Enmeshed.BuildingBlocks.Application.Abstractions.Exceptions;
+using Enmeshed.BuildingBlocks.Application.Abstractions.Infrastructure.UserContext;
+
+namespace Enmeshed.BuildingBlocks.Application.MediatR;
+public class QuotaEnforcerBehavior<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse>
```
tnotheis, line 9: This class should be tested thoroughly.

BuildingBlocks/src/BuildingBlocks.Domain/StronglyTypedIds/MetricKey.cs
tnotheis, line 1: Move this to BuildingBlocks.Domain. The `StronglyTypedIds` project should only contain tools that simplify the creation of strongly typed ids

BuildingBlocks/src/BuildingBlocks.Domain/MetricStatus.cs
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    public MetricStatus(MetricKey metricKey, DateTime isExhaustedUntil)
+    {
+        MetricKey = metricKey;
+        IsExhaustedUntil = isExhaustedUntil;
+    }
+
+    public MetricKey MetricKey { get; private set; }
+    public DateTime IsExhaustedUntil { get; private set; }
```
tnotheis, line 18:
```suggestion
    public MetricKey MetricKey { get; }
    public DateTime IsExhaustedUntil { get; }
```

BuildingBlocks/src/BuildingBlocks.Domain/MetricStatus.cs
```diff
@@ -0,0 +1,19 @@
...
+    public MetricStatus(MetricKey metricKey)
+    {
+        MetricKey = metricKey;
+    }
+
+    public MetricStatus(MetricKey metricKey, DateTime isExhaustedUntil)
+    {
+        MetricKey = metricKey;
+        IsExhaustedUntil = isExhaustedUntil;
+    }
```
tnotheis, line 15: I assume that you only added these constructors so that you can test your changes as long as the other tasks are not done yet?

Common/src/Common.Infrastructure/IServiceCollectionExtension.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Enmeshed.Common.Infrastructure.Persistence.Repository;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Enmeshed.Common.Infrastructure.Persistence.Context;
+
+namespace Enmeshed.Common.Infrastructure;
+public static class IServiceCollectionExtension
+{
+    public static IServiceCollection AddCommonRepositories(this IServiceCollection services, IConfiguration configuration)
```
tnotheis, line 9: Call it `AddMetricStatusesRepository` instead.

Common/src/Common.Infrastructure/Persistence/Context/DapperContext.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Data;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Configuration;
+
+namespace Enmeshed.Common.Infrastructure.Persistence.Context;
+public class DapperContext
```
tnotheis, line 6: Should be a `MetricStatusesDapperContext`

Common/src/Common.Infrastructure/IServiceCollectionExtension.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        services.AddTransient<IMetricStatusesRepository, MetricStatusesRepository>();
+        services.AddSingleton(new DapperContext(configuration));
```
tnotheis, line 12: Instead of passing the whole configuration, you should use the Options pattern instead.

```suggestion
        services.Configure<MetricStatusesDapperContextOptions>(options => {
            options.ConnectionString = configuration.GetSection("Modules:Quotas:Infrastructure:SqlDatabase");
        })
        services.AddSingleton<DapperContext>();
```

Or even better, add a parameter that allows to pass the options configuration callback to the extension method, so that it is possible to configure it from the ConsumerApi

Common/src/Common.Infrastructure/Persistence/Context/DapperContext.cs
```diff
@@ -0,0 +1,29 @@
...
+        var sqlDatabaseSection = configuration.GetSection("Modules:Quotas:Infrastructure:SqlDatabase");
+
+        switch (sqlDatabaseSection.GetSection("Provider").Value)
+        {
+            case "Postgres":
+            case "SqlServer":
+                break;
+            default:
+                throw new Exception($"Unsupported database provider passed.");
+        }
```
tnotheis, line 21: I think we don't need to validate the database provider here.

tnotheis (changes requested):

Modules/Messages/src/Messages.Application/Messages/Commands/SendMessage/SendMessageCommand.cs
```diff
@@ -1,8 +1,10 @@
-﻿using Enmeshed.DevelopmentKit.Identity.ValueObjects;
+﻿using Enmeshed.BuildingBlocks.Application.Attributes;
+using Enmeshed.DevelopmentKit.Identity.ValueObjects;
 using MediatR;
 
 namespace Backbone.Modules.Messages.Application.Messages.Commands.SendMessage;
 
+[ApplyQuotasForMetrics("NumberOfSentMessages", "AnotherMetricKey")]
```
tnotheis, line 7: Remove "AnotherMetricKey" please

Modules/Quotas/src/Quotas.Application/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -13,9 +10,12 @@ public static class IServiceCollectionExtensions
...
     {
-        // services.AddMediatR(typeof(...)); // needs to be filled as soon as there is the first command/query
-        services.TryAddTransient(typeof(IPipelineBehavior<,>), typeof(LoggingBehavior<,>));
-        services.TryAddTransient(typeof(IPipelineBehavior<,>), typeof(RequestValidationBehavior<,>));
+        /*services.AddMediatR(c => c
+            .RegisterServicesFromAssemblyContaining<...>()
+            .AddOpenBehavior(typeof(LoggingBehavior<,>))
+            .AddOpenBehavior(typeof(RequestValidationBehavior<,>))
+            .AddOpenBehavior(typeof(QuotaEnforcerBehavior<,>))
+        );*/
```
tnotheis, line 18: Why is this commented out?

BuildingBlocks/src/BuildingBlocks.Domain/MetricStatus.cs
Dannyps: Correct. I'll be waiting for new changes to let go of this, i.e., load these from the database directly.

Modules/Quotas/src/Quotas.Application/Extensions/IServiceCollectionExtensions.cs
Dannyps: We cannot call `AddBehavior` whithout calling `RegisterServicesFromAssemblyContaining` first, which we're not doing.

Common/src/Common.Infrastructure/Persistence/Context/MetricStatusesDapperContext.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Data;
+using Microsoft.Data.SqlClient;
+
+namespace Enmeshed.Common.Infrastructure.Persistence.Context;
+public class MetricStatusesDapperContext
+{
+    public string ConnectionString;
+
+    public MetricStatusesDapperContext()
+    {}
```
tnotheis, line 10: Remove the empty constructor

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+using Enmeshed.BuildingBlocks.Application.MediatR;
+using Enmeshed.BuildingBlocks.Domain;
+using FluentAssertions;
+using MediatR;
+using Xunit;
+
+namespace Enmeshed.BuildingBlocks.Application.Tests.Mediatr;
+public class QuotaEnforcerBehaviorTests
+{
+    public QuotaEnforcerBehaviorTests() { }
```
tnotheis, line 12: Remove the empty constructor

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+             new TestCommand(),
+            () => {
+                nextHasRan = true;
+                return Task.FromResult(new IResponse());
+            },
+            CancellationToken.None);
+
+        // Assert
+        await acting.Should().ThrowAsync<QuotaExhaustedException>();
+        nextHasRan.Should().BeFalse();
```
tnotheis, line 42: You're testing two things here. These should be separate tests.
The goal is to have a test only fail for exactly one reason (in this case: either because the exception is not thrown or because `next` wasn't called)

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    public QuotaEnforcerBehaviorTests() { }
+
+    [ApplyQuotasForMetrics("MetricKey")]
+    private class TestCommand : IRequest<IResponse> { }
+
+    private class IResponse { }
+
+    [Fact]
+    public async void Throws_QuotaExhaustedException_OnExhaustedQuota()
```
tnotheis, line 20: I usually insert an underscore between each word (except for cohesive words like `QuotaExhaustedException`). And I would add even more details to the description: `Throws_QuotaExhaustedException_when_at_least_one_metric_is_exhausted`

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            new MetricStatus(new MetricKey("KeyOne"), DateTime.UtcNow.AddDays(1)),
+            expectedMetricStatus
+        });
+
+        var userContextStub = new UserContextStub();
+        var behavior = new QuotaEnforcerBehavior<TestCommand, IResponse>(metricStatusesStubRepository, userContextStub);
+        var nextHasRan = false;
+
+        // Act
+        Func<Task> action = async () => await behavior.Handle(
```
tnotheis, line 60:
```suggestion
        Func<Task> acting = async () => await behavior.Handle(
```

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            new TestCommand(),
+            () => {
+                nextHasRan = true;
+                return Task.FromResult(new IResponse());
+            },
+            CancellationToken.None);
+
+        // Assert
+        var ex = await Assert.ThrowsAsync<QuotaExhaustedException>(action);
+        ex.MetricKey.Should().Be(expectedMetricStatus.MetricKey);
```
tnotheis, line 69: Don't mix the assertion libraries. Stay with FluentAssertions.

```suggestion
        (await acting.Should().ThrowAsync<QuotaExhaustedException>())
            .Which.MetricKey.Should().Be(expectedMetricStatus.MetricKey);
```

To get rid of the `await`, you can create a custom extension method:

```c#
public static class NonGenericAsyncFunctionAssertionsExtensions
{
    public static ExceptionAssertions<TException> ThrowAsync2<TException>(this NonGenericAsyncFunctionAssertions assertions, string because = "",
        params object[] becauseArgs)
        where TException : Exception
    {
        return assertions.ThrowAsync<TException>(because, becauseArgs).WaitAsync(CancellationToken.None).Result;
    }
}
```
(you have to find a good name for `ThrowAsync2`)

Then your call would look like:
```c#
acting.Should().ThrowAsync2<QuotaExhaustedException>().Which.MetricKey.Should().Be(expectedMetricStatus.MetricKey);
```

Common/src/Common.Infrastructure/IServiceCollectionExtension.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Enmeshed.Common.Infrastructure;
+public static class IServiceCollectionExtension
+{
+    public static IServiceCollection AddMetricStatusesRepository(this IServiceCollection services, IConfiguration configuration)
+    {
+        services.AddTransient<IMetricStatusesRepository, MetricStatusesRepository>();
+
+        services.Configure<MetricStatusesDapperContext>(c => {
+            c.ConnectionString = configuration.GetSection("Modules:Quotas:Infrastructure:SqlDatabase:ConnectionString").Value
```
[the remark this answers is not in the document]
Dannyps: What action exactly are we talking about here? I'm not sure what you're referring to.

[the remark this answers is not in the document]
tnotheis:
```c#
    public static IServiceCollection AddMetricStatusesRepository(this IServiceCollection services, Action<MetricStatusesDapperContext> configureOptions)
    {
        services.AddTransient<IMetricStatusesRepository, MetricStatusesRepository>();
        services.Configure<MetricStatusesDapperContext>(configureOptions);
    }
```

And then use it as follows:

```c#
services.AddMetricStatusesRepository(options => options.ConnectionString = configuration.GetSection("Modules:Quotas:Infrastructure:SqlDatabase:ConnectionString").Value);
```

BuildingBlocks/src/BuildingBlocks.Domain/MetricStatus.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    public MetricKey MetricKey { get; }
+    public DateTime IsExhaustedUntil { get; }
```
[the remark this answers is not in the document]
tnotheis: You're right. That's one of the situations in which a comment might be handy. :)

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    private class TestCommand : IRequest<IResponse> { }
+
+    private class IResponse { }
+
+    [Fact]
+    public async void Throws_QuotaExhaustedException_OnExhaustedQuota()
+    {
+        // Arrange
+        var metricStatusesStubRepository = new MetricStatusesStubRepository(new() {
+            new MetricStatus(new MetricKey("KeyOne"), DateTime.UtcNow.AddDays(1))
```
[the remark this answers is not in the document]
Dannyps: Do these MetricKeys make sense? And why should those be private?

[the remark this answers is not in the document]
tnotheis: > Do these MetricKeys make sense?
Why not? :) These are the three cases that can exist, and we should probably test all of them.

> And why should the Statuses be private?
That doesn't make sense, of course. :) They should obviously be public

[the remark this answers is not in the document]
tnotheis: > Do these MetricKeys make sense?
Why not? :) These are the three cases that can exist, and we should probably test all of them.

> And why should the Statuses be private?
That doesn't make sense, of course. :) They should obviously be public 

By the way, I wanted to delete this comment in favor of the one below.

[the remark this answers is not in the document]
Dannyps: > Why not? :) These are the three cases that can exist, and we should probably test all of them.

MetricKeys should be something like `NumberOfSentMessages`, and not be related to the Status they're part of. But these are tests, so I suppose we can ignore that.

Common/src/Common.Infrastructure/IServiceCollectionExtension.cs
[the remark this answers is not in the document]
Dannyps: It seems these functions are case-insensitive.

Common/src/Common.Infrastructure/Persistence/Context/MetricStatusesDapperContext.cs
```diff
@@ -0,0 +1,13 @@
...
+
+    public IDbConnection CreateConnection() => new SqlConnection(ConnectionString);
```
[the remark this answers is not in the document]
Dannyps: It seems that the guide I followed was correct. See https://stackoverflow.com/a/58737963/1469991 for further details.

Nonetheless, I have changed the way this works. We probably won't find out much before we run some tests.

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -1,100 +1,109 @@
...
-    public QuotaEnforcerBehaviorTests() { }
+    private readonly IUserContext _userContextStub;
+
+    public QuotaEnforcerBehaviorTests() {
+        _userContextStub = new UserContextStub();
+    }
 
-    [ApplyQuotasForMetrics("MetricKey")]
-    private class TestCommand : IRequest<IResponse> { }
+    [ApplyQuotasForMetrics("*")]
```
tnotheis, line 21: One purpose of tests is documentation. When I am new to a project, the first thing I do is looking at the tests. 

In case of this test, if I would read this line without knowing anything about the implementation, I would probably assume that "*" means that quotas for ALL metrics should be applied.

That's probably even what you intended to express, because the tests don't care about what you pass as a Metric. They always return the same Metric.

But as I said, someone without further knowledge would make the wrong assumption.

In order to avoid that, I'd suggest a comment that says something like "The Metric key doesn't matter for these tests". Now you think "But he said that comments are bad". Yes, but only comments that describe "What is done". Comments that describe "Why something is done" are totally fine - as long as it is impossible to describe it with code. And now that I'm thinking aobut it: in this case you may even pass a Metric key called "DoesntMatterForThisTest" to avoid the comment.

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -1,100 +1,109 @@
...
     private class IResponse { }
 
     [Fact]
-    public async void Throws_QuotaExhaustedException_OnExhaustedQuota()
+    public async void Throws_QuotaExhaustedException_when_at_least_one_metric_is_exhausted()
     {
         // Arrange
         var metricStatusesStubRepository = new MetricStatusesStubRepository(new() {
-            new MetricStatus(new MetricKey("KeyOne"), DateTime.UtcNow.AddDays(1)) 
+            TestData.MetricStatus.ThatIsExhaustedFor1Day, TestData.MetricStatus.ThatWasExhaustedUntilYesterday
```
tnotheis, line 31: Nice test case. But I think this should be a separate one that is called "Throws_QuotaExhaustedException_with_exhaustion_date_when_multiple_metrics_are_exhausted"

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
```diff
@@ -23,21 +22,21 @@ public async Task<TResponse> Handle(TRequest request, RequestHandlerDelegate<TRe
...
 
-            var expiredStatuses = statuses.Where(it=> it.IsExhaustedUntil > SystemTime.UtcNow).ToList();
+            var exhaustedStatuses = statuses.Where(m => m.IsExhausted).ToList();
             
-            if (expiredStatuses.Any())
+            if (exhaustedStatuses.Any())
             {
-                var mostInTheFuture = expiredStatuses.MaxBy(it => it.IsExhaustedUntil);
-                if (mostInTheFuture != null)
+                var mostInTheFuture = exhaustedStatuses.MaxBy(it => it.IsExhaustedUntil);
```
tnotheis: I just realized: there will ALWAYS be at most one MetricStatus for a given key per Identity. The key of a Metric has a unique constraint.

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
tnotheis, line 1: Here a list of test cases that I would write for this class:
- No MetricStatus for the given Metric key exists
- A MetricStatus exists, which is not exhausted (=> IsExhaustedUntil = null)
- A MetricStatus exists, which was exhausted until yesterday
- A MetricStatus exists, which is exhausted until tomorrow

Since, as written in one of the previous comments, the basic assumption for your implementation was wrong, and you have to rewrite the implementation of the QuotaEnforcerBehavior anyway, I would suggest that you delete the whole implementation and the tests, and that you rewrite both in a TDD-way. Doing TDD from time to time helps you to identify test cases more easily.

Common/src/Common.Infrastructure/IServiceCollectionExtension.cs
```diff
@@ -1,20 +1,14 @@
 ﻿using Enmeshed.Common.Infrastructure.Persistence.Repository;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Enmeshed.Common.Infrastructure.Persistence.Context;
 
 namespace Enmeshed.Common.Infrastructure;
 public static class IServiceCollectionExtension
 {
-    public static IServiceCollection AddMetricStatusesRepository(this IServiceCollection services, IConfiguration configuration)
+    public static IServiceCollection AddMetricStatusesRepository(this IServiceCollection services, Action<MetricStatusesDapperContext> configureOptions)
```
tnotheis, line 8: You're using the MetricStatusesDapperContext as Options. But you should introduce `MetricStatusesRepositoryOptions`, which have a property `ConnectionString`, are configured by the action and then injected into the `MetricStatusesDapperContext`

tnotheis (changes requested):

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -1,100 +1,109 @@
 ﻿using Enmeshed.BuildingBlocks.Application.Abstractions.Exceptions;
+using Enmeshed.BuildingBlocks.Application.Abstractions.Infrastructure.UserContext;
 using Enmeshed.BuildingBlocks.Application.Attributes;
 using Enmeshed.BuildingBlocks.Application.MediatR;
 using Enmeshed.BuildingBlocks.Domain;
 using FluentAssertions;
 using MediatR;
 using Xunit;
 
+using DomainMetricStatus = Enmeshed.BuildingBlocks.Domain.MetricStatus;
```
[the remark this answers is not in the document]
Dannyps: By folder, do you mean the `TestData` subclass?
Because if so, then we'd have things like `TestData.MetricStatuses.ThatIsExhaustedFor1Day` which does not sound that good.

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
Dannyps: I never considered there may be more than one. However, I assumed that if a command is registered with several MetricKeys (e.g. `SendMessageCommand` could be registered with `NumberOfSentMessages` and `RateOfSentMessages`) and two or more are currently echausted, it makes sense to inform the user of the one that is exhausted further in the future, i.e., the datetime after which they'll be able to use the command again. And considering this scenario, this code makes sense as far as I can see. Is my assumption wrong? I don't believe this has been cleared yet.

tnotheis: Ouhh, you're right! I never thought about that scenario, tbh.

In that case, I would probably suggest to modify the QuotaExhaustedException. It should contain a list of exausted `MetricStatus`es. Caution: define a dedicated `ExhaustedMetricStatus` class for that purpose; don't use the domain class. At least for the properties. It's okay to have them for the constructor parameter.

So:

```c#
public class QuotaExhaustedException : ApplicationException
{
    public QuotaExhaustedException(MetricStatus[] exhaustedMetricStatuses)
    {
        ExhaustedMetricStatuses = exhaustedMetricStatuses.Select(...);
    }

    public ExhaustedMetricStatus[] ExhaustedMetricStatuses { get; }
}

public class ExhaustedMetricStatus 
{
    public string MetricKey { get; }
    public DateTime IsExhaustedUntil { get; }
```

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
tnotheis: According to the other comment regarding multiple MetricStatuses, there are of course some more test cases.
So here is an updated list of test cases that would make sense in my opinion:

* No MetricStatuses with the given keys exist
* Exactly one MetricStatus exists, but it is not exhausted (=> IsExhaustedUntil = null)
* Exactly one MetricStatus exists, but it only was exhausted until yesterday
* Multiple MetricStatuses exist, but none is exhausted
* Multiple MetricStatuses exist, and one is exhausted until tomorrow
* Multiple MetricStatuses exist, and more than one is exhausted until tomorrow

I don't know if I forgot anything. But I need to leave now. So you should think about some more cases for yourself.

tnotheis:
@Dannyps  I just updated the target branch of this PR in order to reduce the number of changed files. Otherwise the review is too complex.

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -1,87 +1,194 @@
...
+    /// returns zero results</strong> on the `GetMetricStatuses` Method.
+    /// For this reason, the MetricStatuses passed to the constructor have zero relevance for the test.
+    /// </summary>
+    [Fact]
+    public async void Succeeds_when_none_of_the_metrics_exist_in_the_repository()
+    {
+        // Arrange
+        var metricStatusesStubRepository = new MetricStatusesNoMatchStubRepository(new() {
+            TestData.MetricStatus.ThatWasExhaustedUntilYesterday, TestData.MetricStatus.ThatIsNotExhausted
+        });
```
tnotheis, line 105: Wtf, that's the ugliest piece of code I've ever seen. You pass something to a constructor, which has to exist because it is needed by the base class even though the base class doesn't give the sub class anything relevant. How would anyone understand this code without digging deeper? 😲

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
tnotheis, line 1: The tests are a bit messed up now:
- Be careful regarding the naming. 
    - In some tests you talk about "exhausted **quotas**", and in others you talk about "exhausted **metrics**". We should stay with "metrics". 
    - Keep in mind that they are not **passed** (the tests that mention "quotas" always say "**passed** quotas", so don't just rename the tests to "passed metrics")
- Order: First handle all positive cases (start with the most simple one), then all negative cases (also start with the  most simple one)

tnotheis (changes requested):

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/MetricStatusesStubRepository.cs
```diff
@@ -20,8 +16,18 @@ public MetricStatusesStubRepository(List<MetricStatus>? metricStatuses)
...
     }
 
-    public Task<IEnumerable<MetricStatus>> GetMetricStatuses(IdentityAddress identity, IEnumerable<MetricKey> keys)
+    public virtual Task<IEnumerable<MetricStatus>> GetMetricStatuses(IdentityAddress identity, IEnumerable<MetricKey> keys)
     {
         return Task.FromResult(MetricStatuses.AsEnumerable());
     }
 }
+
+public class MetricStatusesNoMatchStubRepository : MetricStatusesStubRepository
```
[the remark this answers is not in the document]
Dannyps: Nothing, really. We can make them brothers instead

BuildingBlocks/src/BuildingBlocks.API/Mvc/ExceptionFilters/CustomExceptionFilter.cs
```diff
@@ -93,10 +93,7 @@ private HttpError CreateHttpErrorApplicationException(ApplicationException appli
     {
         if (applicationException is QuotaExhaustedException quotaExhautedException)
         {
-            return new Dictionary<string, dynamic>() {
-                { quotaExhautedException.MetricKey.GetType().Name, quotaExhautedException.MetricKey.Value },
-                { quotaExhautedException.IsExhaustedUntil.GetType().Name, quotaExhautedException.IsExhaustedUntil},
-            };
+            return quotaExhautedException.ExhaustedMetricStatuses;
```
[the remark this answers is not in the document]
tnotheis: Sorry, the suggestion in my comment wasn't complete. Should have looked as follows: 

```c# 
            return quotaExhautedException.MetricStatuses.Select(m => {
                MetricKey: mMetricKey,
                IsExhaustedUntil: m.IsExhaustedUntil
            });
```

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -2,202 +2,166 @@
...
         // Arrange
-        var metricStatusesStubRepository = new MetricStatusesStubRepository(new() {
-            TestData.MetricStatus.ThatIsExhaustedFor1Day, TestData.MetricStatus.ThatIsExhaustedFor10Days
-        });
-
-        var behavior = CreateQuotaEnforcerBehavior(metricStatusesStubRepository);
+        var behavior = CreateQuotaEnforcerBehavior(
+            TestData.MetricStatus.ThatIsExhaustedFor1Day,
+            TestData.MetricStatus.ThatWasExhaustedUntilYesterday
+            );
```
tnotheis, line 123: Wrong indent

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -2,202 +2,166 @@
...
+            new TestData.TestCommand(),
+            nextMock.Value,
             CancellationToken.None);
 
         // Assert
         acting.Should().AwaitThrowAsync<QuotaExhaustedException>()
-            .Which.ExhaustedMetricStatuses.All(it =>
-                it.IsExhaustedUntil > DateTime.Now
-            ).Should().BeTrue();
+            .Which.ExhaustedMetricStatuses.First().MetricKey.Should().Be(TestData.MetricStatus.ThatIsExhaustedFor1Day.MetricKey);
```
tnotheis, line 134: You should probably assert that there is only one item in ExhaustedMetricStatuses. Check the length in the other tests as well

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.API/Mvc/ExceptionFilters/CustomExceptionFilter.cs
```diff
@@ -94,13 +94,10 @@ private HttpError CreateHttpErrorApplicationException(ApplicationException appli
...
-            return quotaExhautedException.ExhaustedMetricStatuses.Select(it => 
-                new Dictionary<string, object>
-                {
-                    { "MetricKey", it.MetricKey },
-                    { "IsExhaustedUntil", it.IsExhaustedUntil }
-                }
-            );
+            return quotaExhautedException.ExhaustedMetricStatuses.Select(m => new {
+                m.MetricKey,
+                m.IsExhaustedUntil
```
tnotheis, line 99: Not good. This results in the same problem we would have when returning the exception data directly: when we change the property name in the exception, the returned JSON would automatically change as well, leading to a change in the API.

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -109,8 +109,9 @@ public void Throws_QuotaExhaustedException_when_the_metric_is_still_exhausted()
             CancellationToken.None);
 
         // Assert
-        acting.Should().AwaitThrowAsync<QuotaExhaustedException>()
-            .Which.ExhaustedMetricStatuses.First().MetricKey.Should().Be(TestData.MetricStatus.ThatIsExhaustedFor10Days.MetricKey);
+        var exceptionExhaustedMetrics = acting.Should().AwaitThrowAsync<QuotaExhaustedException>().Which.ExhaustedMetricStatuses;
+        exceptionExhaustedMetrics.First().MetricKey.Should().Be(TestData.MetricStatus.ThatIsExhaustedFor10Days.MetricKey);
+        exceptionExhaustedMetrics.Should().HaveCount(1);
```
tnotheis, line 114: Change order so that we get an assertion error instead of an `InvalidOperationException` caused by `First()`. Same for the other occurrences.

tnotheis (changes requested):

tnotheis (approved):

PR #152, opened 2023-05-31
Admin UI Configurable API URL

JLSRKonk:
# Readiness checklist

- [x] I labeled the PR.

### DIFF
--- a/AdminUI/src/environments/environment.ts
+++ b/AdminUI/src/environments/environment.ts
@@ -1,4 +1,4 @@
 export const environment = {
     production: true,
-    apiUrl: 'http://localhost:5010/api/v1',
+    apiUrl: import.meta.env.ADMIN_API__BASE_URL,
 };
--- a/AdminUI/angular.json
+++ b/AdminUI/angular.json
@@ -11,7 +11,7 @@
             "prefix": "app",
             "architect": {
                 "build": {
-                    "builder": "@angular-devkit/build-angular:browser",
+                    "builder": "@ngx-env/builder:browser",
                     "options": {
                         "outputPath": "dist/admin-ui",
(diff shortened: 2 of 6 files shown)


AdminUI/src/environments/environment.ts
tnotheis, line 1: Can't we delete the `environment.development.json` file now?

tnotheis (changes requested):

AdminUI/src/environments/environment.ts
```diff
@@ -1,4 +1,4 @@
 export const environment = {
     production: true,
-    apiUrl: 'http://localhost:5010/api/v1',
+    apiUrl: import.meta.env.ADMIN_API__BASE_URL,
```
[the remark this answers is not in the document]
JLSRKonk: After adding the package, the env.d.ts had a few excerpts of commented documentation. In that documentation it was said that process.env.... was now deprecated in favor of using import.meta.env....

AdminUI/src/environments/environment.ts
JLSRKonk: We should still need the environment.development.json file for local testing, as it is the default environment, and also sets the production flag to false.

tnotheis: 1. Don't you use docker-compose for local testing?
2. You can surely set production to false at another place as well. 

But okay, since I don't have to develop on the UI, we can leave it as it is. :)

AdminUI/src/environments/environment.ts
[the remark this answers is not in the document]
tnotheis: Okay 👍

tnotheis (approved):

PR #160, opened 2023-06-09
Number of sent messages quota: Quota.Application

Dannyps:
# Readiness checklist

- [x] I labeled the PR.


Modules/Quotas/src/Quotas.Application/Extensions/IdentityMetricExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    private static async Task<IEnumerable<Quota>> GetAppliedQuotasForMetricAsync(string metric, IQuotasRepository quotasRepository, CancellationToken cancellationToken)
+    {
+        var allQuotasOfMetric = await quotasRepository.FindQuotasByMetricKey(metric, cancellationToken);
+        var highestWeight = allQuotasOfMetric.OrderByDescending(q => q.Weight).FirstOrDefault().Weight;
+        var appliedQuotas = allQuotasOfMetric.Where(q => q.Weight == highestWeight && q.isCurrentlyValid()).ToList();
+        return appliedQuotas;
+    }
+
+    private static void Update(this Quota quota, uint newValue)
```
Dannyps, line 43: The sample for this method is rather confusing in regards to what comes from where.
It might be useful for us to review the logic behind it.

Modules/Quotas/src/Quotas.Application/Extensions/IdentityMetricExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+            if(newIsExhaustedUntil == null || periodEndTime > newIsExhaustedUntil)
+            {
+                IsExhaustedUntil = periodEndTime; // where to set this?
+            }
+        }
+        else
+        {
+            IsExhaustedUntil = null;
+        }
+    }
```
Dannyps, line 59: The sample for this method is rather confusing in regards to what comes from where.
It might be useful for us to review the logic behind it.

tnotheis: First of all: this is domain logic. So don't make an extension method out of it.

Regarding the period end time: it can be calculated from the period of the quota. So for example, if I would execute this method on June 12 2023:
- if the period is "Day", then periodEndTime would be 2023-06-12T23:59:59.9999
- if the period is "Month", then periodEndTime would be 2023-06-30T23:59:59.9999
- if the period is "Year", then periodEndTime would be 2023-12-31T23:59:59.9999
- ...

The variable `newIsExhaustedUntil` isn't needed. You can set directly set the `IsExhaustedUntil` property of the Quota. The value depends on whether `newValue` is greater than or equal to `quota.Max`. If it is, then the new exhaustion time is the period end time. If it isn't then it should be set to null.

Dannyps: > First of all: this is domain logic. So don't make an extension method out of it.

Should we put this login in a Mediatr command?

tnotheis: Since it's domain logic, it should be located in the Domain project (-> the Quota class)

Modules/Quotas/src/Quotas.Application/Extensions/IdentityMetricExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+    private static async Task UpdateMetricAsync(this Identity identity, string metric, IMetricCalculator metricCalculator, IQuotasRepository quotasRepository, CancellationToken cancellationToken)
+    {
+        var quotas = await GetAppliedQuotasForMetricAsync(metric, quotasRepository, cancellationToken);
+        var unExhaustedQuotas = quotas.Where(q => q.IsExhaustedUntil is null || q.IsExhaustedUntil > SystemTime.UtcNow);
+        foreach (var quota in unExhaustedQuotas)
+        {
+            var newValue = metricCalculator.CalculateUsage(
+                SystemTime.UtcNow.AddDays(-30),
+                SystemTime.UtcNow,
+                identity.Address);
```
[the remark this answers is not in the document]
tnotheis: No, it's the exact same logic.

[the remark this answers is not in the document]
tnotheis: Yes, it should. Maybe I misinterpreted your question. So to clarify: in case of a period of `Month`, `from` is the beginning of the current month, and `to` is the end of the current month.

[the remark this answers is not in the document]
Dannyps: okay, that makes sense. Thanks you.

Modules/Quotas/test/Quotas.Domain.Tests/QuotaPeriodTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        var quotaPeriod = QuotaPeriod.Day;
+
+        // Act
+        var end = quotaPeriod.CalculateEnd();
+
+        // Assert
+        end.Second.Should().Be(59);
+        end.Minute.Should().Be(59);
+        end.Hour.Should().Be(23);
+        end.Day.Should().Be(01);
```
tnotheis, line 122: How about the following?

```c#
end.Should().Be(DateTime.Parse("2023-01-01T23:59:59"));
```

This could be even simplified by implementing a custom extension method  `Be` for DateTime that accepts a string, so that you can do this:

```c#
end.Should().Be("2023-01-01T23:59:59");
```

I think this improves readability by a lot.

Modules/Quotas/src/Quotas.Application/Infrastructure/Persistence/Repository/IIdentitiesRepository.cs
```diff
@@ -7,4 +7,5 @@ public interface IIdentitiesRepository
     Task Add(Identity identity, CancellationToken cancellationToken);
     Task<IEnumerable<Identity>> FindWithTier(TierId tierId, CancellationToken cancellationToken, bool track = false);
     Task Update(IEnumerable<Identity> identities, CancellationToken cancellationToken);
+    Task<Identity> FindById(string identityId, CancellationToken cancellationToken);
```
tnotheis, line 10:
```suggestion
    Task<Identity> FindByAddress(string identityAddress, CancellationToken cancellationToken);
```

Modules/Quotas/src/Quotas.Application/Infrastructure/Persistence/Repository/IQuotasRepository.cs
```diff
@@ -1,6 +1,8 @@
-﻿namespace Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
+﻿using Backbone.Modules.Quotas.Domain.Aggregates.Identities;
+
+namespace Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
 
 public interface IQuotasRepository
 {
-
+    Task<IEnumerable<Quota>> FindQuotasByMetricKey(string metricKey, CancellationToken cancellationToken);
```
tnotheis, line 7: This method is currently not used anywhere, as far as I can see. And it shouldn't be. In fact, the whole interface should be deleted.

Modules/Quotas/src/Quotas.Application/Metrics/Commands/RecalculateMetricStatuses/Handler.cs
```diff
@@ -0,0 +1,31 @@
...
+    public Handler(IMetricCalculatorFactory metricCalculatorFactory, IIdentitiesRepository identitiesRepository, IQuotasRepository quotasRepository)
+    {
+        _metricCalculatorFactory = metricCalculatorFactory;
+        _identitiesRepository = identitiesRepository;
+        _quotasRepository = quotasRepository;
+    }
+
+    public async Task Handle(RecalculateMetricStatusesCommand command, CancellationToken cancellationToken)
+    {
+        foreach (var identityId in command.Identities)
```
tnotheis, line 22:
```suggestion
        foreach (var identityAddress in command.Identities)
```

Modules/Quotas/src/Quotas.Application/Metrics/Commands/RecalculateMetricStatuses/Handler.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var identity = await _identitiesRepository.FindById(identityId, cancellationToken); 
+            await identity.UpdateMetrics(command.Metrics, _metricCalculatorFactory, _quotasRepository, cancellationToken);
+        }
+
```
tnotheis, line 27: Don't forget to save :)

Modules/Quotas/src/Quotas.Application/Metrics/Commands/RecalculateMetricStatuses/Handler.cs
```diff
@@ -0,0 +1,31 @@
...
+        _identitiesRepository = identitiesRepository;
+        _quotasRepository = quotasRepository;
+    }
+
+    public async Task Handle(RecalculateMetricStatusesCommand command, CancellationToken cancellationToken)
+    {
+        foreach (var identityId in command.Identities)
+        {
+            var identity = await _identitiesRepository.FindById(identityId, cancellationToken); 
+            await identity.UpdateMetrics(command.Metrics, _metricCalculatorFactory, _quotasRepository, cancellationToken);
```
tnotheis, line 25: This is very inefficient. You should read all Identities first, then update all of them one by one, and finally save all of them at once.

I think you should verify this with a test so that nobody has the idea to change this in the future (use FakeItEasy-mocks to verify that the methods are called exactly X times)

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Quota.cs
```diff
@@ -14,9 +16,29 @@ protected Quota(string applyTo)
...
     public QuotaId Id { get; }
     public string ApplyTo { get; }
-    public DateTime? IsExhaustedUntil { get; }
+    public DateTime? IsExhaustedUntil { get; private set; }
     public abstract int Weight { get; }
     public abstract MetricKey MetricKey { get; }
     public abstract int Max { get; }
     public abstract QuotaPeriod Period { get; }
+    public DateTime PeriodEndTime => Period.CalculateEnd();
+    public DateTime PeriodBeginTime => Period.CalculateBegin();
```
tnotheis, line 25: Why do these exist? It isn't that hard to call `Period.CalculateEnd()` :)

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Quota.cs
```diff
@@ -14,9 +16,29 @@ protected Quota(string applyTo)
...
+
+    // TODO: What is the definition of a Valid quota?
+    public bool IsCurrentlyValid()
```
tnotheis, line 28: This method shouldn't currently be needed. Only when we have the `ValidFrom` and `ValidTo` properties

Modules/Quotas/src/Quotas.Application/Extensions/IdentityMetricExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+                quota.PeriodEndTime,
+                identity.Address);
+
+            quota.UpdateExhaustion(newValue);
+        }
+    }
+
+    private static async Task<IEnumerable<Quota>> GetAppliedQuotasForMetricAsync(string metric, IQuotasRepository quotasRepository, CancellationToken cancellationToken)
+    {
+        var allQuotasOfMetric = await quotasRepository.FindQuotasByMetricKey(metric, cancellationToken);
```
tnotheis, line 37: The identity already has a list of quotas. Why do you read them again?

Modules/Quotas/src/Quotas.Application/Metrics/MetricCalculatorFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public IMetricCalculator CreateFor(string metricKey)
+    {
+        switch (metricKey)
+        {
+            case "NumberOfSentMessages":
+                return _services.GetRequiredService<NumberOfSentMessagesCalculator>();
+
+            default:
+                throw new Exception();
```
tnotheis, line 22:
```suggestion
                throw new NotSupportedException($"There is currently no {nameof(MetricCalculator) for the Metric with the key '{metricKey}'.");
```

Modules/Quotas/src/Quotas.Application/Metrics/NumberOfSentMessagesCalculator.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
+using Backbone.Modules.Quotas.Domain;
+
+namespace Backbone.Modules.Quotas.Application.Metrics;
+public class NumberOfSentMessagesCalculator : IMetricCalculator
```
tnotheis, line 5:
```suggestion
public class NumberOfSentMessagesMetricCalculator : IMetricCalculator
```

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Quota.cs
```diff
@@ -14,9 +16,29 @@ protected Quota(string applyTo)
...
+    {
+        throw new NotImplementedException();
+    }
+
+    public void UpdateExhaustion(uint newValue)
```
tnotheis, line 33: Just found a better name for the parameter: `newUsage`.

Modules/Quotas/test/Quotas.Domain.Tests/QuotaPeriodTests.cs
tnotheis, line 1: You probably would have done that, but don't forget to add tests for `Quota.UpdateExhaustion` as well.

Modules/Quotas/src/Quotas.Infrastructure/Persistence/IServiceCollectionExtensions.cs
```diff
@@ -45,9 +47,14 @@ public static void AddDatabase(this IServiceCollection services, Action<DbOption
                 throw new Exception($"Unsupported database provider: {options.Provider}");
 
         }
-        services.AddTransient<ITiersRepository, TiersRepository>();
         services.AddTransient<IIdentitiesRepository, IdentitiesRepository>();
+        services.AddTransient<IMessagesRepository, MessagesRepository>();
         services.AddTransient<IMetricsRepository, MetricsRepository>();
+        services.AddTransient<IMetricStatusesRepository, MetricStatusesRepository>();
```
tnotheis, line 53: Why does this exist?

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -21,4 +22,43 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
         var tierQuota = new TierQuota(definition, Address);
         _tierQuotas.Add(tierQuota);
     }
+
+    public async Task UpdateMetrics(IEnumerable<string> metrics, IMetricCalculatorFactory factory, CancellationToken cancellationToken)
```
tnotheis, line 26:
```suggestion
    public async Task UpdateMetrics(IEnumerable<MetricKey> metrics, IMetricCalculatorFactory factory, CancellationToken cancellationToken)
```

Modules/Quotas/src/Quotas.Application/Metrics/MetricCalculatorFactory.cs
tnotheis, line 1: Just had an idea. We should instead of having an `IMetricCalculatorFactory` interface define an abstract class ``MetricCalculatorFactory`. This class should look as follows:

```c#
public abstract class MetricCalculatorFactory
{
    public IMetricCalculator CreateFor(MetricKey metricKey)
    {
        switch (metricKey)
        {
            case MetricKey.NumberOfSentMessages: return CreateNumberOfSentMessagesMetricCalculator();
            ...
        }
    }

    public abstract IMetricCalculator CreateNumberOfSentMessagesMetricCalculator();
}
```

And then there is a `ServiceCollectionMetricCalculatorFactory` which implements the abstract methods.

The advantage of this is that not every factory needs to implement the switch statement.

Modules/Quotas/src/Quotas.Infrastructure/Persistence/IServiceCollectionExtensions.cs
```diff
@@ -45,9 +47,14 @@ public static void AddDatabase(this IServiceCollection services, Action<DbOption
...
+        services.AddTransient<ITiersRepository, TiersRepository>();
+
+        services.AddTransient<IMetricCalculatorFactory, MetricCalculatorFactory>();
+        services.AddTransient<NumberOfSentMessagesMetricCalculator>();
```
tnotheis, line 57: Why do you explicitly add this?

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Repository/IdentitiesRepository.cs
```diff
@@ -27,9 +28,17 @@ public async Task Add(Identity identity, CancellationToken cancellationToken)
         await _dbContext.SaveChangesAsync(cancellationToken);
     }
 
-    public async Task<Identity> FindById(string identityId, CancellationToken cancellationToken)
+    public async Task<Identity> FindById(string identityAddress, CancellationToken cancellationToken)
     {
-        return await _identitiesDbSet.FindAsync(identityId, cancellationToken);
+        return await _identitiesDbSet.FirstOrDefaultAsync(id => id.Address == identityAddress, cancellationToken) ?? throw new NotFoundException(identityAddress);
+    }
```
tnotheis: You should add the `track` parameter and handle it

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Repository/IdentitiesRepository.cs
```diff
@@ -27,9 +28,17 @@ public async Task Add(Identity identity, CancellationToken cancellationToken)
         await _dbContext.SaveChangesAsync(cancellationToken);
     }
 
-    public async Task<Identity> FindById(string identityId, CancellationToken cancellationToken)
+    public async Task<Identity> FindById(string identityAddress, CancellationToken cancellationToken)
     {
-        return await _identitiesDbSet.FindAsync(identityId, cancellationToken);
+        return await _identitiesDbSet.FirstOrDefaultAsync(id => id.Address == identityAddress, cancellationToken) ?? throw new NotFoundException(identityAddress);
```
tnotheis, line 33: `id` is a really confusing name.

Further, the NotFoundException expects the name of the entity that wasn't found as a parameter, and not the Id that wasn't found.

```suggestion
        return await _identitiesDbSet.FirstOrDefaultAsync(i => i.Address == identityAddress, cancellationToken) ?? throw new NotFoundException(identityAddress);
```

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Repository/IdentitiesRepository.cs
```diff
@@ -27,9 +28,17 @@ public async Task Add(Identity identity, CancellationToken cancellationToken)
...
+    }
+
+    public async Task<IEnumerable<Identity>> FindByIds(IReadOnlyCollection<string> identityAddresses, CancellationToken cancellationToken)
+    {
+        return await _identitiesDbSet
+            .Where(i => identityAddresses.Contains(i.Address))
+            .Include(i => i.TierQuotas)
```
tnotheis, line 40: Use `IncludeAll`

Modules/Quotas/test/Quotas.Domain.Tests/QuotaPeriodTests.cs
tnotheis, line 1: Don't all of these tests look more or less the same? So why not use two `Theory`s (one for `CalculateBegin`  and one for `CalculateEnd`) instead of all the `Fact`s? They would have two parameters:

- The DateTime the SystemTime should be set to (or maybe rather a formatted date string like the one that is currently used for the assertion so it is easier to read)
- A string with the expected output

Modules/Quotas/test/Quotas.Domain.Tests/QuotaTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        // Assert
+        quota.IsExhaustedUntil.Should().BeNull();
+    }
+
+    [Fact]
+    public void UpdateExhaustion_on_unexhausted_with_newUsage_over_max_quota_sets_IsExhaustedUntil()
+    {
+        // Arrange
+        var currentDate = new DateTime(2023, 01, 01, 13, 45, 00, 000, DateTimeKind.Utc);
+        SystemTime.Set(currentDate);
```
tnotheis, line 30:
```suggestion
        SystemTime.Set(new DateTime(2023, 01, 01, 13, 45, 00, 000, DateTimeKind.Utc));
```

Modules/Quotas/test/Quotas.Domain.Tests/QuotaTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using Enmeshed.Tooling;
+
+namespace Backbone.Modules.Quotas.Domain.Tests;
+public class QuotaTests
+{
+    [Fact]
+    public void UpdateExhaustion_on_unexhausted_with_newUsage_under_max_quota_keeps_IsExhaustedUntil_null()
+    {
+        // Arrange
+        var definition = new TierQuotaDefinition(Aggregates.Metrics.MetricKey.NumberOfFiles, 50, QuotaPeriod.Hour);
```
tnotheis, line 15:
```suggestion
        var definition = new TierQuotaDefinition(MetricKey.NumberOfFiles, 50, QuotaPeriod.Hour);
```

Modules/Quotas/test/Quotas.Domain.Tests/QuotaTests.cs
tnotheis, line 1: You didn't write any tests for the `Identity` class even though that's a very important one.

Further, I misled you a bit by telling you that you should write tests for Quotas.

Instead, you should test the Quota class via the Identity class. What I mean with that is: setup an Identity with certain Quota, and then call `UpdateMetric`. After that, check the state of all Quotas to see if they have the correct `IsExhaustedUntil` date.

Modules/Quotas/test/Quotas.Domain.Tests/QuotaTests.cs
tnotheis, line 15: Why use such high numbers for Max when you can use lower numbers as well? :) So instead of 50 and 20, just use 2 and 1.

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
Dannyps: This method is called from the RecalculateMetricStatusesCommand handler, which receives MetricKeys as a string Enumerable. Hence, this change may not be desired.

Modules/Quotas/src/Quotas.Infrastructure/Persistence/IServiceCollectionExtensions.cs
Dannyps: Autofac requires all of those

Modules/Quotas/src/Quotas.Application/Metrics/MetricCalculatorFactory.cs
tnotheis: But it is. Imagine you add a new MetricKey. What you now need to do is adding a new `case` to each implementation of the factory. And if you forget one of them, and your tests don't cover it, then you receive a runtime error.
With my suggested solution, you would only have to add the abstract method, and the compiler would tell you which implementation don't have the new case yet.

Modules/Quotas/src/Quotas.Infrastructure/Persistence/IServiceCollectionExtensions.cs
tnotheis: The `NumberOfSentMessagesMetricCalculator`? It shouldn't actually, as far as I know. Are you 100% sure?

Modules/Quotas/src/Quotas.Application/Metrics/MetricCalculatorFactory.cs
Dannyps: That makes absolute sense, good catch.

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -21,4 +22,43 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
...
+
+    private async Task UpdateMetricAsync(string metric, IMetricCalculator metricCalculator, CancellationToken cancellationToken)
+    {
+        if (TierQuotas.Count == 0)
+        {
+            return;
+        }
+
+        var quotas = GetAppliedQuotasForMetricAsync(metric, TierQuotas);
+        var unExhaustedQuotas = quotas.Where(q => q.IsExhaustedUntil is null || q.IsExhaustedUntil > SystemTime.UtcNow);
```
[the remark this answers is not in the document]
tnotheis: No need for the comment, I think. If there are any performance problems, we will have to run a detailed analysis anyway.

But as written in the comment, you should remove the `Where` part.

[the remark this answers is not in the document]
Dannyps: I had removed the where when I wrote my last comment.

Modules/Quotas/src/Quotas.Application/Infrastructure/Persistence/Repository/IIdentitiesRepository.cs
```diff
@@ -7,6 +7,5 @@ public interface IIdentitiesRepository
     Task Add(Identity identity, CancellationToken cancellationToken);
     Task<IEnumerable<Identity>> FindWithTier(TierId tierId, CancellationToken cancellationToken, bool track = false);
     Task Update(IEnumerable<Identity> identities, CancellationToken cancellationToken);
-    Task<Identity> FindById(string identityAddress, CancellationToken cancellationToken);
-    Task<IEnumerable<Identity>> FindByIds(IReadOnlyCollection<string> identityAddresses, CancellationToken cancellationToken);
+    Task<IEnumerable<Identity>> FindByAddresses(IReadOnlyCollection<string> identityAddresses, CancellationToken cancellationToken);
```
tnotheis, line 10: Add a `track` parameter

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
```diff
@@ -9,16 +9,21 @@ namespace Enmeshed.BuildingBlocks.Application.MediatR;
 public class QuotaEnforcerBehavior<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse> where TRequest : notnull
 {
     private readonly IMetricStatusesRepository _metricStatusesRepository;
-    private readonly IUserContext _userContext;
+    private readonly IUserContext? _userContext;
 
-    public QuotaEnforcerBehavior(IMetricStatusesRepository metricStatusesRepositories, IUserContext userContext)
+    public QuotaEnforcerBehavior(IMetricStatusesRepository metricStatusesRepositories, IEnumerable<IUserContext> userContexts)
```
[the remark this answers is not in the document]
Dannyps: > Now you can register the `QuotaCheckerImpl` in case of the ConsumerApi.

How should I do this? Or better, how should I prevent this from happening on the AdminAPI? Should I remove the registering from the Quotas module and add it to the ConsumerAPI `Program.cs` file itself?

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -23,42 +31,67 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
...
+    private void UpdateMetricStatus(MetricKey metricKey, IMetricCalculator metricCalculator)
     {
-        foreach (var metric in metrics)
+        var metricStatus = _metricStatuses.Where(m => m.MetricKey == metricKey).FirstOrDefault();
+        if(metricStatus != null)
         {
-            var metricCalculator = factory.CreateFor(metric);
-            await UpdateMetricAsync(metric, metricCalculator, cancellationToken);
+            var quotasOfMetric = GetAppliedQuotasForMetric(metricKey);
+            metricStatus.Update(quotasOfMetric.Max(q => q.IsExhaustedUntil));
```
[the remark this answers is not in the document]
Dannyps: Good catch!

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -23,42 +31,67 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
...
+            var allQuotas = _tierQuotas;// .Concat(_identityQuotas)
+            var allQuotasOfMetric = allQuotas.Where(q => q.MetricKey == metric);
+            var highestWeight = allQuotasOfMetric.OrderByDescending(q => q.Weight).Max(q=>q.Weight);
+            var appliedQuotas = allQuotasOfMetric.Where(q => q.Weight == highestWeight).ToList();
+            return appliedQuotas;
+        }
+        catch (InvalidOperationException)
+        {
+            return Enumerable.Empty<Quota>();
+        }
```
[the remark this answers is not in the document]
Dannyps: Yes. If no quotas are available, we'll be filtering empty lists/arrays until line 88 where we'll call `.Max()` on an empty enumerable, which will throw an `InvalidOperationException`.

BuildingBlocks/src/BuildingBlocks.Application/MediatR/QuotaEnforcerBehavior.cs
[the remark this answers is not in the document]
tnotheis: > Should I remove the registering from the Quotas module and add it to the ConsumerAPI Program.cs file itself?
Currently there is no such registering. But yes, the registering should be in Program.cs. Since it isn't Quota-specific, it wouldn't make sense to have it in the Quotas module.

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -23,42 +31,67 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
...
-            var metricCalculator = factory.CreateFor(metric);
-            await UpdateMetricAsync(metric, metricCalculator, cancellationToken);
+            var quotasOfMetric = GetAppliedQuotasForMetric(metricKey);
+            metricStatus.Update(quotasOfMetric.Max(q => q.IsExhaustedUntil));
         }
-        return;
     }
 
-    private async Task UpdateMetricAsync(string metric, IMetricCalculator metricCalculator, CancellationToken cancellationToken)
+    public void UpdateAllMetricStatuses()
```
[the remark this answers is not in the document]
Dannyps: We're currently using it for tests, and I figure we may need it in the foreseable future.

As for the way I'm using it, it's a mere matter of simplicity. The way I understand the `_metricStatuses` attribute, it should actually be a `Map<MetricKey, MetricStatus>` or `Map<MetricKey, DateTime? (IsExhaustedUntil)>`. That would certainly simplify this.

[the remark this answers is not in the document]
tnotheis: > We're currently using it for tests, and I figure we may need it in the foreseable future.

We will never update just MetricStatuses. It will only happen via `UpdateMetric` or `UpdateAllMetrics`. Plus: having a public method only for a test is a contradiction by itself. :) Because why would you test a method that is not used anywhere in the production code?

> As for the way I'm using it, it's a mere matter of simplicity. The way I understand the _metricStatuses attribute, it should actually be a Map<MetricKey, MetricStatus> or Map<MetricKey, DateTime? (IsExhaustedUntil)>. That would certainly simplify this.

I agree that an Map would be a better fit, but I'm not sure whether EF Core can handle this. But since the only class that accesses the _metricKeys is the `Identity` class, we can always check if the `MetricKey` is already present. But I don't see how a Map would make any difference in the process of updating the MetricStatuses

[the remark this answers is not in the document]
Dannyps: By removing that method, we've destroyed any way to refresh the MetricStatuses for a given Identity when there's no change to quota's usage.

I've been going round and round about this and I just noticed that you mentioned the `UpdateAllMetrics` method. We do not have such a thing, though it does sound terribly similar to the method I just removed. 

So, I suppose the question is should we have a way to refresh the MetricStatuses for a given Identity (a smart thing to do regularly because quotas can become available again), even if we're not using it in our code just yet (but it is needed for [this](https://github.com/nmshd/backbone/pull/160#discussion_r1236636894))?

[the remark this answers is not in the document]
tnotheis: > even if we're not using it

Never write code that you don't use, even when you think that you will use it some day in the future. :) Such code needs to be maintained, and who knows, maybe we some day we decide that we don't even need it at all. Then all the effort for implementation and maintenance was in vain.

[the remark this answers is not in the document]
Dannyps: Premature optimization is the root of all evil, yes. Can I then ditch the test I linked?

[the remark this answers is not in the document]
tnotheis: That's not what I meant. :) I aimed at the UpdateAllMetricStatuses method. 
But still: when I think about it, you can ignore the comment you linked, because an IsExhaustedUntil date in the past is synonymous with a null value. Let's save the database capacity that would be needed for the update.

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -21,4 +22,43 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
...
+        {
+            var metricCalculator = factory.CreateFor(metric);
+            await UpdateMetricAsync(metric, metricCalculator, cancellationToken);
+        }
+        return;
+    }
+
+    private async Task UpdateMetricAsync(string metric, IMetricCalculator metricCalculator, CancellationToken cancellationToken)
+    {
+        if (TierQuotas.Count == 0)
```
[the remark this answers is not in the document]
Dannyps: I was wondering if we could have something like
```c#
public IReadOnlyCollection<Quota> AllQuotas => _tierQuotas.Concat(_identityQuotas);
```

With the relevant casts.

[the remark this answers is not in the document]
tnotheis: That's okay for me. But make it private, not public (at least as long as we don't need it publicly)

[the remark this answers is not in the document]
tnotheis: Though even with this change, the problem that we need to change this property persists: there will be one place that needs to be changed as soon as individual quotas are implemented.

tnotheis (changes requested):

Common/src/Common.Infrastructure/Persistence/EntityConfigurations/MetricStatusEntityTypeConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Enmeshed.BuildingBlocks.Domain;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace Enmeshed.Common.Infrastructure.Persistence.EntityConfigurations;
+public class MetricStatusEntityTypeConfiguration : IEntityTypeConfiguration<MetricStatus>
+{
+    public void Configure(EntityTypeBuilder<MetricStatus> builder)
+    {
+        builder.Property(x => x.MetricKey).HasColumnType("string");
```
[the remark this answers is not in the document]
Dannyps: Do you mean I should make it extend the StronglyTypedId class (or a class similar to it, that I should create)?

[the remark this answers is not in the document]
tnotheis: No, I mean something like this:

```c#
configurationBuilder.Properties<RelationshipId>().AreUnicode(false).AreFixedLength().HaveMaxLength(RelationshipId.MAX_LENGTH).HaveConversion<RelationshipIdEntityFrameworkValueConverter>();
```

(see Backbone.Modules.Messages.Infrastructure.Persistence.Database.MessagesDbContext)

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -23,42 +34,61 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
...
+            await UpdateMetric(metric, metricCalculator, cancellationToken);
         }
-        return;
     }
 
-    private async Task UpdateMetricAsync(string metric, IMetricCalculator metricCalculator, CancellationToken cancellationToken)
+    private void UpdateMetricStatus(MetricKey metricKey, IEnumerable<Quota> quotasFromMetric)
     {
-        if (TierQuotas.Count == 0)
+        var metricStatus = _metricStatuses.FirstOrDefault(m => m.MetricKey == metricKey);
```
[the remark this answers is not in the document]
Dannyps: Makes sense. We're not addressing the MetricStatus entity in regard to its database representation and that has become an issue because the application will not run properly. I have invested quite some time in creating a migration but I keep facing EF errors. Anyhow, it seems to me that it is unclear how we're to store this on the database. To me, it would make sense that there would be a composite key with the Identity (Address) and the MetricKey. Is that how we'll proceed?

Sorry for highjacking the post, but this matter has been bugging me.

AdminApi/Program.cs
```diff
@@ -52,8 +55,14 @@ static void ConfigureServices(IServiceCollection services, IConfiguration config
     .AddCustomSwaggerWithUi()
     .AddDevices(parsedConfiguration.Modules.Devices)
     .AddQuotas(parsedConfiguration.Modules.Quotas)
+    .AddMetricStatusesRepository(c =>
+    {
+        c.ConnectionString = configuration.GetSection("Modules:Quotas:Infrastructure:SqlDatabase:ConnectionString").Value;
+    })
```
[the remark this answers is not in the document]
Dannyps: Not anymore, no!

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -23,42 +34,61 @@ public void AssignTierQuotaFromDefinition(TierQuotaDefinition definition)
...
+        var highestWeight = -1;
+
+        try
+        {
+            highestWeight = allQuotasOfMetric.Max(q => q.Weight);
+        }
+        catch (InvalidOperationException)
+        {
+            return Enumerable.Empty<Quota>();
+        }
```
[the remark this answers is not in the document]
Dannyps: Thanks 🤪

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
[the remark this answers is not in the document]
tnotheis: > We're not addressing the MetricStatus entity in regard to its database representation

We're not? But that should be part of this PR.

> To me, it would make sense that there would be a composite key with the Identity (Address) and the MetricKey.
Yes, that's what it should look like.

Common/src/Common.Infrastructure/Persistence/EntityConfigurations/MetricStatusEntityTypeConfiguration.cs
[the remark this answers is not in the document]
Dannyps: Thinking more about this, it doesn't really make a lot of sense, does it? We're using Dapper to access this entity, and it is defined in the `BuildingBlocks.Domain` project. I think I added this but it really makes no sense.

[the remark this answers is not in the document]
tnotheis: You read it with Dapper, yes. But you write it with EF Core.

Modules/Quotas/src/Quotas.Application/Metrics/ServiceProviderMetricCalculatorFactory.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    public ServiceProviderMetricCalculatorFactory(IServiceProvider serviceProvider)
+    {
+        _serviceProvider = serviceProvider;
+    }
+
+    public override IMetricCalculator CreateNumberOfSentMessagesMetricCalculator()
+    {
+        var messagesRepository = _serviceProvider.GetService<IMessagesRepository>();
+        return new NumberOfSentMessagesMetricCalculator(messagesRepository);
```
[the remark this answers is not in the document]
Dannyps: Like this a46d6d4b214cc9fa142544d8ce3f16e36ca019e?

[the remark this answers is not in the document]
tnotheis: Yes. But is it necessary to register the NumberOfSentMessagesMetricCalculator explicitly in order to do this?

Common/src/Common.Infrastructure/Persistence/EntityConfigurations/MetricStatusEntityTypeConfiguration.cs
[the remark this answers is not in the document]
Dannyps: We read this with Dapper:
`backbone\BuildingBlocks\src\BuildingBlocks.Domain\MetricStatus.cs`

And write this with EF:
`backbone\Modules\Quotas\src\Quotas.Domain\Aggregates\Identities\MetricStatus.cs`

So any EF configuration should only apply to the latter, which was not the case here. Does this make sense?

Modules/Quotas/src/Quotas.Application/Metrics/ServiceProviderMetricCalculatorFactory.cs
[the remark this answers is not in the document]
Dannyps: How should it work if we do not register it first?
(I'm unable to test it right now)

Modules/Quotas/test/Quotas.Application.Tests/Tests/QuotaCheck/QuotaCheckerImplTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+internal class UserContextStub : IUserContext
+{
+    public UserContextStub()
+    {
+    }
+
+    public IdentityAddress GetAddress()
+    {
+        return IdentityAddress.Create(Convert.FromBase64String("mJGmNbxiVZAPToRuk9O3NvdfsWl6V+7wzIc+/57bU08="), "id1");
```
tnotheis, line 73:
```suggestion
        return IdentityAddress.Create(Convert.FromBase64String(new byte[]{0}), "id1");
```

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -75,55 +31,11 @@ public async void Succeeds_when_none_of_the_metrics_exist_in_the_repository()
...
-
     [Fact]
     public void Throws_QuotaExhaustedException_when_exactly_one_metric_is_exhausted()
     {
         // Arrange
-        var behavior = CreateQuotaEnforcerBehavior(
-            TestData.MetricStatus.ThatIsExhaustedFor1Day,
-            TestData.MetricStatus.ThatWasExhaustedUntilYesterday
-        );
+        var behavior = CreateQuotaEnforcerBehavior(TestData.MetricStatus.ThatIsExhaustedFor1Day);
```
tnotheis, line 38: It's hard to understand that the parameters of this method are the exhausted metric statuses. I would rather use a named parameter:

```csharp
var behavior = CreateQuotaEnforcerBehavior(exhaustedMetricStatus = TestData.MetricStatus.ThatIsExhaustedFor1Day);
```

And for multiple exhausted metrics statuses, you could have a second method:

```csharp
var behavior = CreateQuotaEnforcerBehavior(exhaustedMetricStatuses = new [] { TestData.MetricStatus.ThatIsExhaustedFor1Day, TestData.MetricStatus.ThatIsExhaustedFor10Days });
```

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -161,9 +75,22 @@ public void Throws_QuotaExhaustedException_when_more_than_one_metric_is_exhauste
...
     private static QuotaEnforcerBehavior<TestData.TestCommand, TestData.IResponse> CreateQuotaEnforcerBehavior(params MetricStatus[] metricStatuses)
     {
-        var metricStatusesRepository = new MetricStatusesStubRepository(metricStatuses.ToList());
-        var userContextStub = new UserContextStub();
-        return new QuotaEnforcerBehavior<TestData.TestCommand, TestData.IResponse>(metricStatusesRepository, userContextStub);
+        return new QuotaEnforcerBehavior<TestData.TestCommand, TestData.IResponse>(new QuotaCheckerMock(new(metricStatuses)));
+    }
+}
+
+internal class QuotaCheckerMock : IQuotaChecker
```
tnotheis, line 82: Stub

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -8,56 +8,12 @@
 using Enmeshed.UnitTestTools.Extensions;
 using Enmeshed.UnitTestTools.Behaviors;
 using Enmeshed.BuildingBlocks.Application.QuotaCheck;
-using Enmeshed.DevelopmentKit.Identity.ValueObjects;
 
 namespace Enmeshed.BuildingBlocks.Application.Tests.Mediatr;
 public class QuotaEnforcerBehaviorTests
 {
     [Fact]
     public async void Succeeds_when_the_metric_is_not_exhausted()
```
tnotheis, line 16:
```suggestion
    public async void Calls_next_when_no_metric_is_exhausted()
```

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
```diff
@@ -75,55 +31,11 @@ public async void Succeeds_when_none_of_the_metrics_exist_in_the_repository()
...
     [Fact]
     public void Throws_QuotaExhaustedException_when_exactly_one_metric_is_exhausted()
     {
         // Arrange
-        var behavior = CreateQuotaEnforcerBehavior(
-            TestData.MetricStatus.ThatIsExhaustedFor1Day,
-            TestData.MetricStatus.ThatWasExhaustedUntilYesterday
-        );
+        var behavior = CreateQuotaEnforcerBehavior(TestData.MetricStatus.ThatIsExhaustedFor1Day);
         var nextMock = new NextMock<TestData.IResponse>();
```
tnotheis, line 39: You can inline this, because you don't do any assertions on it anyway. This reduces noise.

BuildingBlocks/test/BuildingBlocks.Application.Tests/Mediatr/QuotaEnforcerBehaviorTests.cs
tnotheis, line 1: Why are the assertions for the test with one exhausted metric different from the assertions for the test with two exhausted metrics?

One:
```        
        exceptionExhaustedMetrics.Should().HaveCount(1);
        exceptionExhaustedMetrics.First().MetricKey.Should().Be(TestData.MetricStatus.ThatIsExhaustedFor1Day.MetricKey);
```
(count and metric key)

Two:
```
        exceptionExhaustedMetrics.Should().HaveCount(2);
        exceptionExhaustedMetrics.All(it => it.IsExhaustedUntil > DateTime.Now).Should().BeTrue();
```
(count and IsExhaustedUntil)

As a reader one could think that the application logic for building the exhausted metrics is different in those two cases.

Modules/Quotas/src/Quotas.Application/Metrics/ServiceProviderMetricCalculatorFactory.cs
```diff
@@ -14,7 +13,7 @@ public ServiceProviderMetricCalculatorFactory(IServiceProvider serviceProvider)
 
     public override IMetricCalculator CreateNumberOfSentMessagesMetricCalculator()
     {
-        var messagesRepository = _serviceProvider.GetService<IMessagesRepository>();
-        return new NumberOfSentMessagesMetricCalculator(messagesRepository);
+        var messagesRepository = _serviceProvider.GetRequiredService<NumberOfSentMessagesMetricCalculator>();
```
tnotheis, line 16:
```suggestion
        var calculator= _serviceProvider.GetRequiredService<NumberOfSentMessagesMetricCalculator>();
```

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -43,52 +45,48 @@ public async Task UpdateMetrics(IEnumerable<MetricKey> metrics, MetricCalculator
...
-            _metricStatuses.Add(new(metricKey, maxExhaustionDate));
-        }
-    }
-
     private async Task UpdateMetric(MetricKey metric, IMetricCalculator metricCalculator, CancellationToken cancellationToken)
     {
-        var quotasFromMetric = GetAppliedQuotasForMetric(metric);
-        foreach (var quota in quotasFromMetric)
+        var quotasForMetric = GetAppliedQuotasForMetric(metric);
+        var quotasExhaustion = new Dictionary<Quota, DateTime?>();
```
tnotheis, line 51: Why don't you just remember the latest exhaustion date?

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -43,52 +45,48 @@ public async Task UpdateMetrics(IEnumerable<MetricKey> metrics, MetricCalculator
...
     private IEnumerable<Quota> GetAppliedQuotasForMetric(MetricKey metric)
     {
         var allQuotasOfMetric = AllQuotas.Where(q => q.MetricKey == metric);
-        var highestWeight = -1;
+        if (allQuotasOfMetric.Any())
+        {
+            var highestWeight = allQuotasOfMetric.Max(q => q.Weight);
+            var appliedQuotas = allQuotasOfMetric.Where(q => q.Weight == highestWeight).ToArray();
+            return appliedQuotas;
+        }
```
tnotheis, line 74: You should invert the if to reduce nesting:

```suggestion
        if (!allQuotasOfMetric.Any())
        {
           return Enumerable.Empty<Quota>();
        }
        
        var highestWeight = allQuotasOfMetric.Max(q => q.Weight);
        var appliedQuotas = allQuotasOfMetric.Where(q => q.Weight == highestWeight).ToArray();
        return appliedQuotas;
```

Modules/Quotas/src/Quotas.Infrastructure.Database.SqlServer/Migrations/20230628132645_MetricStatus.cs
```diff
@@ -0,0 +1,73 @@
...
+            migrationBuilder.DropColumn(
+                name: "IsExhaustedUntil",
+                table: "TierQuotas");
+
+            migrationBuilder.AlterColumn<string>(
+                name: "MetricKey",
+                table: "TierQuotaDefinitions",
+                type: "nvarchar(50)",
+                maxLength: 50,
+                nullable: true,
```
tnotheis, line 23: Why is this column nullable? It shouldn't be, right?

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Database/EntityConfigurations/IdentityEntityTypeConfiguration.cs
```diff
@@ -15,5 +15,6 @@ public void Configure(EntityTypeBuilder<Identity> builder)
         builder.HasMany<TierQuota>(x => x.TierQuotas).WithOne().HasForeignKey(x => x.ApplyTo);
         builder.Property(x => x.Address).IsUnicode(false).IsFixedLength().HasMaxLength(IdentityAddress.MAX_LENGTH);
         builder.Property(x => x.TierId).IsUnicode(false).IsFixedLength().HasMaxLength(20);
+        builder.Ignore(x => x.MetricStatuses);
```
tnotheis, line 18: Why do you ignore this?

Modules/Quotas/test/Quotas.Application.Tests/Tests/QuotaCheck/QuotaCheckerImplTests.cs
tnotheis, line 1: Add assertions for the value of `IsSuccess`

tnotheis, line 1: What about a test case where the repository returns 2 metric statuses, but only one of them is exhausted?

tnotheis (changes requested):

Modules/Quotas/test/Quotas.Domain.Tests/QuotaTests.cs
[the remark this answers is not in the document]
Dannyps: Do you mean I should drop the usage of `TierQuotaDefinition`?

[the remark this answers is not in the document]
tnotheis: You should delete the whole tests and instead test it via the Identity class (or rather I will write the tests for it)

Modules/Quotas/src/Quotas.Infrastructure.Database.SqlServer/Migrations/20230628132645_MetricStatus.cs
Dannyps: I don't even understand why this was added. I removed this from the Up and Down, recreated the database and created another migration... which came up clean.

tnotheis: Of course it came up clean, because you probably didn't change the Snapshot.
You should properly delete the migration (via the CLI), then do the fixes necessary for it to not create this code during the next migration generation.

Dannyps:
With the latest fixes to `MetricStatus`, we are back to the Transport tests tripping on that one test that fails most times, i.e., the other tests are now passing.

Dannyps:
@tnotheis I have merged the target branch with main as well.

BuildingBlocks/src/UnitTestTools/Extensions/DateTimeAssertionsExtensions.cs
```diff
@@ -11,20 +12,28 @@ public static AndConstraint<DateTimeAssertions> Be(this DateTimeAssertions it, s
...
     public static AndConstraint<NullableDateTimeAssertions> BeEndOfHour(this NullableDateTimeAssertions it)
     {
-        return it.NotBeNull().
-            And.HaveMinute(59).
-            And.HaveSecond(59);
+        return it.NotBeNull()
+            .And.HaveYear(it.Subject!.Value.Year)
+            .And.HaveMonth(it.Subject!.Value.Year)
+            .And.HaveDay(it.Subject!.Value.Year)
+            .And.HaveHour(it.Subject!.Value.Hour)
```
Dannyps, line 19: Aren't all of these garanteed to be true? Also, the `HaveMonth` and `HaveDay` methods seem to be being called with the wrong parameter.

Modules/Quotas/src/Quotas.Infrastructure.Database.SqlServer/Migrations/QuotasDbContextModelSnapshot.cs
```diff
@@ -43,6 +43,25 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                    b.Property<string>("Owner")
+                        .HasColumnType("char(36)");
+
+                    b.Property<string>("MetricKey")
+                        .HasMaxLength(50)
+                        .IsUnicode(true)
+                        .HasColumnType("nvarchar(50)")
+                        .IsFixedLength(false);
+
+                    b.Property<DateTime?>("IsExhaustedUntil")
```
tnotheis, line 57: You should re-generate the migrations, because `IsExhaustedUntil` is not nullable anymore. This was part of my changes. It is now set to `ExhaustionDate.Unexhausted`, which contains `DateTime.MinValue`

BuildingBlocks/src/UnitTestTools/Extensions/DateTimeAssertionsExtensions.cs
tnotheis: > Aren't all of these garanteed to be true?
You are right, of course. 🤣 I have no idea what I was expecting to achieve.

> Also, the HaveMonth and HaveDay methods seem to be being called with the wrong parameter.
Yup. Copy&paste error...

I will fix it.

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Domain/MetricStatus.cs
```diff
@@ -5,9 +5,11 @@ public class MetricStatus
 {
     public bool IsExhausted => IsExhaustedUntil.HasValue && IsExhaustedUntil > SystemTime.UtcNow;
 
-    public MetricStatus(MetricKey metricKey)
+    public MetricStatus(string metricKey, string owner, DateTime isExhaustedUntil)
```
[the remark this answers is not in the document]
Dannyps: We can use an empty private constructor.

PR #166, opened 2023-06-12
Listing metrics

filipemmagalhaes:
# Readiness checklist

- [x] I labeled the PR.

### DIFF
--- a/Modules/Quotas/src/Quotas.Application/Metrics/Queries/ListMetrics/Handler.cs
+++ b/Modules/Quotas/src/Quotas.Application/Metrics/Queries/ListMetrics/Handler.cs
@@ -1,0 +1,26 @@
+﻿using AutoMapper;
+using Backbone.Modules.Quotas.Application.DTOs;
+using Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
+using MediatR;
+
+namespace Backbone.Modules.Quotas.Application.Metrics.Queries.ListMetrics;
+public class Handler : IRequestHandler<ListMetricsQuery, IEnumerable<MetricDTO>>
+{
+    private readonly IMapper _mapper;
+    private readonly IMetricsRepository _metricsRepository;
+
+    public Handler(IMapper mapper, IMetricsRepository metricsRepository)
+    {
+        _mapper = mapper;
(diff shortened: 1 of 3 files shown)


Modules/Quotas/src/Quotas.Application/Metrics/Queries/ListMetrics/Handler.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public Handler(IMapper mapper, IMetricsRepository metricsRepository)
+    {
+        _mapper = mapper;
+        _metricsRepository = metricsRepository;
+    }
+
+    public async Task<IEnumerable<MetricDTO>> Handle(ListMetricsQuery request, CancellationToken cancellationToken)
+    {
+        var metric = await _metricsRepository.FindAll(cancellationToken);
```
tnotheis, line 20:
```suggestion
        var metrics = await _metricsRepository.FindAll(cancellationToken);
```

Modules/Quotas/src/Quotas.Application/Metrics/Queries/ListMetrics/Handler.cs
```diff
@@ -0,0 +1,26 @@
+﻿using AutoMapper;
+using Backbone.Modules.Quotas.Application.DTOs;
+using Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
+using MediatR;
+
+namespace Backbone.Modules.Quotas.Application.Metrics.Queries.ListMetrics;
+public class Handler : IRequestHandler<ListMetricsQuery, IEnumerable<MetricDTO>>
```
tnotheis, line 7: Create a response class, in order to stay consistent with other handlers like ListRelationships, ListMessages, etc.

Btw, I just noticed that you did the same mistake for the ListClients handlers. Can you fix that as well please?

tnotheis (commented):
The API and UI parts are completely missing. I will close this PR. Open a new one when you're actually done with the task.

tnotheis:
I just realized that it's of course not necessary to write a UI. The metrics are only used in a dropdown list or something like that when creating quotas.
But nonetheless the Admin API needs a route for fetching the metrics.

PR #176, opened 2023-06-21
Admin UI tiers quotas

filipemmagalhaes:
# Readiness checklist

- [x] I labeled the PR.


filipemmagalhaes:
image
image
image
image
image
image

tnotheis:
The task said that you should create a UI for creating Quotas for Tiers. It said nothing about listing created Quotas.

I appreciate your effort, but please remove this for the current PR. Otherwise it will get hard to read the commit history, because we don't know which feature is part of which commit.

tnotheis (changes requested):
Forgot to request changes

JLSRKonk:
tier-quota-empty
tier-quota-create

AdminUI/src/app/app-routing.module.ts
```diff
@@ -13,7 +13,7 @@ const routes: Routes = [
     { path: 'identities', component: IdentityListComponent },
     { path: 'tiers', component: TierListComponent },
     { path: 'tiers/create', component: TierEditComponent },
-    { path: 'tiers/:id', component: TierEditComponent },
+    { path: 'tiers/:id/:name', component: TierEditComponent },
```
tnotheis, line 16: This route looks weird. Why is there id AND name?

JLSRKonk: Yeah I agree it looks weird, but currently, we have no endpoint to get a single tier. By routing like this, we can at least create a tier object with both the Id and the name, otherwise, the edit tier page would have to display an empty name.

tnotheis: Why don't you just implement the getById method of the service so that it fetches all Tiers and then filters them for the one with the id we are looking for?

AdminUI/src/app/services/tier-service/tier.service.ts
```diff
@@ -32,15 +32,32 @@ export class TierService {
...
-        return this.http.get<HttpResponseEnvelope<Tier>>(
-            this.apiUrl,
-            httpOptions
-        );
+        return new Observable<HttpResponseEnvelope<Tier>>((subscriber) => {
+            this.getTiers(0, maxPageSize).subscribe({
+                next: (data: PagedHttpResponseEnvelope<Tier>) => {
+                    const tier = data.result.find((tier) => {
+                        return tier.id == id;
+                    });
```
tnotheis, line 43: Can be simplified:
```suggestion
                    const tier = data.result.find(t => t.id == id);
```
I will approve this PR anyway. You can make this change with your next PR.

AdminUI/src/app/services/tier-service/tier.service.ts
tnotheis, line 1: To be honest, I wouldn't return an HttpResponseEnvelope from the service methods. IMO the places that use the services shouldn't know that the services communicate via Http.

This is just a general note. I don't want you to change this in the current PR. You can resolve this comment as soon as you read it.

tnotheis (approved):

PR #177, opened 2023-06-21
Helm chart

tnotheis:
# Readiness checklist

- [x] I labeled the PR.


.ci/helm/buildChart.ts
```diff
@@ -0,0 +1,32 @@
...
+    .toString("utf-8")
+    .replace(HELM_CHART_VERSION_PLACEHOLDER, version);
+
+  await writeFile(HELM_CHART_FILENAME, chartFile);
+}
+
+async function restoreBackupOfChart() {
+  await rm(HELM_CHART_FILENAME);
+  await rename(`${HELM_CHART_FILENAME}.bak`, HELM_CHART_FILENAME);
+}
```
jkoenig134, line 32:
```suggestion
await $`helm dependency update helm`;
await $`helm package helm --version ${version}`;
```

build-helm.sh
jkoenig134, line 1: this looks like added by mistake?

helm/values.yaml
```diff
@@ -0,0 +1,338 @@
...
+
+  resources:
+    requests:
+      cpu: "200m"
+      memory: "128Mi"
+    limits:
+      cpu: "400m"
+      memory: "1Gi"
+
+  # securityContext:
```
jkoenig134, line 304: you shouldn't set the security context, you can do `securityContext: {}` as you did with `podAnnotations` and `affinity` and leave it to the user to apply a security context.

jkoenig134: this also counts for the other parts where you use securityContext and podSecurityContext.

helm/templates/rabbitmq/service.yaml
jkoenig134, line 1: newline

helm/templates/rabbitmq/deployment.yaml
jkoenig134, line 1: newline

jkoenig134 (changes requested):

PR #184, opened 2023-06-29
Combination of Admin UI and Admin API

tnotheis:
I've removed the service from docker-compose for now because I couldn't get it to work there. We will do this in another PR, because I want to finally deploy the AdminUi in the customer environment. In the meantime, we can launch the AdminUi in a separate process for development purposes. I will explain how to do this via chat.

Another change whose reason isn't so obvious is the one in [BuildingBlocks/src/BuildingBlocks.API/Mvc/ApiControllerBase.cs](https://github.com/nmshd/backbone/compare/combine-admin-ui-and-admin-api?expand=1#diff-6fdb6bb6b460c1674efd5ce830f59f1c85b13498da755fa4d9607f542a0de0e6). I had to remove the `ConsumesAttribute` because otherwise no controllers would be found when I add the call to `MapFallbackToFile` in Program.cs. This is because the attribute says that the controllers are only suitable for requests with a `Content-Type` header of `application/json`. And if there is no content type set (e.g. in a request that has no body), then the controller is considered non-fitting and the fallback is taken.
As this attribute is only relevant for Swagger UI, I have removed it for now.

# Readiness checklist

- [x] I labeled the PR.


Dannyps (approved):
I'm concerned about the way we're going to be deploying the AdminUI for production. As I understand it, we'll be using `ng serve` (with an API proxy). That's highly unrecommended. However, it does make the process simpler for us. Moreover, it is the AdminUI, which should not be exposed to the common end user who might try to exploit the web server serving the application.

Anyway, we will be able to improve on this should we deem it necessary.

tnotheis:
> As I understand it, we'll be using `ng serve`

Only for development. When executing `dotnet publish`, `npm run build` is used to build the angular app. After that, the content of the `dist` folder is moved to the `wwwroot` folder of the built dotnet app. You can find the code for that at the bottom of the `AdminUi.csproj` file.
So this means that for production the ASP.NET server is responsible for delivering the static files of the angular app. No development server of angular involved.

But I really appreciate your comment!

AdminUi/src/AdminUi/appsettings.override.json
```diff
@@ -38,7 +33,7 @@
         "SqlDatabase": {
           "Provider": "SqlServer",
           //"ConnectionString": "User ID=devices;Password=Passw0rd;Server=postgres;Port=5432;Database=enmeshed;" // postgres
```
daniel-almeida-konkconsulting, line 35: Shouldn't the placeholders for the postgres connection strings be changed also?
And on that note, did you test with postgres?

tnotheis: > Shouldn't the placeholders for the postgres connection strings be changed also?
Good idea.

> And on that note, did you test with postgres?
I didn't, but I don't see a reason why it wouldn't work with postgres. I didn't change anything regarding data access. Did I miss something?

Dannyps (approved):

PR #193, opened 2023-07-07
Add support for FirebaseCloudMessaging notifications

Dannyps:
# Readiness checklist

- [x] I labeled the PR.

### DIFF
--- a/Modules/Devices/src/Devices.Infrastructure/PushNotifications/IServiceCollectionExtensions.cs
+++ b/Modules/Devices/src/Devices.Infrastructure/PushNotifications/IServiceCollectionExtensions.cs
@@ -2,6 +2,7 @@
 using Backbone.Modules.Devices.Infrastructure.PushNotifications.AzureNotificationHub;
 using Backbone.Modules.Devices.Infrastructure.PushNotifications.DirectPush;
 using Backbone.Modules.Devices.Infrastructure.PushNotifications.Dummy;
+using Backbone.Modules.Devices.Infrastructure.PushNotifications.FirebaseCouldMessaging;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Backbone.Modules.Devices.Infrastructure.PushNotifications;
@@ -26,6 +27,12 @@ public static class IServiceCollectionExtensions
         }
 
         services.AddTransient<PnsConnectorFactory, PnsConnectorFactoryImpl>();
+
+        services.Configure<FireCouldMessagingConnectorContextOptions>(c =>
+        {
+            c.APIKey = options.FCM_API_Key;
+        });
+        services.AddTransient<IPnsConnector, FireCouldMessagingConnector>();
     }
 }
 
@@ -37,4 +44,6 @@ public class PushNotificationOptions
     public string Provider { get; set; } = IServiceCollectionExtensions.PROVIDER_AZURE_NOTIFICATION_HUB;
 
     public AzureNotificationHub.IServiceCollectionExtensions.AzureNotificationHubPushNotificationsOptions AzureNotificationHub { get; set; }
+
+    public string FCM_API_Key { get; set; }
 }
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -20,6 +20,8 @@ services:
 
       - Modules__Tokens__Infrastructure__BlobStorage__ConnectionInfo=${ENMESHED_BLOB_STORAGE_CONNECTION_STRING} # set this environment variable on your local system to an appropriate value (DefaultEndpointsProtocol=https;AccountName=<account-name>;AccountKey=<account-key>;EndpointSuffix=core.windows.net)
 
+      - Modules__Devices__Infrastructure__PushNotifications__FCM_API_Key=${ENMESHED_FCM_API_KEY} # set this environment variable on your local system to an appropriate value (FCM API KEY)
+
     ports:
       - "127.0.0.10:80:80"
       - "5000:80"
--- a/Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCouldMessaging/FireCouldMessagingConnector.cs
+++ b/Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCouldMessaging/FireCouldMessagingConnector.cs
@@ -1,0 +1,71 @@
+﻿using System.Net.Http.Headers;
+using System.Text;
+using Backbone.Modules.Devices.Infrastructure.PushNotifications.DirectPush;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Backbone.Modules.Devices.Infrastructure.PushNotifications.FirebaseCouldMessaging;
+public class FireCouldMessagingConnector : IPnsConnector
+{
+    private readonly string _apiKey;
+    private readonly HttpClient _client;
+
+    public FireCouldMessagingConnector(IOptions<FireCouldMessagingConnectorContextOptions> options)
+    {
+        _apiKey = options.Value.APIKey;
+        _client = new HttpClient();
+    }
+
+    public async Task Send(IEnumerable<PnsRegistration> registrations, object notification)
+    {
+        var values = new FCMMessage
+        {
+            Data =
+            {
+                AndroidChannelId = "Enmeshed",
+                ContentAvailable = "1",
+                Content = new()
+                {
+                    { "accRef", "id1KJnD8ipfckRQ1ivAhNVLtypmcVM5vPX4j"},
+                    { "eventName", "dynamic"}
+                }
+            },
+            Notification = notification,
+            Recipients = registrations.Select(r=>r.Handle.Value)
+         };
+
+        var httpContent = new StringContent(JsonConvert.SerializeObject(values), Encoding.UTF8, "application/json");
+        _client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue(_apiKey);
+
+        await _client.PostAsync("https://fcm.googleapis.com/fcm/send", httpContent);
+    }
+}
+
+public sealed class FCMMessage
+{
+    [JsonProperty("data")]
+    public FCMData Data { get; set; }
+
+    [JsonProperty("notification")]
+    public object Notification { get; set; }
+
+    [JsonProperty("registration_ids")]
+    public IEnumerable<string> Recipients { get; set; }
+}
+
+public sealed class FCMData
+{
+    [JsonProperty("android_channel_id")]
+    public string AndroidChannelId;
+
+    [JsonProperty("content_available")]
+    public string ContentAvailable;
+
+    [JsonProperty("content")]
+    public Dictionary<string, dynamic> Content;
+}
+
+public sealed class FireCouldMessagingConnectorContextOptions
+{
+    public string APIKey { get; set; }
+}


Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        switch (pushNotification)
+        {
+            case null:
+                return ("", "");
+            case JsonElement jsonElement:
+            {
+                var notification = jsonElement.Deserialize<NotificationTextAttribute>();
+                return notification == null ? ("", "") : (notification.Title, notification.Body);
+            }
```
Dannyps, line 54: I imagine this may be unnecessary and I'm ready to remove it :)

tnotheis: Why didn't you just copy the method from `AzureNotificationHubPushService`?

Modules/Devices/src/Devices.Application/Extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿namespace Backbone.Modules.Devices.Application.Extensions;
+public static class IEnumerableExtensions
+{
+    public static IEnumerable<IEnumerable<T>> Split<T>(this IEnumerable<T> ienum, int batchSize = 100)
```
tnotheis, line 4: I wouldn't assign a default value. This strongly depends on the use case.

tnotheis, line 4:
```suggestion
    public static IEnumerable<IEnumerable<T>> Split<T>(this IEnumerable<T> items, int batchSize = 100)
```

Modules/Devices/src/Devices.Application/Extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            return new List<IEnumerable<T>>();
+        }
+
+        if (list.Count <= batchSize)
+        {
+            return new List<IEnumerable<T>>(new[] { list });
+        }
+
+        var response = new List<IEnumerable<T>>();
```
tnotheis, line 23:
```suggestion
        var batches = new List<IEnumerable<T>>();
```

Modules/Devices/src/Devices.Infrastructure/PushNotifications/DirectPush/PnsConnectorFactory.cs
```diff
@@ -5,6 +5,15 @@ public abstract class PnsConnectorFactory
 {
     public IPnsConnector CreateFor(PushNotificationPlatform platform)
     {
+        switch (platform)
+        {
+            case PushNotificationPlatform.Fcm:
+                return CreateForFirebaseCloudMessaging();
+            case PushNotificationPlatform.Apns:
+                break;
```
tnotheis, line 13: remove this case

Modules/Devices/src/Devices.Application/Extensions/IEnumerableExtensions.cs
tnotheis, line 1: You could write a few tests for this class. This improves stability of the Connector without having to write tests for the connector itself

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
```diff
@@ -0,0 +1,62 @@
...
+    { }
+
+    public async Task Send(IEnumerable<PnsRegistration> registrations, object notification)
+    {
+        var recipients = registrations.Select(r => r.Handle.Value).ToList();
+        var recipientsLists = recipients.Split(500);
+        var data = new Dictionary<string, string>();
+        var (notificationTitle, notificationBody) = GetNotificationText(notification);
+
+        foreach (var list in recipientsLists)
```
tnotheis, line 21:
```suggestion
        foreach (var batch in recipientsLists)
```

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
```diff
@@ -0,0 +1,62 @@
...
+namespace Backbone.Modules.Devices.Infrastructure.PushNotifications.FirebaseCloudMessaging;
+public class FirebaseCloudMessagingConnector : IPnsConnector
+{
+    public FirebaseCloudMessagingConnector()
+    { }
+
+    public async Task Send(IEnumerable<PnsRegistration> registrations, object notification)
+    {
+        var recipients = registrations.Select(r => r.Handle.Value).ToList();
+        var recipientsLists = recipients.Split(500);
```
tnotheis, line 17:
```suggestion
        var recipientsBatches = recipients.Split(500);
```

Modules/Devices/src/Devices.Infrastructure/PushNotifications/IServiceCollectionExtensions.cs
```diff
@@ -1,14 +1,19 @@
...
+using FirebaseAdmin;
+using Google.Apis.Auth.OAuth2;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Backbone.Modules.Devices.Infrastructure.PushNotifications;
 
 public static class IServiceCollectionExtensions
 {
     public const string PROVIDER_AZURE_NOTIFICATION_HUB = "AzureNotificationHub";
+    public const string PROVIDER_PNS_CONNECTOR = "PNSConnector";
```
tnotheis, line 16:
```suggestion
    public const string PROVIDER_DIRECT = "Direct";
```

Modules/Devices/src/Devices.Infrastructure/PushNotifications/IServiceCollectionExtensions.cs
```diff
@@ -21,20 +26,31 @@ public static void AddPushNotifications(this IServiceCollection services, PushNo
...
 {
     [Required]
     [RegularExpression(
-        $"{IServiceCollectionExtensions.PROVIDER_AZURE_NOTIFICATION_HUB}|{IServiceCollectionExtensions.PROVIDER_DUMMY}")]
+        $"{IServiceCollectionExtensions.PROVIDER_AZURE_NOTIFICATION_HUB}|{IServiceCollectionExtensions.PROVIDER_PNS_CONNECTOR}|{IServiceCollectionExtensions.PROVIDER_DUMMY}")]
     public string Provider { get; set; } = IServiceCollectionExtensions.PROVIDER_AZURE_NOTIFICATION_HUB;
 
     public AzureNotificationHub.IServiceCollectionExtensions.AzureNotificationHubPushNotificationsOptions AzureNotificationHub { get; set; }
+
+    public string FCM_Service_Account { get; set; }
```
tnotheis, line 55: This name doesn't comply with any naming convention...

tnotheis, line 55: You should do something similar to the AzureNotificationHub options.

```c#
public DirectPnsCommunicationOptions DirectPnsCommunication { get; set; }
```

```c#
public class DirectPnsCommunicationOptions
{
    public FcmOptions Fcm { get; set; }

    public class FcmOptions
    {
        public string? ServiceAccountJson { get; set; } = string.Empty;
    }
}
```

And when you use it, you should check if ServiceAccountJson is null, and if so, you should use the default credentials:

```c#
FirebaseApp.Create(new AppOptions()
{
    Credential = GoogleCredential.GetApplicationDefault(),
});
```

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
```diff
@@ -0,0 +1,62 @@
...
+        var data = new Dictionary<string, string>();
+        var (notificationTitle, notificationBody) = GetNotificationText(notification);
+
+        foreach (var list in recipientsLists)
+        {
+            var message = CreateMulticastMessage(list, notificationTitle, notificationBody, data);
+            await FirebaseMessaging.DefaultInstance.SendMulticastAsync(message);
```
tnotheis, line 24: I don't like using static classes. You should instead register an instance of FirebaseMessaging in the IServiceCollectionExtensions and inject it in the constructor of the Connector.

tnotheis (changes requested):

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
Dannyps: This implementation supports sending messages with notifications from the test endpoint

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
```diff
@@ -0,0 +1,62 @@
...
+using Backbone.Modules.Devices.Application.Extensions;
+using Backbone.Modules.Devices.Application.Infrastructure.PushNotifications;
+using Backbone.Modules.Devices.Infrastructure.PushNotifications.AzureNotificationHub;
+using Backbone.Modules.Devices.Infrastructure.PushNotifications.DirectPush;
+using FirebaseAdmin.Messaging;
+
+namespace Backbone.Modules.Devices.Infrastructure.PushNotifications.FirebaseCloudMessaging;
+public class FirebaseCloudMessagingConnector : IPnsConnector
+{
+    public FirebaseCloudMessagingConnector()
```
tnotheis, line 11: Remove the empty constructor

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
```diff
@@ -0,0 +1,62 @@
...
+    { }
+
+    public async Task Send(IEnumerable<PnsRegistration> registrations, object notification)
+    {
+        var recipients = registrations.Select(r => r.Handle.Value).ToList();
+        var recipientsLists = recipients.Split(500);
+        var data = new Dictionary<string, string>();
```
tnotheis, line 18: Why do you just send an empty object as `data` for every notification? Am I missing something?

Modules/Devices/src/Devices.Infrastructure/PushNotifications/IServiceCollectionExtensions.cs
```diff
@@ -21,20 +26,34 @@ public static void AddPushNotifications(this IServiceCollection services, PushNo
...
+                    Credential = options.DirectPnsCommunication is null
+                        ? GoogleCredential.GetApplicationDefault()
+                        : GoogleCredential.FromJson(options.DirectPnsCommunication.Fcm.ServiceAccountJson)
+                });
+
+                services.AddTransient<PnsConnectorFactory, PnsConnectorFactoryImpl>();
+                services.AddTransient<FirebaseCloudMessagingConnector>();
+                services.AddTransient<IPushService, DirectPushService>();
+                services.AddSingleton(FirebaseMessaging.DefaultInstance);
+                break;
```
tnotheis, line 41: Extract all of this into a separate extension method, just like with AzureNotificationHub or the Dummy service

Modules/Devices/test/Devices.Application.Tests/Tests/Extensions/IEnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    public void Empty_List_Returns_Empty_List_Of_Lists()
+    {
+        // Arrange
+        var list = new List<int>().AsReadOnly();
+
+        // Act
+        var result = list.Split(10);
+
+        // Assert
+        result.Should().BeEmpty("The argument list was empty.");
```
tnotheis, line 29: It's not necessary to pass a reason. The error messages of FluentAssertions are good enough.

Modules/Devices/test/Devices.Application.Tests/Tests/Extensions/IEnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    [Fact]
+    public void Large_List_Returns_Lists_Of_Lists()
+    {
+        // Arrange
+        var list = new List<int>();
+        for (var i = 0; i < 192; i++)
```
tnotheis, line 37: I always try to make my test data as small as possible. This reduces the complexity of the test.

In this case, this means that I would probably use a total size of 3 and a batch size of 2. Or, if you want to test that multiple lists are returned, 5 and 2.

Modules/Devices/test/Devices.Application.Tests/Tests/Extensions/IEnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            list.Add(i);
+        }
+
+        const int BATCH_SIZE = 10;
+
+        // Act
+        var result = list.Split(BATCH_SIZE);
+
+        // Assert
+        list.Count.Should().BeGreaterThan(BATCH_SIZE);
```
tnotheis, line 48: Why are you asserting on your input data?

Modules/Devices/test/Devices.Application.Tests/Tests/Extensions/IEnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        result.Should().HaveCount((int)Ceiling((double)list.Count / BATCH_SIZE));
```
tnotheis, line 49: Your tests shouldn't have any logic. Every piece of logic has the potential to hide an error. Plus it get's harder to read/understand

Modules/Devices/test/Devices.Application.Tests/Tests/Extensions/IEnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        // Act
+        var result = list.Split(10);
+
+        // Assert
+        result.Should().BeEmpty("The argument list was empty.");
+    }
+
+    [Fact]
+    public void Large_List_Returns_Lists_Of_Lists()
```
tnotheis, line 33: According to the title, if I call `Split` on a list with 1,000,000 items (which is a large list, I suppose?), it always returns multiple lists. But what if I pass 2,000,000 as batch size?

A better name would be "List_bigger_than_batch_size_returns_multiple_lists"

Modules/Devices/test/Devices.Application.Tests/Tests/Extensions/IEnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        var result = list.Split(BATCH_SIZE);
+
+        // Assert
+        list.Count.Should().BeGreaterThan(BATCH_SIZE);
+        result.Should().HaveCount((int)Ceiling((double)list.Count / BATCH_SIZE));
+        result.Aggregate(0, (acc, x) => acc + x.Count()).Should().Be(list.Count);
+    }
+
+    [Fact]
+    public void Small_List_Returns_Single_List_Of_Lists()
```
tnotheis, line 54: Same problem with the test name as above

Modules/Devices/test/Devices.Application.Tests/Tests/Extensions/IEnumerableExtensionsTests.cs
tnotheis, line 1: I'll add some comments to this test. But since you will remove batching of messages, you don't need to implement the changes. Just keep them in mind for the next tests you write.

docker-compose/docker-compose.yml
```diff
@@ -20,6 +20,10 @@ services:
 
       - Modules__Tokens__Infrastructure__BlobStorage__ConnectionInfo=${ENMESHED_BLOB_STORAGE_CONNECTION_STRING} # set this environment variable on your local system to an appropriate value (DefaultEndpointsProtocol=https;AccountName=<account-name>;AccountKey=<account-key>;EndpointSuffix=core.windows.net)
 
+      - Modules__Devices__Infrastructure__PushNotifications__Provider=Direct
+
+      # - 'Modules__Devices__Infrastructure__PushNotifications__DirectPnsCommunication__Fcm__ServiceAccountJson=env_var
+
```
tnotheis, line 26: Don't push that

Modules/Devices/test/Devices.Infrastructure.Tests/Tests/FcmMessageBuilderTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            .SetNotificationText(notificationTitle, notificationText)
+            .AddContent(new NotificationContent(IdentityAddress.Parse("id1KJnD8ipfckRQ1ivAhNVLtypmcVM5vPX4j"), new { SomeProperty = "someValue" }))
+            .Build();
+
+       message.Notification.Title.Should().Be(notificationTitle);
+       message.Notification.Body.Should().Be(notificationText);
+       message.Data.Should().ContainKey("_content");
+       message.Data.Should().ContainKey("tag");
+       message.Data.Should().ContainKey("android_channel_id");
+       message.Data["_content"].Should().BeValidJson();
```
tnotheis, line 33: You should validate the exact JSON here.

Modules/Devices/test/Devices.Infrastructure.Tests/Tests/FcmNotificationBuilderTests.cs
tnotheis, line 1: Don't delete this

tnotheis (changes requested):

Modules/Devices/src/Devices.Infrastructure/PushNotifications/AzureNotificationHub/NotificationBuilder.cs
```diff
@@ -14,7 +14,6 @@ public static NotificationBuilder Create(NotificationPlatform platform)
     {
         NotificationBuilder builder = platform switch
         {
-            NotificationPlatform.Fcm => new FcmNotificationBuilder(),
```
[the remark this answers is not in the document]
Dannyps: I'm going to use this thread to discuss some of the changes commented in this review.
The old `FcmNotificationBuilder` extends the `NotificationBuilder` class. This class is used to build a `Microsoft.Azure.NotificationHubs.Notification`. Hence the deletion: this does not make sense in an FCM context. Furthermore,
 - the `NotificationBuilder` class is in the `AzureNotification` namespace.
 - the equivalent of an AzureNotification Notification is an FCM (Multicast)**Message**, not an FCM **Notification**.

The way I see it, we have a few options here:
1. make the base class Build method return an object or dynamic, so that it can accomodate both these realities.
2. implement another class (be it `FcmMessageBuilder`) which, while sharing some similarities with `ApnsNotificationBuilder` (in the type of methods that are available), will return a totally different type -- what I did.
3. create another class/method which will receive a `Microsoft.Azure.NotificationHubs.Notification` and turn it into an Fcm Message.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
```diff
@@ -1,44 +1,43 @@
...
-    { }
+    private readonly FirebaseMessaging _firebaseMessaging;
 
-    public async Task Send(IEnumerable<PnsRegistration> registrations, object notification)
+    public FirebaseCloudMessagingConnector(FirebaseMessaging firebaseMessaging)
+    {
+        _firebaseMessaging = firebaseMessaging;
+    }
+
+    public async Task Send(IEnumerable<PnsRegistration> registrations, IdentityAddress recipient,
```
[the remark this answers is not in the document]
Dannyps: I had considered this. But then I think that it could be possible that some weird shenanigans might occur, like a bunch of people registering the same identity on a lot of devices. I will remove it nonetheless.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/AzureNotificationHub/NotificationBuilder.cs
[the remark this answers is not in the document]
tnotheis: We can't just remove the support for AzureNotificationHub. 😲 

You should completely ignore the classes in the AzureNotificationHub folder and just create your own classes in your own folder/namespace. So this means: a new base class as well as new subclasses.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/FirebaseCloudMessaging/FirebaseCloudMessagingConnector.cs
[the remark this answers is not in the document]
tnotheis: Yeah, we should rather add some validation for that case, so that you can register more than n devices. But for now I'm okay with it to fail in case someone has that many devices.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/AzureNotificationHub/NotificationBuilder.cs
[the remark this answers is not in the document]
Dannyps: Oh, so AzureNotificationHub is already working with FCM? I hadn't realised that. Very well, I believe I know how to right things now.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/DirectPush/FcmMessageBuilder.cs
```diff
@@ -0,0 +1,72 @@
...
+    private readonly Notification _notification = new();
+
+    public FcmMessageBuilder()
+    {
+        SetAndroidChannelId("ENMESHED");
+    }
+
+    private void SetAndroidChannelId(string channelId)
+    {
+        _data["android_channel_id"] = channelId;
```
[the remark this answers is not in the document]
tnotheis: Read the comment again. It says **additionally**.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/DirectPush/FcmMessageBuilder.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private void SetContentAvailable(bool contentAvailable)
+    {
+        _data["content-available"] = contentAvailable ? "1" : "0";
+    }
+
+    public FcmMessageBuilder SetTag(int notificationId)
+    {
+        _data["tag"] = notificationId.ToString();
```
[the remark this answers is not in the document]
tnotheis: Read the comment again. It says **additionally**.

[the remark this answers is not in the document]
Dannyps: I did add it, while keeping the previous code. Was that not the objective?

Modules/Devices/src/Devices.Infrastructure/PushNotifications/DirectPush/FirebaseCloudMessaging/NotificationContent.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Text.Json;
+using System.Text.Json.Serialization;
+using Enmeshed.DevelopmentKit.Identity.ValueObjects;
+using Enmeshed.Tooling;
+using Newtonsoft.Json;
+
+namespace Backbone.Modules.Devices.Infrastructure.PushNotifications.DirectPush.FirebaseCloudMessaging;
+
+public class NotificationContent
```
tnotheis, line 9: I just thought about it, and we should actually use the same class for AzureNotificationHub and the direct push. Sorry for the inconveniences. So you can move one of the two classes to the push notifications root folder and delete the other one.

tnotheis (changes requested):

Modules/Devices/test/Devices.Infrastructure.Tests/Tests/FcmMessageBuilderTests.cs
tnotheis: I didn't mean that you should serialize the whole message. The comment was only for the `content` property.

Modules/Devices/test/Devices.Infrastructure.Tests/Tests/DirectPush/FcmMessageBuilderTests.cs
```diff
@@ -0,0 +1,85 @@
...
+    public void Message_Without_Content_Matches_Expected_Json()
+    {
+        // Arrange
+        var notificationTitle = "someNotificationTextTitle";
+        var notificationText = "someNotificationTextBody";
+
+        // Act
+        var message = new FcmMessageBuilder()
+            .SetTag(1)
+            .SetNotificationText(notificationTitle, notificationText)
```
tnotheis: You forgot to test the tokens

tnotheis (approved):
I fixed the three comments myself, together with some other stuff.

PR #199, opened 2023-07-11
Send message returns 500

tnotheis:
The problem was that `IN` is not supported by Postgres (see https://stackoverflow.com/a/8388333). 

So in order to fix that, we need two separate DB queries. And since the class would have became to complex, I splitted the repository into one for SqlServer and one for Postgres. I also removed the provider from the options because we don't need it there anymore.

# Readiness checklist

- [x] I labeled the PR.


docker-compose/docker-compose.yml
```diff
@@ -129,8 +129,8 @@ volumes:
   rabbitmq-volume:
   mssql-server-volume:
     external: true
-  #postgres-volume:
-    #external: true
+  postgres-volume:
+    external: true
```
Dannyps, line 133: Perhaps this should stay commented out.

tnotheis: Doesn't harm to have it uncommented. Commenting it out just reduces the effort to change the database provider

Dannyps (approved):

PR #202, opened 2023-07-12
Add pipeline action for integration tests

Dannyps:
# Readiness checklist

- [x] I labeled the PR.


Dannyps:
I realised now someone fixed the AdminUi tests. It was a good exercise, too bad it was needless 😭

.github/workflows/test.yml
```diff
@@ -189,3 +189,23 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install script dependencies
+        run: npm install --prefix ./.ci
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: "7.0.x"
+      - name: Build project
+        run: dotnet restore "Backbone.sln" && dotnet build  /property:WarningLevel=0 --no-restore "Backbone.sln"
```
tnotheis, line 205: Why do you restore and build the solution? This should be done by the Dockerfiles during docker compose up.

Btw, you should probably use Dockerfile instead of Dockerfile.debug in case of the Consumer API

.github/workflows/test.yml
```diff
@@ -189,3 +189,23 @@ jobs:
...
+      - name: Run Consumer API and Admin Ui/API
+        run: docker compose -f ./docker-compose/docker-compose.test.yml up consumer-api admin-ui seed-mssql seed-client -d
+      - name: Run tests
+        run: dotnet test --no-restore --no-build "Backbone.sln" --filter "Category=Integration&TestCategory!~ignore"
+      - name: Stop Consumer API
+        run: docker compose -f ./docker-compose/docker-compose.test.yml down
```
tnotheis, line 211: Extract these commands into a zx script, so that everyone can run the tests manually on their PC.

Same for the Consumer API commands

tnotheis (changes requested):

.github/workflows/test.yml
Dannyps: That's something I hadn't done initially, but it turns out that building the solution inside Docker makes it impossible to run the tests on the host machine later. See https://github.com/nmshd/backbone/actions/runs/5530508228/jobs/10090204253#step:7:12

.github/workflows/test.yml
Dannyps: > Same for the Consumer API commands

Are you referring to the transport tests?

tnotheis: Oops. Yes, of course I do :)

docker-compose/docker-compose.test.yml
```diff
@@ -27,17 +27,24 @@ services:
...
+        condition: service_started
+      rabbitmq:
+        condition: service_started
+      seed-mssql:
+        condition: service_completed_successfully
+      azure-storage-emulator:
+        condition: service_started
+    configs:
+      - source: ConsumerApiConfig
+        target: app/appsettings.json
```
tnotheis, line 40: Is this really necessary? Because this file should actually be copied into the output directory during the build process. If it doesn't there is something wrong

docker-compose/docker-compose.test.yml
```diff
@@ -127,6 +137,10 @@ services:
 configs:
   AdminUiOverrideConfig:
     file: ../AdminUi/src/AdminUi/appsettings.override.json
+  ConsumerApiConfig:
+    file: ../ConsumerApi/appsettings.json
```
tnotheis, line 141: Remember to remove this if it is not necessary (see comment above)

docker-compose/docker-compose.test.yml
```diff
@@ -26,16 +27,47 @@ services:
...
+      azure-storage-emulator:
+        condition: service_started
+    configs:
+      - source: ConsumerApiConfig
+        target: app/appsettings.json
+      - source: ConsumerApiOverrideConfig
+        target: app/appsettings.override.json
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost/health"]
+      test: ["CMD", "wget", "http://localhost/health"]
```
tnotheis, line 44: Why did you remove the `start_period` from this healthcheck?

tnotheis (changes requested):

docker-compose/docker-compose.test.yml
Dannyps: image
It seems the `appsettings.override.json` file is missing. How should I proceed?

tnotheis: My comment was only for appsettings.json.

docker-compose/docker-compose.test.yml
Dannyps: `start_period` is the time Docker will wait before starting the healthchecks.
It came to my attention that in some scenarios (when testing locally and the consumer api is already built) the _booting up_ of the consumer api is rather quick. Moreover, the value we had before was more random than anything.

So, by removing it, we're performing healthchecks from the moment the container starts running. Considering it doesn't harm the loading process, and it's not a heavy operation, I figured it best to leave this way.

tnotheis: I probably would have configured a 5-10s waiting time. But you're right, it probably doesn't harm.

.github/workflows/test.yml
```diff
@@ -23,6 +23,49 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install script dependencies
+        run: npm install --prefix ./.ci
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: "7.0.x"
```
tnotheis, line 37: I think you don't need dotnet, now that you don't run dotnet build/restore anymore

.github/workflows/test.yml
```diff
@@ -23,6 +23,49 @@ jobs:
...
+      - name: Run integration tests
+        run: ./.ci/integration-test.ts
+
+  cns-transport-test:
+    name: Run CNS-Transport Tests
```
tnotheis, line 42:
```suggestion
    name: Run cns-transport Tests
```

tnotheis (changes requested):

tnotheis (approved):

PR #213, opened 2023-07-17
Fix formatting

tnotheis:
This PR is a preparation for the task to add a pipeline that checks the formatting.

There shouldn't be any changes of behavior, so you don't need to review it - famous last words.

# Readiness checklist

- [x] I labeled the PR.


tnotheis:
Okay, seems like the automatic cleanup broke some project references. I will fix this and create a new PR afterwards. Sorry.

PR #221, opened 2023-07-19
Quotas for Identities Application Refactoring

daniel-almeida-konkconsulting:
# Readiness checklist

- [X] I added/updated unit tests.
- [X] I labeled the PR.


Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
```diff
@@ -49,6 +52,37 @@ public async Task Creates_quota_for_identity()
...
+    {
+        // Arrange
+        var command = new CreateQuotaForIdentityCommand(IdentityAddress.Parse("id1KJnD8ipfckRQ1ivAhNVLtypmcVM5vPX4j"), "An-Invalid-Metric-Key", 5, QuotaPeriod.Month);
+        var identitiesRepository = A.Fake<IIdentitiesRepository>();
+        var metricsRepository = new FindMetricsStubRepository(new Metric(MetricKey.NumberOfSentMessages, "Number Of Sent Messages"));
+        var handler = CreateHandler(identitiesRepository, metricsRepository);
+
+        // Act
+        var exception = await Assert.ThrowsAsync<DomainException>(async () => await handler.Handle(command, CancellationToken.None));
+        exception.Message.Should().Contain("The given metric key is not supported.");
```
tnotheis, line 66:
```suggestion
        var exception = await Assert.ThrowsAsync<DomainException>(async () => await handler.Handle(command, CancellationToken.None));
        
        // Assert
        exception.Message.Should().Contain("The given metric key is not supported.");
```

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
```diff
@@ -49,6 +52,37 @@ public async Task Creates_quota_for_identity()
...
+        // Arrange
+        var command = new CreateQuotaForIdentityCommand(IdentityAddress.Parse("id1KJnD8ipfckRQ1ivAhNVLtypmcVM5vPX4j"), "An-Invalid-Metric-Key", 5, QuotaPeriod.Month);
+        var identitiesRepository = A.Fake<IIdentitiesRepository>();
+        A.CallTo(() => identitiesRepository.Find(A<string>._, A<CancellationToken>._, A<bool>._)).Returns((Identity)null);
+        var metricsRepository = new FindMetricsStubRepository(new Metric(MetricKey.NumberOfSentMessages, "Number Of Sent Messages"));
+        var handler = CreateHandler(identitiesRepository, metricsRepository);
+
+        // Act
+        var exception = await Assert.ThrowsAsync<NotFoundException>(async () => await handler.Handle(command, CancellationToken.None));
+        exception.Message.Should().Be("Identity not found. Make sure the ID exists and the record is not expired.");
```
tnotheis, line 82:
```suggestion
        var exception = await Assert.ThrowsAsync<NotFoundException>(async () => await handler.Handle(command, CancellationToken.None));
        
        // Assert
        exception.Message.Should().Be("Identity not found. Make sure the ID exists and the record is not expired.");
```

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
```diff
@@ -49,6 +52,37 @@ public async Task Creates_quota_for_identity()
...
+    public async Task Create_quota_with_invalid_metric_key_throws_domain_exception()
+    {
+        // Arrange
+        var command = new CreateQuotaForIdentityCommand(IdentityAddress.Parse("id1KJnD8ipfckRQ1ivAhNVLtypmcVM5vPX4j"), "An-Invalid-Metric-Key", 5, QuotaPeriod.Month);
+        var identitiesRepository = A.Fake<IIdentitiesRepository>();
+        var metricsRepository = new FindMetricsStubRepository(new Metric(MetricKey.NumberOfSentMessages, "Number Of Sent Messages"));
+        var handler = CreateHandler(identitiesRepository, metricsRepository);
+
+        // Act
+        var exception = await Assert.ThrowsAsync<DomainException>(async () => await handler.Handle(command, CancellationToken.None));
```
tnotheis, line 65: Use fluent assertions

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
```diff
@@ -49,6 +52,37 @@ public async Task Creates_quota_for_identity()
...
+    {
+        // Arrange
+        var command = new CreateQuotaForIdentityCommand(IdentityAddress.Parse("id1KJnD8ipfckRQ1ivAhNVLtypmcVM5vPX4j"), "An-Invalid-Metric-Key", 5, QuotaPeriod.Month);
+        var identitiesRepository = A.Fake<IIdentitiesRepository>();
+        A.CallTo(() => identitiesRepository.Find(A<string>._, A<CancellationToken>._, A<bool>._)).Returns((Identity)null);
+        var metricsRepository = new FindMetricsStubRepository(new Metric(MetricKey.NumberOfSentMessages, "Number Of Sent Messages"));
+        var handler = CreateHandler(identitiesRepository, metricsRepository);
+
+        // Act
+        var exception = await Assert.ThrowsAsync<NotFoundException>(async () => await handler.Handle(command, CancellationToken.None));
```
tnotheis, line 81: Use fluent assertions

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
tnotheis, line 82: Don't check for the exact error message. This will cause lots of false positives once we change the error message even the slightest bit.
In this case, it's sufficient to check if the message starts with "Identity"

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
tnotheis, line 66: Since the error code is expressive enough, there is no need to check for the error message

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
```diff
@@ -62,8 +62,10 @@ public async Task Create_quota_with_invalid_metric_key_throws_domain_exception()
         var handler = CreateHandler(identitiesRepository, metricsRepository);
 
         // Act
-        var exception = await Assert.ThrowsAsync<DomainException>(async () => await handler.Handle(command, CancellationToken.None));
-        exception.Message.Should().Contain("The given metric key is not supported.");
+        var act = async () => await handler.Handle(command, CancellationToken.None);
```
tnotheis, line 65: Don't we call it `acting` in other tests?

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
```diff
@@ -62,8 +62,10 @@ public async Task Create_quota_with_invalid_metric_key_throws_domain_exception()
...
+
+        // Assert
+        var exception = (await act.Should().ThrowAsync<DomainException>()).And;
```
tnotheis, line 68: There is a `AwaitAndThrowAsync` method we implemented which allows you to remove the `await`

tnotheis, line 68: Why do you save the `And` instead of chaining the Assertions like so:

```csharp
acting.Should().ThrowAsync<DomainException>().Which.Code.Should().Be(...)
```

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
```diff
@@ -78,8 +80,11 @@ public async Task Create_quota_for_non_existent_identity_throws_not_found_except
         var handler = CreateHandler(identitiesRepository, metricsRepository);
 
         // Act
-        var exception = await Assert.ThrowsAsync<NotFoundException>(async () => await handler.Handle(command, CancellationToken.None));
-        exception.Message.Should().Be("Identity not found. Make sure the ID exists and the record is not expired.");
+        var act = async () => await handler.Handle(command, CancellationToken.None);
+
+        // Assert
+        var exception = (await act.Should().ThrowAsync<NotFoundException>()).And;
+        exception.Message.Should().StartWith("Identity");
```
tnotheis, line 87: See the three comments above.

tnotheis (changes requested):

tnotheis:
Did you notice that the transport and integration tests are failing? It seems like the requests to Messages and Files are causing a 500 status code, which lead to consequential errors.

tnotheis (changes requested):
Failing tests

tnotheis (approved):

PR #222, opened 2023-07-20
Quota on Number Of Relationships

Dannyps:
# Readiness checklist

- [x] I labeled the PR.

### DIFF
--- a/Modules/Quotas/src/Quotas.Infrastructure/Persistence/Repository/RelationshipsRepository.cs
+++ b/Modules/Quotas/src/Quotas.Infrastructure/Persistence/Repository/RelationshipsRepository.cs
@@ -1,0 +1,21 @@
+﻿using Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
+using Backbone.Modules.Quotas.Domain.Aggregates.Relationships;
+using Backbone.Modules.Quotas.Infrastructure.Persistence.Database;
+using Microsoft.EntityFrameworkCore;
+
+namespace Backbone.Modules.Quotas.Infrastructure.Persistence.Repository;
+public class RelationshipsRepository : IRelationshipsRepository
+{
+    private readonly IQueryable<Relationship> _relationshipsReadonly;
+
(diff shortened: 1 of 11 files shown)


Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/RelationshipChange.cs
tnotheis, line 1: Delete all properties that aren't required for this module. Is this class even used at all? If not, delete it.

Modules/Quotas/src/Quotas.Application/Quotas.Application.csproj
```diff
@@ -8,9 +8,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="12.0.1" />
```
tnotheis, line 11: Why was it necessary to add this?

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Repository/RelationshipsRepository.cs
```diff
@@ -0,0 +1,24 @@
...
+        _relationshipsReadonly = dbContext.Relationships.AsNoTracking();
+    }
+
+    public async Task<uint> Count(string createdBy, DateTime createdAtFrom, DateTime createdAtTo, CancellationToken cancellationToken)
+    {
+        var relationshipsCount = await _relationshipsReadonly
+            .Where(r => r.Status == RelationshipStatus.Pending && r.From == createdBy ||
+                r.Status == RelationshipStatus.Active && (r.From == createdBy || r.To == createdBy))
+            .CountAsync(cancellationToken);
+        return (uint)relationshipsCount;
```
tnotheis, line 22: I think there is too much logic for it to be untested. :(

You can use Enmeshed.UnitTestTools.TestDoubles.Fakes.FakeDbContextFactory to instantiate Sqlite DbContexts which you can use to test

tnotheis (changes requested):

PR #223, opened 2023-07-20
Quotas for Identities Infrastructure

daniel-almeida-konkconsulting:
# Readiness checklist

- [X] I labeled the PR.


Modules/Quotas/src/Quotas.Infrastructure/Persistence/Database/EntityConfigurations/IndividualQuotaEntityTypeConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Backbone.Modules.Quotas.Infrastructure.Persistence.Database.EntityConfigurations;
+
+public class IndividualQuotaEntityTypeConfiguration : IEntityTypeConfiguration<IndividualQuota>
+{
+    public void Configure(EntityTypeBuilder<IndividualQuota> builder)
+    {
+        builder.HasKey(x => x.Id);
+        builder.Property(x => x.Period);
+        builder.Property(x => x.MetricKey);
+        builder.Property(x => x.Max);
```
tnotheis, line 14: Can you ignore the `Weight` property, so that it's not accidentally added some day in the future? Do the same for the TierQuotaEntityTypeConfiguration

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Database/QuotasDbContext.cs
```diff
@@ -20,6 +20,8 @@ public class QuotasDbContext : AbstractDbContextBase
 
     public DbSet<TierQuota> TierQuotas { get; set; }
 
+    public DbSet<IndividualQuota> IndividualQuotas { get; set; }
```
tnotheis, line 23: We shouldn't need this. Also, the `TierQuotas` and `TierQuotasDefinitions` aren't currently used, so they can be removed as well.

To make sure it doesn't affect the migrations, try to create one. If there are changes, tell me. But I don't expect there to be.

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Database/EntityConfigurations/IndividualQuotaEntityTypeConfiguration.cs
tnotheis: You forgot to ignore the Weight of the TierQuota

tnotheis (approved):

PR #229, opened 2023-07-26
Deletion of Tiers

Dannyps:
# Readiness checklist

- [x] I labeled the PR.

## UI Overview

Tiers are listed like before:
image

When editing a Tier, there is a Delete button on the end of the page:
image

Pressing the button will ask for user confirmation before triggering the delete service:
image

En error is shown when trying to delete a non-deletable tier:
image


tnotheis:
Watch out for the PR title ;)

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/DeleteTierCommand.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Backbone.Modules.Devices.Domain.Aggregates.Tier;
+using MediatR;
+
+namespace Backbone.Modules.Devices.Application.Tiers.Commands.DeleteTier;
+
+public class DeleteTierCommand : IRequest<DeleteTierResponse>
```
tnotheis, line 6:
```suggestion
public class DeleteTierCommand : IRequest
```
(don't forget to delete the `DeleteTierResponse`

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/Handler.cs
```diff
@@ -0,0 +1,44 @@
...
+
+public class Handler : IRequestHandler<DeleteTierCommand, DeleteTierResponse>
+{
+    private readonly ITiersRepository _tierRepository;
+    private readonly ILogger<Handler> _logger;
+    private readonly IEventBus _eventBus;
+
+    public Handler(ITiersRepository tierRepository, ILogger<Handler> logger, IEventBus eventBus)
+    {
+        _tierRepository = tierRepository;
```
tnotheis, line 19:
```suggestion
        _tiersRepository = tiersRepository;
```

Modules/Devices/src/Devices.Infrastructure/Persistence/Database/EntityConfigurations/IdentityEntityTypeConfiguration.cs
```diff
@@ -9,7 +10,7 @@ public class IdentityEntityTypeConfiguration : IEntityTypeConfiguration<Identity
     public void Configure(EntityTypeBuilder<Identity> builder)
     {
         builder.HasKey(x => x.Address);
-
         builder.Property(x => x.ClientId).HasMaxLength(200);
+        builder.HasOne<Tier>().WithMany(x => x.Identities).HasForeignKey(x => x.TierId).IsRequired(false);
```
tnotheis, line 14: The reason why it isn't required is that it is null when migrating from an old version, right? If that's the case, could you please add a comment that explains this?

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/Handler.cs
```diff
@@ -0,0 +1,44 @@
...
+        _logger = logger;
+        _eventBus = eventBus;
+    }
+
+    public async Task<DeleteTierResponse> Handle(DeleteTierCommand request, CancellationToken cancellationToken)
+    {
+        var tier = await _tierRepository.FindById(request.TierId, cancellationToken);
+
+        if (tier == await _tierRepository.GetBasicTierAsync(cancellationToken))
```
tnotheis, line 28: This check will never succeed in a real world scenario, because you're checking for reference equality. You should either compare the Ids or actually override the Equals method. I would prefer the latter.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/TiersRepository.cs
```diff
@@ -38,6 +44,11 @@ public async Task<DbPaginationResult<Tier>> FindAll(PaginationFilter paginationF
         return paginationResult;
     }
 
+    public async Task<Tier> FindById(TierId tierId, CancellationToken cancellationToken)
+    {
+        return await _tiersDbSet.Include("Identities").SingleAsync(t => t.Id == tierId, cancellationToken);
```
tnotheis, line 49: `SingleAsync` throws an exception, which leads to an http status code 500... But we want a 404. So return null and throw a NotFoundException in the command handler

tnotheis, line 49: Include all, and not just the Identities. 

Plus: if you ever want to include only a single property, do it with a lambda and not with a string. Or at least use the `nameof` operator to get the property name. Otherwise this fails when renaming the property.....

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+namespace Backbone.Modules.Devices.Application.Tests.Tests.Tiers.Commands.DeleteTier;
+public class HandlerTests
+{
+    private readonly ITiersRepository _tierRepository;
+    private readonly IEventBus _eventBus;
+    private readonly Handler _handler;
+
+    public HandlerTests()
+    {
+        _tierRepository = A.Fake<ITiersRepository>();
```
tnotheis, line 24:
```suggestion
        _tiersRepository = A.Fake<ITiersRepository>();
```

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        _eventBus = A.Fake<IEventBus>();
+        _handler = CreateHandler();
+    }
+
+    private Handler CreateHandler()
+    {
+        var logger = A.Fake<ILogger<Handler>>();
+        return new Handler(_tierRepository, logger, _eventBus);
+    }
```
tnotheis, line 33: Private methods should be below the public methods that use them...

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    [Fact]
+    public async Task Cannot_Delete_Basic_Tier()
+    {
+        // Arrange
+        var basicTierName = TierName.Create("basic-tier").Value;
```
tnotheis, line 39: Please inline all simple variables in all tests of this class that are only used once throughout the tests, like the `basicTierName`

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/Handler.cs
tnotheis, line 28: I would rather suggest a `IsBasicTier` method on the `Tier` class that returns `true` if the Name is `"Basic"`. This way 
1. We save a database call
2. We have the knowledge of what the Basic Tier is in domain layer instead of in the infrastructure layer.

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    public async Task Cannot_Delete_Tier_With_Associated_Identities()
+    {
+        // Arrange
+        var basicTier = new Tier(TierName.Create("basic-tier").Value);
+        var someIdentity = new Identity(
+            TestDataGenerator.CreateRandomDeviceId(),
+            TestDataGenerator.CreateRandomIdentityAddress(),
+            TestDataGenerator.CreateRandomBytes(),
+            TestDataGenerator.CreateRandomTierId(),
+            1);
```
tnotheis, line 61: How about a `TestDataGenerator.CreateIdentity` method?

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        var anotherTier = new Tier(TierName.Create("random-tier").Value) { Identities = new List<Identity>() { someIdentity } };
```
tnotheis, line 62: `anotherTier` sounds as if it would be some irrelevant supporting actor, while in fact it's the Tier that is most important for this test case.

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        // Act
+        var acting = async () => await _handler.Handle(new DeleteTierCommand(anotherTier.Id), CancellationToken.None);
+
+        // Assert
+        await acting.Should().ThrowAsync<ApplicationException>().WithErrorCode("error.platform.validation.device.usedTierCannotBeDeleted");
+    }
+
+    [Fact]
+    public async Task Delete_Non_Basic_Tier_Without_Identities_Publishes_IntegrationEvent()
```
tnotheis, line 75: What if there were 10 more preconditions for a successful deletion? Would you add all of them to this test name?

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        // Arrange
+        var basicTier = new Tier(TierName.Create("basic-tier").Value);
+        var anotherTier = new Tier(TierName.Create("random-tier").Value) { Identities = new List<Identity>() };
```
tnotheis, line 79: See comment in the previous test (=> bad naming)

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierDeleted/TierDeletedIntegrationEventHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Backbone.Modules.Quotas.Application.IntegrationEvents.Incoming.TierDeleted;
+public class TierDeletedIntegrationEventHandler : IIntegrationEventHandler<TierDeletedIntegrationEvent>
+{
+    private readonly ITiersRepository _tierRepository;
+    private readonly ILogger<TierDeletedIntegrationEventHandler> _logger;
+    private readonly IMediator _mediator;
+
+    public TierDeletedIntegrationEventHandler(ILogger<TierDeletedIntegrationEventHandler> logger, ITiersRepository tierRepository, IMediator mediator)
+    {
+        _tierRepository = tierRepository;
```
tnotheis, line 16:
```suggestion
        _tiersRepository = tiersRepository;
```

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierDeleted/TierDeletedIntegrationEventHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    public async Task Handle(TierDeletedIntegrationEvent integrationEvent)
+    {
+        var tier = await _tierRepository.Find(integrationEvent.Id, CancellationToken.None);
+
+        var tasks = new List<Task>();
+        foreach (var tierQuotaDefinition in tier.Quotas)
+        {
+            tasks.Add(_mediator.Send(new DeleteTierQuotaDefinitionCommand(tier.Id, tierQuotaDefinition.Id)));
+        }
```
tnotheis, line 29: I messed it up in the task description. You don't need to delete the TierQuotaDefinitions manually, of course. They should be deleted by the database automatically when you delete the Tier.

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierDeleted/TierDeletedIntegrationEventHandler.cs
tnotheis, line 1: There are currently no tests for the IntegrationEventHandler

tnotheis (changes requested):
Add screenshots to the PR description, please.

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/Handler.cs
Dannyps: We're letting go of this, but it should be noted that it does work.

AdminUi/test/AdminUi.Tests.Integration/Features/Tiers/DELETE.feature
```diff
@@ -0,0 +1,9 @@
+@Integration
+Feature: DELETE Tier
+
+User deletes a Tier
+
```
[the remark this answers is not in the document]
Dannyps: > * The user tries to delete a Tier that is in use

I'm unsure we can test this right now. The Admin API does not support setting a Tier for an Identity, as far as I can see.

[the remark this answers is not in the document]
tnotheis: It does. There is a `POST /Identities/{address}/Quotas` route.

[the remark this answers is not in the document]
tnotheis: Aaaaand I'm should read your comment before I answer it. ;-) You're right, it isn't possible to set a Tier. :-(

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -5,41 +5,31 @@
...
-
     [Fact]
     public async Task Cannot_Delete_Basic_Tier()
     {
         // Arrange
-        var basicTierName = TierName.Create("basic-tier").Value;
-        var basicTier = new Tier(basicTierName);
-        A.CallTo(() => _tierRepository.GetBasicTierAsync(A<CancellationToken>._)).Returns(Task.FromResult(basicTier));
-        A.CallTo(() => _tierRepository.FindById(basicTier.Id, A<CancellationToken>._)).Returns(Task.FromResult(basicTier));
+        var basicTier = new Tier(TierName.Create(TierName.BASIC_DEFAULT_NAME).Value);
```
tnotheis, line 31: Remember to update this after you changed the type of the BASIC_DEFAULT_NAME. The compiler won't complain because it's still valid code due to the implemented implicit conversion from TierName to string.

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -52,39 +42,35 @@ public async Task Cannot_Delete_Basic_Tier()
...
-        var acting = async () => await _handler.Handle(new DeleteTierCommand(anotherTier.Id), CancellationToken.None);
+        var acting = async () => await _handler.Handle(new DeleteTierCommand(tier.Id), CancellationToken.None);
 
         // Assert
         await acting.Should().ThrowAsync<ApplicationException>().WithErrorCode("error.platform.validation.device.usedTierCannotBeDeleted");
     }
 
     [Fact]
-    public async Task Delete_Non_Basic_Tier_Without_Identities_Publishes_IntegrationEvent()
+    public async Task Delete_Valid_Tier_Publishes_IntegrationEvent()
```
tnotheis, line 58:
```suggestion
    public async Task Publishes_TierDeletedIntegrationEvent_after_successful_deletion()
```

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierDeleted/TierDeletedIntegrationEventHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        _tiersRepository = tiersRepository;
+        _logger = logger;
+    }
+
+    public async Task Handle(TierDeletedIntegrationEvent integrationEvent)
+    {
+        var tier = await _tiersRepository.Find(integrationEvent.Id, CancellationToken.None);
+
+        await _tiersRepository.Remove(tier);
```
tnotheis, line 21: Now that I see this, I think that since we don't perform any checks on the retrieved Tier, I think we can as well just have a `RemoveById` method on the repository. This would save us one database roundtrip

tnotheis (changes requested):

Modules/Devices/src/Devices.Domain/Aggregates/Tier/Tier.cs
```diff
@@ -10,4 +12,10 @@ public Tier(TierName name)
 
     public TierId Id { get; }
     public TierName Name { get; }
+    public ICollection<Identity> Identities { get; internal set; }
+
+    public bool IsBasicTier()
+    {
+        return Name == TierName.BASIC_DEFAULT_NAME;
```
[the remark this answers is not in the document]
Dannyps: As far as I can understand, there's no way to keep it a constant if we need a constructor to instantiate it.

Does this look acceptable?
`public static TierName BASIC_DEFAULT_NAME = new("Basic");`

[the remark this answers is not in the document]
tnotheis: > Does this look acceptable?
`public static TierName BASIC_DEFAULT_NAME = new("Basic");`

Yes, sure. It's the same as for the MetricKeys.

> Becasue TierName is a record, we cannot override the == operator. That would help simplify the rest of the code.

What would you override the operator with? The == operator of a record checks all properties of the two compared instances for equality, which should be exactly what we need here. Am I missing something?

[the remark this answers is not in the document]
Dannyps: I was under the impression it checked for object reference. Great then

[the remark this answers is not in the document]
tnotheis: Nope, that's one of the cool features of records, and the main reason why you should choose records over classes for value objects.

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierDeleted/TierDeletedIntegrationEventHandler.cs
Dannyps: What EF method should I use in this scenario? I can only find `Remove` and `RemoveRange`.

tnotheis: How about https://stackoverflow.com/a/75185453/4046585

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/Handler.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            throw new ApplicationException(ApplicationErrors.Devices.BasicTierCannotBeDeleted());
+        }
+
+        var identitiesCount = await _tiersRepository.GetIdentitiesCount(tier, cancellationToken);
+
+        if (identitiesCount > 0)
+        {
+            throw new ApplicationException(ApplicationErrors.Devices.UsedTierCannotBeDeleted(identitiesCount));
+        }
```
tnotheis, line 46: You should do this in a more domain centric way:
```suggestion
        var identitiesCount = await _tiersRepository.GetIdentitiesCount(tier, cancellationToken);

        var impediment = tier.CanBeDeleted(identitiesCount)
        if(impediment != null)
        {
            throw new DomainException(impediment);
        }
```

The `CanBeDeleted` method then checks for the Name and the number of identities and returns a DomainError if there is something wrong. If everything is fine, it returns null.

Regarding the tests:
* you should write tests for the `CanBeDeleted` method. 
* you can simplify the HandlerTests by boiling them down to only a single test that checks for an exception. You can simulate the exception by returning an invalid Tier from the fake repository.

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,79 @@
...
+    private readonly IEventBus _eventBus;
+    private readonly Handler _handler;
+
+    public HandlerTests()
+    {
+        _tiersRepository = A.Fake<ITiersRepository>();
+        _eventBus = A.Fake<IEventBus>();
+        _handler = CreateHandler();
+    }
+
```
tnotheis, line 27: A test for the success case is missing

Dannyps: Isn't `Publishes_TierDeletedIntegrationEvent_after_successful_deletion` just that?

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/Handler.cs
Dannyps: Regarding the tests, do you mean we can move the exception-specific logic from the handler tests to the `CanBeDeleted` tests?

AdminUi/src/AdminUi/ClientApp/src/app/services/tier-service/tier.service.ts
```diff
@@ -36,8 +36,8 @@ export class TierService {
         return this.http.put<HttpResponseEnvelope<Tier>>(this.apiUrl, tier);
     }
 
-    deleteTierById(id: string): Observable<HttpResponseEnvelope<boolean>> {
-        return this.http.delete<HttpResponseEnvelope<boolean>>(this.apiUrl + `/${id}`);
+    deleteTierById(id: string): Observable<HttpResponseEnvelope<any>> {
+        return this.http.delete<HttpResponseEnvelope<any>>(this.apiUrl + `/${id}`);
```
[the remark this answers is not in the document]
Dannyps: image
It's what we've been doing with `.delete` calls. I can look into the best practices in this scenario if that's preferred.

AdminUi/src/AdminUi/ClientApp/src/app/utils/http-error-response-wrapper.ts
```diff
@@ -0,0 +1,15 @@
+export interface HttpErrorResponseWrapper {
```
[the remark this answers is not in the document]
Dannyps: There was indeed space for improvement here. I've updated the interface.
Further context can be found on https://angular.io/api/common/http/HttpErrorResponse

Modules/Devices/src/Devices.Application/Tiers/Commands/DeleteTier/Handler.cs
tnotheis: Yep. Except that the CanBeDeleted method won't throw exceptions but instead return a DomainError if there is something wrong

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
tnotheis: What if the method simply triggers the event but doesn't delete the Tier from the database?

Btw, in your negative test case, you should also check that the Tier was NOT deleted. Because it may be possible that the Delete method is called before the exception is thrown

Dannyps: Makes sense, ty

Modules/Devices/test/Devices.Application.Tests/Tests/Tiers/Commands/DeleteTier/HandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        // Act
+        var acting = async () => await _handler.Handle(new DeleteTierCommand(tier.Id), CancellationToken.None);
+
+        // Assert
+        await acting.Should().ThrowAsync<DomainException>();
+        A.CallTo(() => _tiersRepository.Remove(tier)).MustNotHaveHappened();
+    }
+
+    [Fact]
+    public async Task Publishes_TierDeletedIntegrationEvent_and_calls_Remove_after_successful_deletion()
```
tnotheis, line 46: I always put the happy path to the top

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierCanBeDeleted.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Backbone.Modules.Devices.Domain.Aggregates.Tier;
+using FluentAssertions;
+using Xunit;
+
+namespace Backbone.Modules.Devices.Domain.Tests.Tiers;
+public class CanBeDeletedTests
+{
+    [Fact]
+    public void Tier_with_name_Basic_Cannot_be_deleted()
```
tnotheis, line 9:
```suggestion
    public void Basic_Tier_cannot_be_deleted()
```

tnotheis (changes requested):

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierIdTests.cs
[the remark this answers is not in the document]
Dannyps: I figured it would be best to isolate the tests about Tiers.

AdminUi/src/AdminUi/ClientApp/src/app/services/tier-service/tier.service.ts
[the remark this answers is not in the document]
tnotheis: I missed that comment. 

You should probably do it just like it's done in the integration tests. -> create a separate HttpResponseEnvelope with only the properties that make sense in a 204 case.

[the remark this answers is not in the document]
Dannyps: So, when receiving a 204 response, Angular's HttpClient will return an Observable emitting a `null` value. Error responses will make the observable emit an `error`. I suggest we do it like this:
```ts
deleteTierById(id: string): Observable<null> {
        return this.http.delete<null>(this.apiUrl + `/${id}`);
    }
```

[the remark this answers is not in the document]
tnotheis: I'm less bothered by the call of `http.delete` than I am by the return type definition of the `deleteTierById` method.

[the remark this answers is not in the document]
Dannyps: They are tightly connected, however. The way we call the `http.delete` function directly affects the return type definition of the `deleteTierById` function.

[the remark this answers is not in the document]
tnotheis: Nobody said that you have to return exactly what's returned by the HttpClient. 🤔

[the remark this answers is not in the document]
Dannyps: I've had quite sufferable debug sessions due to bad return types on this kind of calls 😅
In this case, unless we interfere with the observable, I don't see how we can return something else.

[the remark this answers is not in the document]
tnotheis: I digged into it a little bit, and it looks as if it was more complicated than I expected. These angular types infected everything. :D 

The following is the best I could come up with:

```ts
    deleteClient(clientId: string): Observable<never> {
        return this.http.delete<never>(`${this.apiUrl}/${clientId}`);
    }
```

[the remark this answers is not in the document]
Dannyps: I had seen the `never` with Daniel Almeida. I'm unsure if we should use it, it's a return type for [functions that never _return_](https://www.typescriptlang.org/docs/handbook/basic-types.html#never), It's not the case here, we _can_ get the return. It's just null.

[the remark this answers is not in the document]
tnotheis: Hmm, okay. You're probably right. And what about `Observable<void>`?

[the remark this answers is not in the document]
Dannyps:
```ts
deleteTierById(id: string): Observable<void> {
        return this.http.delete<void>(this.apiUrl + `/${id}`);
    }
```

this works nicely, yes.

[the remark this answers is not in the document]
tnotheis: Any reason why you return `null` instead of `void` now?

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierIdTests.cs
```diff
@@ -61,7 +63,7 @@ public void Cannot_create_tier_id_with_invalid_id_characters()
     {
         char[] invalidCharacters = { '|', '-', '!' };
         var validTierIdPrefix = "TIR";
-        var tierIdValue = TestDataGenerator.GenerateString(TierId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX, invalidCharacters);
+        var tierIdValue = TestDataGenerator.GenerateString(StronglyTypedId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX, invalidCharacters);
```
tnotheis, line 66:
```suggestion
        var TIER_ID_WITH_INVALID_CHARACTER = "TIRxxxxxxxxxxxxxxxx!";
```

You can also remove the `invalidCharacters` variable. It doesn't make sense to test more than one. As long as you don't test every single invalid character, testing more than one doesn't improve the chance that there isn't one invalid character that ISN'T recognized.

AdminUi/src/AdminUi/ClientApp/src/app/services/tier-service/tier.service.ts
```diff
@@ -35,6 +35,10 @@ export class TierService {
     updateTier(tier: Tier): Observable<HttpResponseEnvelope<Tier>> {
         return this.http.put<HttpResponseEnvelope<Tier>>(this.apiUrl, tier);
     }
+
+    deleteTierById(id: string): Observable<null> {
+        return this.http.delete<null>(this.apiUrl + `/${id}`);
```
tnotheis, line 40: `void` instead of `null`?

tnotheis (changes requested):

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierIdTests.cs
```diff
@@ -61,10 +60,8 @@ public void Cannot_create_tier_id_with_invalid_id_length()
     [Fact]
     public void Cannot_create_tier_id_with_invalid_id_characters()
     {
-        char[] invalidCharacters = { '|', '-', '!' };
-        var validTierIdPrefix = "TIR";
-        var tierIdValue = TestDataGenerator.GenerateString(StronglyTypedId.DEFAULT_MAX_LENGTH_WITHOUT_PREFIX, invalidCharacters);
-        var tierId = TierId.Create(validTierIdPrefix + tierIdValue);
+        const string TIER_ID_WITH_21_CHARACTERS = "TIRxxxxxxxxxxxxxxxxxx";
```
tnotheis, line 63: Copy paste error

tnotheis (changes requested):

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierIdTests.cs
Dannyps: oops

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierIdTests.cs
```diff
@@ -60,7 +60,7 @@ public void Cannot_create_tier_id_with_invalid_id_length()
     [Fact]
     public void Cannot_create_tier_id_with_invalid_id_characters()
     {
-        const string TIER_ID_WITH_21_CHARACTERS = "TIRxxxxxxxxxxxxxxxxxx";
+        const string TIER_ID_WITH_21_CHARACTERS = "TIRxxxxxxxxxxxxxxxxx!";
```
tnotheis, line 63:
```suggestion
        const string TIER_ID_WITH_INVALID_CHARACTER = "TIRxxxxxxxxxxxxxxxxx!";
```

tnotheis (changes requested):

tnotheis (approved):

PR #230, opened 2023-07-26
List identity quotas

JLSRKonk:
# Readiness checklist

- [x] I labeled the PR.


JLSRKonk:
**Screenshots**

Screenshot 2023-07-26 173830

tnotheis:
Regarding the screenshots: you forgot to "disable" all Tier Quotas (e.g. make them grey) on a Metric if there are any Individual Quotas. And ideally there is a tooltip on the disabled Quotas that explains why they are disabled.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
```diff
@@ -49,18 +53,50 @@ export class IdentityEditComponent {
...
+
+        if (quotas[0].metric.key == metricKey) {
+            this.quotasTableData.push(quotas[0]);
+            return this.iterateQuotasByMetric(quotas.slice(1), metricKey);
+        }
+
+        return quotas;
+    }
+
+    isGroup(index: any, item: any): boolean {
```
tnotheis, line 96: This method isn't used anywhere. Is it used by angular?

AdminUi/src/AdminUi/Controllers/IdentitiesController.cs
```diff
@@ -50,6 +54,41 @@ public async Task<CreatedResult> CreateIndividualQuota([FromRoute] string identi
         var createdIndividualQuotaDTO = await _mediator.Send(new CreateQuotaForIdentityCommand(identityAddress, request.MetricKey, request.Max, request.Period), cancellationToken);
         return Created(createdIndividualQuotaDTO);
     }
+
+    [HttpGet("{address}")]
+    [ProducesResponseType(typeof(GetIdentityByAddressResponse), StatusCodes.Status200OK)]
+    [ProducesError(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> GetIdentityByAddressAsync([FromRoute] string address, CancellationToken cancellationToken)
```
tnotheis, line 61: Remove the "Async" suffix. From the other methods in this class as well please

AdminUi/test/AdminUi.Tests.Integration/Features/Identities/GET.feature
```diff
@@ -7,3 +7,9 @@ Scenario: Requesting an Identity List
 	When a GET request is sent to the /Identities endpoint
 	Then the response status code is 200 (OK)
 	And the response contains a paginated list of Identities
+
```
tnotheis, line 10: Where is the testcase for 404?

Modules/Devices/src/Devices.Application/DTOs/IdentitySummaryDTO.cs
```diff
@@ -20,6 +20,8 @@ public class IdentitySummaryDTO
 
     public IEnumerable<DeviceDTO> Devices { get; set; }
 
+    public IEnumerable<QuotaDTO> Quotas { get; set; }
```
tnotheis, line 23: Wtf. Why does the Devices module now know what Quotas are? 😮 That's definitely not what we want! The modules have to stay strongly separated.

And even if it was correct to have it here: why do you add the Quotas to the SummaryDTO?!

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/GetIdentityByAddress/HandlerTests.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Backbone.Modules.Devices.Application.Tests.Tests.Identities.Queries.GetIdentityByAddress;
+public class HandlerTests
+{
+    [Fact]
+    public async void Gets_identity_by_address()
+    {
+        // Arrange
+        var expectedClientId = TestDataGenerator.CreateRandomDeviceId();
+        var expectedAddress = TestDataGenerator.CreateRandomIdentityAddress();
+        var expectedTierId = TestDataGenerator.CreateRandomTierId();
```
tnotheis, line 20: Remove these variables. You can use `identity.Address` for the assertions (and similarly for the other properties)

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Queries/GetIdentityByAddress/HandlerTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        var handler = CreateHandler(new FindByAddressStubRepository(identity));
+
+        // Act
+        var result = await handler.Handle(new GetIdentityByAddressQuery(identity.Address), CancellationToken.None);
+
+        // Assert
+        result.ClientId.Should().Be(expectedClientId);
+        result.Address.Should().Be(expectedAddress);
+        result.PublicKey.Should().BeEquivalentTo(Array.Empty<byte>());
```
tnotheis, line 31: Using an empty array is not ideal, as it might as well be empty because it is not mapped correctly.

AdminUi/src/AdminUi/Controllers/IdentitiesController.cs
```diff
@@ -50,6 +54,41 @@ public async Task<CreatedResult> CreateIndividualQuota([FromRoute] string identi
...
+    {
+        var identity = await _mediator.Send(new GetIdentityByAddressQuery(address), cancellationToken);
```
tnotheis, line 63:
```suggestion
        var identity = await _mediator.Send(new GetIdentityQuery(address), cancellationToken);
```

AdminUi/src/AdminUi/Controllers/IdentitiesController.cs
```diff
@@ -50,6 +54,41 @@ public async Task<CreatedResult> CreateIndividualQuota([FromRoute] string identi
...
+        var identityWithQuotas = await _mediator.Send(new GetIdentityQuotasByAddressQuery(address), cancellationToken);
```
tnotheis, line 64: Instead of the `GetIdentityQuotasByAddressQuery`, there should be a `GetIdentityQuery` in the Quotas module as well, which for now has two properties: `Address` and `Quotas`.

Modules/Devices/src/Devices.Application/DTOs/QuotaDTO.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        Id = id;
+        Source = source;
+        Metric = metric;
+        Max = max;
+        Period = period;
+    }
+
+    public string Id { get; set; }
+    public string Source { get; set; }
```
tnotheis, line 14: Should be an enum

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
JLSRKonk: It's used for determining whether the table row is a quota or a metric group:
`<tr mat-row *matRowDef="let row; columns: ['metricGroup']; when: isGroup"></tr>`

Modules/Devices/src/Devices.Application/DTOs/IdentitySummaryDTO.cs
tnotheis: You last suggestion is close to what we need.

But instead of having only two properties, merge the properties of the two IdentityDTOs you receive from the two queries, so that it results in the following JSON response:

```json
{
  "address": "id1...",
  "createdAt": "2023-...."
  ...
  "quotas": [
    {
      "id": "...",
      "max": 5,
      ...
    }
  ]
}
```

JLSRKonk:
Untitled

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.html
```diff
@@ -47,6 +47,51 @@ <h2 class="header-title">{{ header }}</h2>
...
+                        {{ Quota.period }}
+                    </td>
+                </ng-container>
+                <tr mat-header-row *matHeaderRowDef="quotasTableDisplayedColumns"></tr>
+                <tr
+                    mat-row
+                    *matRowDef="let row; columns: quotasTableDisplayedColumns"
+                    [ngClass]="{ 'disabled-tier': row.tierDisabled }"
+                    [matTooltipDisabled]="!row.tierDisabled"
+                    matTooltip="Tier quotas are disabled if an individual quota is present"
```
tnotheis, line 79:
```suggestion
                    matTooltip="Tier quotas do not take effect if there is an individual quota."
```

AdminUi/src/AdminUi/ClientApp/src/app/services/quotas-service/quotas.service.ts
```diff
@@ -60,6 +60,7 @@ export interface Quota {
     metric: Metric;
     max: number;
     period: string;
+    tierDisabled: boolean;
```
tnotheis, line 63: Why is this property called `tierDisabled`? Shouldn't it just be called `disabled`? The UI doesn't care if it is a tier quota or an individual quota

AdminUi/test/AdminUi.Tests.Integration/Features/Identities/GET.feature
```diff
@@ -13,3 +13,8 @@ Scenario: Requesting an Identity
 	When a GET request is sent to the /Identities/{i.address} endpoint
 	Then the response status code is 200 (OK)
 	And the response contains Identity i
+
+Scenario: Requesting an inexistent Identity
+	When a GET request is sent to the /Identities/inexistentIdentityAddress endpoint
```
tnotheis, line 18: Just had a better idea for endpoints like this:
```suggestion
	When a GET request is sent to the /Identities/{address} endpoint with an inexistent address
```

Modules/Quotas/src/Quotas.Application/DTOs/TierQuotaDTO.cs
tnotheis, line 1: Can you delete this class please? I don't know if we will ever need it. And if we do, it's not hard to recreate it.

tnotheis (changes requested):

tnotheis (approved):

PR #243, opened 2023-08-09
Only allow 1 Quota per Type per Period per Metric

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.


Modules/Quotas/src/Quotas.Application/Tiers/Commands/CreateQuotaForIdentity/Handler.cs
```diff
@@ -32,6 +33,11 @@ public async Task<IndividualQuotaDTO> Handle(CreateQuotaForIdentityCommand reque
 
         var metric = await _metricsRepository.Find(parseMetricKeyResult.Value, cancellationToken);
 
+        if (identity.IndividualQuotas.Any(q => q.MetricKey == metric.Key && q.Period == request.Period))
+        {
+            throw new DomainException(DomainErrors.DuplicateQuota());
+        }
```
tnotheis, line 39: Once again: please put as much logic as possible into the Domain classes.

Modules/Quotas/src/Quotas.Application/Tiers/Commands/CreateQuotaForTier/Handler.cs
```diff
@@ -37,6 +38,11 @@ public async Task<TierQuotaDefinitionDTO> Handle(CreateQuotaForTierCommand reque
 
         var metric = await _metricsRepository.Find(parseMetricKeyResult.Value, cancellationToken);
 
+        if (tier.Quotas.Any(q => q.MetricKey == metric.Key && q.Period == request.Period))
+        {
+            throw new DomainException(DomainErrors.DuplicateQuota());
+        }
```
tnotheis, line 44: See above.

Modules/Quotas/src/Quotas.Domain/DomainErrors.cs
```diff
@@ -13,4 +13,9 @@ public static DomainError MaxValueCannotBeLowerOrEqualToZero()
     {
         return new DomainError("error.platform.quotas.invalidValueForMaxLimitInQuota", "A quota max value cannot be lower of equal to zero.");
     }
+
+    public static DomainError DuplicateQuota()
+    {
+        return new DomainError("error.platform.quotas.duplicateQuota", "There is already a quota for this Tier/Identity targeting the same Metric and the same Period.");
```
tnotheis, line 19: I would remove the "Tier/Identity" part, because it may be confusing.
```suggestion
        return new DomainError("error.platform.quotas.duplicateQuota", "A quota targeting the same metric and period already exists.");
```

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Application/Tiers/Commands/CreateQuotaForIdentity/Handler.cs
tnotheis: Still not what I expected. 
The goal is that it is impossible for the user of the Identity class to do something that leads to an invalid Identity (may it be on purpose or not). Currently you could remove the call of the `IndividualQuotaAlreadyExists` method.

So to solve this, simply move the check into the `CreateIndividualQuota` method.

Modules/Quotas/src/Quotas.Application/Tiers/Commands/CreateQuotaForTier/Handler.cs
tnotheis: See above.

Modules/Quotas/src/Quotas.Application/Tiers/Commands/CreateQuotaForIdentity/Handler.cs
Dannyps: That makes sense. I came across a weird test however, which I had to change. Please ensure that change makes sense.

https://github.com/nmshd/backbone/blob/d7d2ad833f06a1669bb4c37d288bd9ed1b452296/Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/DeleteTierQuotaDefinition/HandlerTests.cs#L57

tnotheis: That test needs to be updated, of course. :)

Dannyps:
@tnotheis I'm unable to re-request review for some reason, but it should be good to go now.

tnotheis:
> @tnotheis I'm unable to re-request review for some reason, but it should be good to go now.

The reason is that I didn't request changes. :)

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -78,6 +81,11 @@ public void DeleteTierQuotaFromDefinitionId(TierQuotaDefinitionId tierQuotaDefin
         }
     }
 
+    public bool IndividualQuotaAlreadyExists(MetricKey metricKey, QuotaPeriod period)
```
tnotheis, line 84: Private

Modules/Quotas/src/Quotas.Domain/Aggregates/Tiers/Tier.cs
```diff
@@ -40,4 +43,9 @@ public Tier(TierId id, string name)
 
         return Result.Success<TierQuotaDefinitionId, DomainError>(quotaDefinition.Id);
     }
+
+    public bool TierQuotaAlreadyExists(MetricKey metricKey, QuotaPeriod period)
```
tnotheis, line 47: Private

tnotheis (changes requested):
Whoops. Wanted to request changes :)

tnotheis (changes requested):
Just noticed that you still test the Handlers, even though this logic is part of the Identity/Tier.

Dannyps:
Indeed, it did not make a lot of sense. I'm reworked the tests.

tnotheis (changes requested):
Damn... Another thing just came in my mind. Sorry...

Can you "catch" this error in the UI and show a proper error message please? Or do we already show the error message that is returned by the Backend?

Dannyps:
This was already done, yes:

image

It is not working anymore, however (blame those merges from main 😅). I'm working on this with Mr. Almeida.

Dannyps:
That last commit fixes it.

tnotheis (approved):

PR #244, opened 2023-08-10
Admin UI security

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.


tnotheis (commented):
I only skimmed the changes, but as far as I can see you only implemented changes regarding a XSRF. What about other attacks?

AdminUi/src/AdminUi/Program.cs
```diff
@@ -111,6 +111,7 @@ static void Configure(WebApplication app)
             .AddDefaultSecurityHeaders()
             .AddCustomHeader("Strict-Transport-Security", "max-age=5184000; includeSubDomains")
             .AddCustomHeader("X-Frame-Options", "Deny")
+            .AddCustomHeader("Access-Control-Allow-Credentials", "true")
```
tnotheis, line 114: Why is this necessary? As far as I understand, it is only required if you want to send your credentials to **other** sites. But since our UI is delivered by the API-server, this is not the case.

AdminUi/test/AdminUi.Tests.Integration/Models/HttpResponse.cs
```diff
@@ -9,6 +10,8 @@ public class HttpResponse<T>
     public bool IsSuccessStatusCode { get; set; }
     public string? ContentType { get; set; }
     public string? RawContent { get; set; }
+    public IReadOnlyCollection<HeaderParameter>? Headers { get; set; }
+    public CookieCollection? Cookies { get; set; }
```
tnotheis, line 14: Don't use RestSharp classes here. We do not want to expose these to the rest of the tests, so that we are able to switch the used http client without having to touch all the tests.

AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs
```diff
@@ -11,17 +11,43 @@ namespace AdminUi.Tests.Integration.API;
 
 public class BaseApi
 {
-    protected const string ROUTE_PREFIX = "/api/v1";
     private readonly RestClient _client;
 
+    protected const string ROUTE_PREFIX = "/api/v1";
+    private const string TOKEN_HEADER_NAME = "X-XSRF-TOKEN";
+    private const string TOKEN_COOKIE_NAME = "X-XSRF-COOKIE";
```
tnotheis, line 18:
```suggestion
    private const string XSRF_TOKEN_HEADER_NAME = "X-XSRF-TOKEN";
    private const string XSRF_TOKEN_COOKIE_NAME = "X-XSRF-COOKIE";
```

AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs
```diff
@@ -11,17 +11,43 @@ namespace AdminUi.Tests.Integration.API;
...
+    {
+        Task.Run(LoadXSRFTokensAsync).Wait();
+        _client.AddDefaultHeader(TOKEN_HEADER_NAME, _xsrfToken);
+        _client.AddDefaultHeader("Cookie", $"{TOKEN_COOKIE_NAME}={_xsrfCookie}");
+    }
+
+    private async Task LoadXSRFTokensAsync()
+    {
+        var token = await Get<string>("/xsrf", new() { AcceptHeader = "text/plain" });
+        if (token.RawContent != null && token.Cookies != null && token.Cookies.Count > 0)
```
tnotheis, line 43: Shouldn't we throw in the else case so that we know something went wrong?

AdminUi/test/AdminUi.Tests.Integration/Models/HttpResponse.cs
tnotheis: Why did you even add them? They aren't used anywhere. Or am I missing something?

tnotheis (changes requested):
From your comment in the task:
> As per dotnet conventions, AutoValidateAntiforgeryTokenAttribute only requires validation of antiforgery tokens for unsafe HTTP methods, i.e, all methods except GET, HEAD, OPTIONS, and TRACE. For this reason, any GET Controller Actions that perform unsafe operations must be annotated with the ValidateAntiForgeryToken Attribute.

Does this mean that other HTTP methods are annotated by default? Because afaik those are vulnerable as well. 

Regarding SQL Injection: did you check all our raw SQL queries for vulnerabilities?

tnotheis:
Also: did you check pages like https://www.tripwire.com/state-of-security/most-common-website-security-attacks-and-how-to-protect-yourself or https://developer.mozilla.org/en-US/docs/Web/Security/Types_of_attacks?

AdminUi/src/AdminUi/Program.cs
Dannyps: We need to review the proper way to deploy the AdminUi, maybe write a README with instructions for development. In the past few days, I have not been able to get Angular to run when launching the AdminUi in Visual Studio. When I did, however, the API and the UI were still being served from different ports, hence that special header.

AdminUi/test/AdminUi.Tests.Integration/Models/HttpResponse.cs
Dannyps: The Headers attribute is not used anywhere indeed, but the Cookies one is. See below

https://github.com/nmshd/backbone/blob/admin-ui-security/AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs#L45C32-L45C39

AdminUi/src/AdminUi/Program.cs
tnotheis: Okay, but after deployment it is just a single server. So you should only add this header if necessary. You can for example add a correpsonding boolean property to the CorsConfiguration that is then stringified for to pass it to "AddCustomHeader".

AdminUi/test/AdminUi.Tests.Integration/Models/HttpResponse.cs
tnotheis: Okay, then you should use a custom class for the cookies

Dannyps: I'm unsure what you mean by custom class. Changing the client does not entail changing the tests, just the BaseApi.

Dannyps:
>Does this mean that other HTTP methods are annotated by default? Because afaik those are vulnerable as well.

Other methods -- those not considered safe by [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#page-22) (POST, PUT & DELETE) -- are subject to the default annotation. These _should_ be the only vulnerable methods.

>Regarding SQL Injection: did you check all our raw SQL queries for vulnerabilities?

I have updated the task page with this and other information, but yes I have. We're good as long as we keep using parameters properly instead of adding them to the query manually.

AdminUi/test/AdminUi.Tests.Integration/Models/HttpResponse.cs
tnotheis: > Changing the client does not entail changing the tests, just the BaseApi.

At the moment yes. But as soon as you expose something, it **may** be used wherever you want.

> I'm unsure what you mean by custom class.

Something like this:
```c#
public class Cookie
{
    public string Value { get; init; }
}

tnotheis:
> Other methods -- those not considered safe by [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#page-22) (POST, PUT & DELETE) -- are subject to the default annotation. These should be the only vulnerable methods anyway.

"default annotation" means they **are** annotated with the `AutoValidateAntiforgeryTokenAttribute`?

Dannyps:
They aren't explicitly but it's as if they were due to the Filter added when registering the Controllers.

tnotheis:
> They aren't explicitly but it's as if they were due to the Filter added when registering the Controllers.

And that's definitely the case? Did you test that? I'm just asking because it would be really bad if something didn't work. :)

tnotheis (approved):

PR #245, opened 2023-08-11
Limit content width of Admin UI

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/AdminUi/src/AdminUi/ClientApp/src/app/app.component.css
+++ b/AdminUi/src/AdminUi/ClientApp/src/app/app.component.css
@@ -8,6 +8,8 @@
 
 .layout-content {
     padding: 1rem;
+    max-width: 1140px;
+    margin: auto;
 }
 
 .login-layout {
--- a/AdminUi/src/AdminUi/ClientApp/src/app/app.component.css
+++ b/AdminUi/src/AdminUi/ClientApp/src/app/app.component.css
@@ -8,7 +8,7 @@
 
 .layout-content {
     padding: 1rem;
-    max-width: 1140px;
+    max-width: 1350px;
     margin: auto;
 }
 
--- a/AdminUi/src/AdminUi/ClientApp/src/app/components/dashboard/dashboard.component.css
+++ b/AdminUi/src/AdminUi/ClientApp/src/app/components/dashboard/dashboard.component.css
@@ -30,9 +30,9 @@
 }
 
 .panel-icon {
-    font-size: 3rem;
-    width: 3rem;
-    height: 3rem;
+    font-size: 100px;
+    width: 100px;
+    height: 100px;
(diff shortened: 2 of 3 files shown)


JLSRKonk:
**Screenshots Widescreen**

Screenshot 2023-08-11 113437

Screenshot 2023-08-11 113454

Screenshot 2023-08-11 113513

Screenshot 2023-08-11 113530

Screenshot 2023-08-11 113546

Screenshot 2023-08-11 113602

Screenshot 2023-08-11 113617

Screenshot 2023-08-11 113639

Screenshot 2023-08-11 113659

Screenshot 2023-08-11 113727

Screenshot 2023-08-11 113740

Screenshot 2023-08-11 113753

tnotheis:
1. How wide would you have to make the content area to fit the column names of the Identities table on one row?
2. Is it hard to center the dialogs based on the actual content area and not the whole screen? Right now they look a little off.

tnotheis (changes requested):
See above

JLSRKonk:
> 1. How wide would you have to make the content area to fit the column names of the Identities table on one row?
> 2. Is it hard to center the dialogs based on the actual content area and not the whole screen? Right now they look a little off.

1. The content width is currently set to 1140px. If we increase it 1350px, all columns names will fit.
Screenshot 2023-08-23 103156

2. I'm still looking around, but so far, it seems to be a bit of a problem to actually change the center of the dialog.
There doesn't seem to be a way to attach the dialog to a different container, so we would need to manually calculate its left-right positions. This would also partially break the responsiveness, as the calculations would be made when the dialog is called. If we resized screen after opening the dialog, we would need to reopen it in order to recalculate its left-right positions.
Screenshot 2023-08-23 115728
But I'm still investigating this one.

tnotheis:
> > 1. How wide would you have to make the content area to fit the column names of the Identities table on one row?
> > 2. Is it hard to center the dialogs based on the actual content area and not the whole screen? Right now they look a little off.
> 
> 1. The content width is currently set to 1140px. If we increase it 1350px, all columns names will fit.
>    Screenshot 2023-08-23 103156
> 2. I'm still looking around, but so far, it seems to be a bit of a problem to actually change the center of the dialog.
>    There doesn't seem to be a way to attach the dialog to a different container, so we would need to manually calculate its left-right positions. This would also partially break the responsiveness, as the calculations would be made when the dialog is called. If we resized screen after opening the dialog, we would need to reopen it in order to recalculate its left-right positions.
>    Screenshot 2023-08-23 115728
>    But I'm still investigating this one.

1. Looks good.
2. Hm okay, that's sad. But don't spend too much time on it. It's not that important.

tnotheis (approved):

PR #252, opened 2023-08-16
Admin UI frontend logging

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.


AdminUi/src/AdminUi/Controllers/LogsController.cs
```diff
@@ -0,0 +1,81 @@
...
+using ApplicationException = Enmeshed.BuildingBlocks.Application.Abstractions.Exceptions.ApplicationException;
+
+namespace AdminUi.Controllers;
+
+[Route("api/v1/[controller]")]
+[Authorize("ApiKey")]
+public class LogsController : ApiControllerBase
+{
+    private readonly ApplicationOptions _options;
+    private readonly ILoggerFactory _logger;
```
tnotheis, line 18: Why do you label it as a "logger" when in fact it is a "LoggerFactory"?

AdminUi/src/AdminUi/Controllers/LogsController.cs
```diff
@@ -0,0 +1,81 @@
...
+        IMediator mediator, IOptions<ApplicationOptions> options, ILoggerFactory logger) : base(mediator)
+    {
+        _options = options.Value;
+        _logger = logger;
+    }
+
+    [HttpPost]
+    [ProducesResponseType(StatusCodes.Status204NoContent)]
+    [ProducesError(StatusCodes.Status400BadRequest)]
+    public IActionResult CreateLog([FromBody] LogRequest request)
```
tnotheis, line 30:
```suggestion
    public IActionResult CreateLog(LogRequest request)
```

It should automatically be read from the body.

AdminUi/src/AdminUi/Controllers/LogsController.cs
```diff
@@ -0,0 +1,81 @@
...
+
+public enum LogLevel
+{
+    TRACE,
+    DEBUG,
+    INFORMATION,
+    LOG,
+    WARNING,
+    ERROR,
+    CRITICAL
```
tnotheis, line 80: Enum values in C# are usually not capitalized.

AdminUi/test/AdminUi.Tests.Integration/Models/LogRequest.cs
```diff
@@ -0,0 +1,19 @@
...
+
+public enum LogLevel
+{
+    TRACE,
+    DEBUG,
+    INFORMATION,
+    LOG,
+    WARNING,
+    ERROR,
+    CRITICAL
```
tnotheis, line 18: Do caps

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/LogsStepDefinitions.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    [When(@"a POST request is sent to the /Logs endpoint")]
+    public async Task WhenAPOSTRequestIsSentToTheLogsEndpoint()
+    {
+        var createTierRequest = new LogRequest
+        {
+            LogLevel = LogLevel.TRACE,
+            Category = TestDataGenerator.GenerateString(12),
+            MessageTemplate = TestDataGenerator.GenerateString(12),
+            Arguments = Array.Empty<object>()
```
tnotheis, line 27: You should pass something here so that it is tested.

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/LogsStepDefinitions.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    [When(@"a POST request is sent to the /Logs endpoint")]
+    public async Task WhenAPOSTRequestIsSentToTheLogsEndpoint()
+    {
+        var createTierRequest = new LogRequest
+        {
+            LogLevel = LogLevel.TRACE,
+            Category = TestDataGenerator.GenerateString(12),
+            MessageTemplate = TestDataGenerator.GenerateString(12),
```
tnotheis, line 26: Pass some constant data instead of random values.

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/LogsStepDefinitions.cs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    [When(@"a POST request is sent to the /Logs endpoint with an invalid Log Level")]
+    public async Task WhenAPOSTRequestIsSentToTheLogsEndpointWithAnInvalidLogLevel()
+    {
+        var createTierRequest = new LogRequest
+        {
+            LogLevel = (LogLevel)16,
+            Category = TestDataGenerator.GenerateString(12),
+            MessageTemplate = TestDataGenerator.GenerateString(12),
```
tnotheis, line 44: Use constant values

AdminUi/src/AdminUi/ClientApp/src/app/services/logger-server-service/logger-server.service.ts
```diff
@@ -0,0 +1,45 @@
...
+                setHeaders: {
+                    "X-API-KEY": this.authService.getApiKey()!
+                }
+            });
+        }
+
+        return httpRequest;
+    }
+
+    public customiseRequestBody(metadata: INGXLoggerMetadata) {
```
tnotheis, line 30: Shouldn't this be an override as well?
```suggestion
    public override customiseRequestBody(metadata: INGXLoggerMetadata) {
```

AdminUi/src/AdminUi/ClientApp/src/app/services/logger-server-service/logger-server.service.ts
```diff
@@ -0,0 +1,45 @@
...
+})
+export class LoggerServerService extends NGXLoggerServerService {
+    isLoggedIn$: Observable<boolean> | undefined;
+
+    constructor(httpBackend: HttpBackend, ngZone: NgZone, private authService: AuthService) {
+        super(httpBackend, ngZone);
+    }
+
+    protected override alterHttpRequest(httpRequest: HttpRequest<any>): HttpRequest<any> {
+        this.isLoggedIn$ = this.authService.isLoggedIn;
```
tnotheis, line 18: What is the dollar sign at the end of the variable for?

AdminUi/src/AdminUi/ClientApp/src/app/services/logger-server-service/logger-server.service.ts
```diff
@@ -0,0 +1,45 @@
...
+import { Injectable, NgZone } from "@angular/core";
+import { INGXLoggerMetadata, NGXLoggerServerService, NgxLoggerLevel } from "ngx-logger";
+import { AuthService } from "../auth-service/auth.service";
+import { Observable } from "rxjs";
+
+@Injectable({
+    providedIn: "root"
+})
+export class LoggerServerService extends NGXLoggerServerService {
+    isLoggedIn$: Observable<boolean> | undefined;
```
tnotheis, line 11: Why is it an `Observable<boolean>` and not just a `boolean`?

AdminUi/src/AdminUi/ClientApp/src/app/app.module.ts
```diff
@@ -72,6 +76,20 @@ import { ChangeSecretDialogComponent } from "./components/client/change-secret-d
         BrowserAnimationsModule,
         ClipboardModule,
         HttpClientModule,
+        LoggerModule.forRoot(
+            {
+                serverLoggingUrl: environment.apiUrl + "/Logs",
+                level: environment.production ? NgxLoggerLevel.INFO : NgxLoggerLevel.TRACE,
+                serverLogLevel: NgxLoggerLevel.INFO,
```
tnotheis, line 83: Set it to error please.

AdminUi/src/AdminUi/ClientApp/src/app/app.module.ts
```diff
@@ -72,6 +76,20 @@ import { ChangeSecretDialogComponent } from "./components/client/change-secret-d
...
+                enableSourceMaps: true
```
tnotheis, line 84: What's this property for? To get better stacktraces?

tnotheis (changes requested):

tnotheis:
Did you try to add some log statements to see if it actually works?

JLSRKonk:
> Did you try to add some log statements to see if it actually works?

Yes, I tested logging in one of the components and the log is correctly logged to the console and a log request is sent to our logs endpoint with the correct data.

image

Screenshot 2023-08-24 164021

Screenshot 2023-08-24 164125

AdminUi/src/AdminUi/ClientApp/src/app/app.module.ts
tnotheis: Okay, that's what I expected. 👍

tnotheis:
> > Did you try to add some log statements to see if it actually works?
> 
> Yes, I tested logging in one of the components and the log is correctly logged to the console and a log request is sent to our logs endpoint with the correct data.
> 
> image
> 
> Screenshot 2023-08-24 164021
> 
> Screenshot 2023-08-24 164125

The log message in the console looks weird. Is there a way to format it differently?

AdminUi/src/AdminUi/ClientApp/src/app/services/logger-server-service/logger-server.service.ts
tnotheis: 1. Why don't you use the HttpClient then?
2. I still don't understand what the dollar sign is for.

AdminUi/src/AdminUi/ClientApp/src/app/services/logger-server-service/logger-server.service.ts
tnotheis: What about this?

tnotheis (changes requested):

JLSRKonk:
I'm going to try and explain everything in this comment.

`isLoggedIn$`$ is an Observable, and Observables typically have a "$" suffix in order to identify them at first glance.
[Angular Source](https://angular.io/guide/rx-library#naming-conventions-for-observables)

Why is `isLoggedIn$` an Observable?
This is our current implementation of the interceptor:
image
The `isLoggedIn$` is a reference to the `authService.isLoggedIn` method.
image
The `isLoggedIn` method returns the `LoggedIn` variable, which is a BehaviorSubject on the `this.hasApiKey()` method. I assume we return a BehaviorSubject and not a boolean directly for better performance.

Why don't we use the `HttpClient` instead of `HttpBackend` to avoid duplicating code?
Well, the ngx logger library uses the `HttpBackend` for all its logging calls.
The logger-server-service I created for this task extends their `NGXLoggerServerService` and overrides a few of their methods in order to modify the `HttpRequest` that we send, so I can't really swap out `HttpBacked` for `HttpClient` without rewriting their service.
[Service Source Code](https://github.com/dbfannin/ngx-logger/blob/master/src/lib/server/server.service.ts#L16)

And finally, regarding the log message on the console.
We can override the log method from NGXLoggerWriterService to make it look exactly how we want, do you have any suggestions on what the message should look like in the console?
[Service Source Code](https://github.com/dbfannin/ngx-logger/blob/master/src/lib/writer/writer.service.ts#L96)

In the screenshot I used the following logging command:
`this.logger.error("The page {0} has the following description: {1}", this.header, this.displayedColumns);`
Using {0} and {1} for the logger in .NET to auto replace them with the arguments.

tnotheis:
> The isLoggedIn method returns the LoggedIn variable, which is a BehaviorSubject on the this.hasApiKey() method. I assume we return a BehaviorSubject and not a boolean directly for better performance.

Okay. I don't want to say that I understand it but I can live with it. :)

> I can't really swap out HttpBacked for HttpClient without rewriting their service.

I didn't see that the base constructor needs this dependency. What if you simply add another dependency for the HttpClient and ignore the HttpBackend?
Do you know why they even use the HttpBackend? Are there any performance implications when using HttpClient?

> do you have any suggestions on what the message should look like in the console?

Well, just like in the Backend. So the placeholders should be replaced with their values, instead of logging the log message with the placeholders followed by the values.

JLSRKonk:
> > The isLoggedIn method returns the LoggedIn variable, which is a BehaviorSubject on the this.hasApiKey() method. I assume we return a BehaviorSubject and not a boolean directly for better performance.
> 
> Okay. I don't want to say that I understand it but I can live with it. :)
> 
> > I can't really swap out HttpBacked for HttpClient without rewriting their service.
> 
> I didn't see that the base constructor needs this dependency. What if you simply add another dependency for the HttpClient and ignore the HttpBackend? Do you know why they even use the HttpBackend? Are there any performance implications when using HttpClient?
> 
> > do you have any suggestions on what the message should look like in the console?
> 
> Well, just like in the Backend. So the placeholders should be replaced with their values, instead of logging the log message with the placeholders followed by the values.

I was looking the their service source code and they have this comment:

image

I'm not sure if this concern even applies to our project.
If not, I can override their logOnServer() method, and just swap every use of HttpBackend with the HttpClient. This way we won't need to duplicate the interceptor logic here.

tnotheis:
> > > The isLoggedIn method returns the LoggedIn variable, which is a BehaviorSubject on the this.hasApiKey() method. I assume we return a BehaviorSubject and not a boolean directly for better performance.
> > 
> > 
> > Okay. I don't want to say that I understand it but I can live with it. :)
> > > I can't really swap out HttpBacked for HttpClient without rewriting their service.
> > 
> > 
> > I didn't see that the base constructor needs this dependency. What if you simply add another dependency for the HttpClient and ignore the HttpBackend? Do you know why they even use the HttpBackend? Are there any performance implications when using HttpClient?
> > > do you have any suggestions on what the message should look like in the console?
> > 
> > 
> > Well, just like in the Backend. So the placeholders should be replaced with their values, instead of logging the log message with the placeholders followed by the values.
> 
> I was looking the their service source code and they have this comment:
> 
> image
> 
> I'm not sure if this concern even applies to our project. If not, I can override their logOnServer() method, and just swap every use of HttpBackend with the HttpClient. This way we won't need to duplicate the interceptor logic here.

Oh man, I should have looked at your code in more detail. Sorry. I didn't even notice that the method containing the authorization stuff is only altering the http request and not sending it. I guess I'm okay with it then.

AdminUi/src/AdminUi/ClientApp/src/app/services/logger-server-service/logger-server.service.ts
```diff
@@ -27,11 +27,16 @@ export class LoggerServerService extends NGXLoggerServerService {
         return httpRequest;
     }
 
-    public override customiseRequestBody(metadata: INGXLoggerMetadata) {
+    protected override customiseRequestBody(metadata: INGXLoggerMetadata) {
+        let messageTemplate: string = metadata.message;
+        for (let i = 0; i < metadata.additional!.length; i++) {
+            messageTemplate = messageTemplate.replace("%s", `{${i}}`);
```
tnotheis, line 33: What if I want to actually log "%s"? Or what if I use "%d" for example?

You should therefore use the following regex:

```
(?<!%)%[dsj]
```

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/services/logger-server-service/logger-server.service.ts
```diff
@@ -30,7 +30,7 @@ export class LoggerServerService extends NGXLoggerServerService {
     protected override customiseRequestBody(metadata: INGXLoggerMetadata) {
         let messageTemplate: string = metadata.message;
         for (let i = 0; i < metadata.additional!.length; i++) {
-            messageTemplate = messageTemplate.replace("%s", `{${i}}`);
+            messageTemplate = messageTemplate.replace(new RegExp("(?<!%)%[dsj]"), `{${i}}`);
```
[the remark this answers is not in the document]
tnotheis: If yes, feel free to merge.

tnotheis (approved):

PR #254, opened 2023-08-22
Support for sending push notifications to multiple apps

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


tnotheis (changes requested):
As discussed in Rocket Chat, implement the following changes:
1. Rename "AppBundleIdentifier" to "BundleIdentifier" in the ApnsOptions
2. Change the configuration as follows:
```json
{
	"apns": {
		"keys": {
			"my-key-name": {
				"keyId": "...",
				"teamId": "...",
				"privateKey": "..."
			},
			"my-other-key-name": {
				"keyId": "...",
				"teamId": "...",
				"privateKey": "..."
			}
		},
		"bundles": {
			"my-bundle-id": {
			  "keyName": "my-key-name",
			  "serverType": "Production"
			}
		}
	},
	"fcm": {
		"serviceAccounts": {
			"my-serviceAccount-name": {
				"serviceAccountJson": "..."
			},
		},
		"apps": {
			"my-application-id": {
			  "serviceAccountName": "my-serviceAccount-name"
			}
		}
	}
}
```
3. When the new version is started for the first time, all existing device registrations should be deleted. Or, in other words: all registrations without an app id should be deleted.
4. We need to be able to configure a default bundle id. For that, add a "defaultBundleId" property to both "apns" and "fcm", which is used when the user tries to register a device registration without passing an "appId" (btw: the "appId" property has to be optional).

tnotheis (changes requested):
I'll request changes so that you can re-request review after you're done.

tnotheis (changes requested):

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20230818153325_AppIdForPnsRegistrations.cs
```diff
@@ -2,47 +2,48 @@
...
+            name: "IX_Identities_TierId",
+            table: "Identities");
+
+        migrationBuilder.AddColumn<string>(
+            name: "AppId",
+            table: "PnsRegistrations",
+            type: "text",
+            nullable: false);
+
+        migrationBuilder.Sql(""" DELETE FROM "PnsRegistrations" WHERE "AppId" IS NULL """);
```
tnotheis, line 27: Mhh, I told you to FIRST delete all registrations, and THEN add the AppId column. Currently it will fail because you add a non-nullable column even though there are rows that don't have a value for that column and there is no default value.

tnotheis, line 27:
```suggestion
        migrationBuilder.Sql("""DELETE FROM "PnsRegistrations" WHERE "AppId" IS NULL""");
```

Modules/Devices/src/Devices.Infrastructure.Database.SqlServer/Migrations/20230817101526_AppIdForPnsRegistrations.cs
```diff
@@ -2,47 +2,48 @@
...
+            name: "IX_Identities_TierId",
+            table: "Identities");
+
+        migrationBuilder.AddColumn<string>(
+            name: "AppId",
+            table: "PnsRegistrations",
+            type: "nvarchar(max)",
+            nullable: false);
+
+        migrationBuilder.Sql("DELETE FROM [Devices].[PnsRegistrations] WHERE AppId IS NULL");
```
tnotheis, line 27: See above

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20230818153325_AppIdForPnsRegistrations.cs
daniel-almeida-konkconsulting: Need the space between the last quotation marks now that there is no WHERE statement.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/DirectPush/ApplePushNotificationService/ApplePushNotificationServiceConnector.cs
```diff
@@ -25,17 +27,19 @@ public ApplePushNotificationServiceConnector(IHttpClientFactory httpClientFactor
...
-        var jwt = _jwtGenerator.Generate(_options.PrivateKey, _options.KeyId, _options.TeamId);
-
-        var tasks = recipients.Select(device =>
+        var tasks = registrations.Select(pnsRegistration =>
         {
-            var request = new ApnsMessageBuilder(_options.AppBundleIdentifier, $"{_options.Server}{device}", jwt.Value)
+            ValidateRegistration(pnsRegistration);
+            var device = pnsRegistration.Handle.Value;
+            var bundle = _options.Bundles[pnsRegistration.AppId!];
+            var bundleInformation = _options.Keys[bundle.KeyName];
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: I didn't add this because I also require the bundle to retrieve the server. Should I break it down into two methods?

Modules/Devices/test/Devices.Infrastructure.Tests/Tests/DirectPush/ApplePushNotificationServiceConnectorTests.cs
```diff
@@ -18,6 +18,9 @@ namespace Backbone.Modules.Devices.Infrastructure.Tests.Tests.DirectPush;
 
 public class ApplePushNotificationServiceConnectorTests
 {
+    private const string APP_ID = "some-app-id";
+    private const string KEY_NAME = "test-key-name";
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: Just think it's more readable as opposed to having magic strings in both the tests and the CreateConnector method.

Modules/Devices/src/Devices.ConsumerApi/DevicesModule.cs
```diff
@@ -41,8 +43,31 @@ public void ConfigureServices(IServiceCollection services, IConfigurationSection
...
+
+        var supportedApnsAppIds = apnsOptions.GetSupportedBundleIds();
+        var supportedFcmAppIds = fcmOptions.GetSupportedAppIds();
+
+        //var devicesWithInvalidAppId = devicesDbContext.PnsRegistrations.Where(x =>
+        //        (x.Handle.Platform == PushNotificationPlatform.Apns && !supportedApnsAppIds.Contains(x.AppId)) ||
+        //        (x.Handle.Platform == PushNotificationPlatform.Fcm && !supportedFcmAppIds.Contains(x.AppId)))
+        //    .Select(x => x.DeviceId);
+
+        var devicesWithInvalidAppId = devicesDbContext.PnsRegistrations.Where(x => !supportedApnsAppIds.Contains(x.AppId) && !supportedFcmAppIds.Contains(x.AppId)).Select(x => x.DeviceId);
```
tnotheis, line 68: Change this, as discussed via chat.

tnotheis (changes requested):

tnotheis (approved):

PR #267, opened 2023-09-04
Web application factory for integration tests

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


ConsumerApi.Tests.Integration/Features/Tokens/POST.feature
tnotheis, line 1: The formatting in this file was completely messed up. :)

ConsumerApi.Tests.Integration/Support/Dependencies.cs
```diff
@@ -23,17 +21,12 @@ public static IServiceCollection CreateServices()
...
-        var restClient = new RestClient(httpConfig.BaseUrl);
-
-        var tokensApi = new TokensApi(restClient);
-        var challengesApi = new ChallengesApi(restClient);
-
-        services.AddSingleton(tokensApi);
-        services.AddSingleton(challengesApi);
+        services.AddSingleton(new HttpClientFactory(GetWebApplicationFactory()));
+        services.AddTransient<TokensApi>();
+        services.AddSingleton<ChallengesApi>();
```
tnotheis, line 26: Why is the `TokensApi` Transient and the `ChallengesApi` a Singleton?

ConsumerApi.Tests.Integration/Support/Dependencies.cs
```diff
@@ -23,17 +21,12 @@ public static IServiceCollection CreateServices()
...
 
         return services;
     }
+
+    private static CustomWebApplicationFactory<Program> GetWebApplicationFactory() => new();
```
tnotheis, line 31: Same as above: this is a weird method.

tnotheis (changes requested):

.github/workflows/test.yml
```diff
@@ -48,6 +48,8 @@ jobs:
         run: npm install --prefix ./.ci
       - name: Run integration tests
         run: ./.ci/integrationTest.ts
+        env:
+          BASE_ADDRESS: http://localhost:5000
```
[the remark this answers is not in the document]
tnotheis: It's because I like to define them as closest to the place it is used as possible. 

Or in other words: the script should do as much as possible. 

Otherwise there would be several problems, e.g. it wouldn't be set if we ran the script locally.

AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs
```diff
@@ -118,6 +122,13 @@ private async Task<HttpResponse<T>> ExecuteRequest<T>(HttpMethod method, string
             RawContent = responseRawContent
         };
 
+        if (hasCookies)
+            response.Cookies = cookies?.Select(it =>
+            {
+                var test = it.Split('=', 2);
```
tnotheis, line 128: Naming

AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs
```diff
@@ -118,6 +122,13 @@ private async Task<HttpResponse<T>> ExecuteRequest<T>(HttpMethod method, string
             RawContent = responseRawContent
         };
 
+        if (hasCookies)
+            response.Cookies = cookies?.Select(it =>
```
tnotheis, line 126: Isn't `cookies` always not-null if `hasCookies` is true?

AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs
```diff
@@ -118,6 +122,13 @@ private async Task<HttpResponse<T>> ExecuteRequest<T>(HttpMethod method, string
             RawContent = responseRawContent
         };
 
+        if (hasCookies)
```
tnotheis, line 125: I would probably inline `hasCookies`:

```suggestion
        if (httpResponse.Headers.TryGetValues("Set-Cookie", out var cookies))
```

AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs
```diff
@@ -27,14 +31,14 @@ protected BaseApi(IOptions<HttpClientOptions> httpConfiguration, WebApplicationF
...
-        _client.AddDefaultHeader("Cookie", $"{XSRF_TOKEN_COOKIE_NAME}={_xsrfCookie}");
+        _httpClient.DefaultRequestHeaders.Add(XSRF_TOKEN_HEADER_NAME, _xsrfToken);
+        _httpClient.DefaultRequestHeaders.Add("Cookie", $"{XSRF_TOKEN_COOKIE_NAME}={_xsrfCookie}");
     }
 
     private async Task LoadXSRFTokensAsync()
     {
-        var token = await Get<string>("/xsrf", new() { AcceptHeader = "text/plain" });
-        if (token.RawContent != null && token.Cookies != null && token.Cookies.Count > 0)
+        var token = await Get<string>("/xsrf", new RequestConfiguration { AcceptHeader = "text/plain" });
```
tnotheis, line 40: Hmm, wouldn't it be easier if we manually executed this request instead of building so much cookie specific functionality into the `ExecuteRequest` method? We also don't need an `HttpResponse` object here. All we need is the value of the cookie.

tnotheis (changes requested):

AdminUi/test/AdminUi.Tests.Integration/API/BaseApi.cs
daniel-almeida-konkconsulting: I agree, only reason I kept it this way was since it was already approved in a separate PR.
However, the usage of RestClient made things easier.

PR #270, opened 2023-09-05
Re-calculate metric statuses when adding quotas

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/QuotaCreatedForTier/QuotaCreatedForTierIntegrationEventHandler.cs
```diff
@@ -41,6 +44,8 @@ public async Task Handle(QuotaCreatedForTierIntegrationEvent @event)
 
         await _identitiesRepository.Update(identitiesWithTier, CancellationToken.None);
 
+        await _metricStatusesService.RecalculateMetricStatuses(identitiesWithTier.Select(i => i.Address).ToList(), new List<string>() { tierQuotaDefinition.MetricKey.Value }, CancellationToken.None);
```
tnotheis, line 47: Extract two variables for the two parameters please. This line is way too long.

tnotheis, line 47:
```suggestion
        await _metricStatusesService.RecalculateMetricStatuses(identitiesWithTier.Select(i => i.Address).ToList(), new List<string> { tierQuotaDefinition.MetricKey.Value }, CancellationToken.None);
```

Modules/Quotas/src/Quotas.Application/Metrics/MetricStatusesService.cs
```diff
@@ -44,3 +44,8 @@ private static IEnumerable<MetricKey> ParseMetricKeys(IEnumerable<string> metric
         return parsedMetricKeys;
     }
 }
+
+public interface IMetricStatusesService
```
tnotheis, line 48: Why the interface? Afaik FakeItEasy allows mocking classes as well (may be wrong though).

Modules/Quotas/src/Quotas.Application/Tiers/Commands/CreateQuotaForIdentity/Handler.cs
```diff
@@ -38,6 +41,8 @@ public async Task<IndividualQuotaDTO> Handle(CreateQuotaForIdentityCommand reque
 
         _logger.LogTrace($"Successfully created Quota for Identity. Identity Address: '{identity.Address}'");
 
+        await _metricStatusesService.RecalculateMetricStatuses(new List<string>() { identity.Address }, new List<string>() { metric.Key.Value }, cancellationToken);
```
tnotheis, line 44: See comments above

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/QuotaCreatedForTierIntegrationEventHandlerTests.cs
```diff
@@ -1,18 +1,21 @@
...
 using Microsoft.Extensions.Logging;
-using Quotas.Application.Tests.TestDoubles;
 using Xunit;
 
-namespace Quotas.Application.Tests.Tests.Identities;
+namespace Backbone.Modules.Quotas.Application.Tests.Tests.Identities;
 
 public class QuotaCreatedForTierIntegrationEventHandlerTests
 {
+    private IMetricStatusesService _metricStatusesService;
```
tnotheis, line 17: Why is this a field? It's only used in one test. As I said more than once: only use fields if it's really necessary/if it improves the code quality a lot.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/QuotaCreatedForTierIntegrationEventHandlerTests.cs
```diff
@@ -43,9 +45,38 @@ public async void Creates_tier_quota_after_consuming_integration_event()
...
+        A.CallTo(() => identitiesRepository.FindWithTier(tierId, CancellationToken.None, true)).Returns(identities);
+
+        var handler = CreateHandler(identitiesRepository, tierQuotaDefinitionsRepository);
+
+        // Act
+        await handler.Handle(new QuotaCreatedForTierIntegrationEvent(tier.Id, tierQuotaDefinition.Id));
+
+        // Assert
+        A.CallTo(() => _metricStatusesService.RecalculateMetricStatuses(A<List<string>>.That.Matches(x => x.Contains("some-identity-address-one") && x.Contains("some-identity-address-two")),
+            A<List<string>>.That.Contains(tierQuotaDefinition.MetricKey.Value), A<CancellationToken>._)).MustHaveHappened();
```
tnotheis, line 73: Reformat this so that one can actually read it. :)

tnotheis, line 73: What if "some-identity-address-three" was passed as well? (=> check for the length)

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/QuotaCreatedForTierIntegrationEventHandlerTests.cs
```diff
@@ -43,9 +45,38 @@ public async void Creates_tier_quota_after_consuming_integration_event()
         ).MustHaveHappened();
     }
 
+    [Fact]
+    public async void Updates_metric_statuses_after_creating_tier_quota()
+    {
+        // Arrange
+        var tierId = new TierId("TIRFxoL0U24aUqZDSAWc");
+        var tier = new Tier(tierId, "some-tier-name");
```
tnotheis, line 53: This is never used or rather never has to be used (its only usage is `tier.Id`, which can be replaced with `tierId`.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/QuotaCreatedForTierIntegrationEventHandlerTests.cs
```diff
@@ -43,9 +45,38 @@ public async void Creates_tier_quota_after_consuming_integration_event()
...
+
+        var max = 5;
+        var period = QuotaPeriod.Month;
```
tnotheis, line 56: Inline these

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForIdentity/HandlerTests.cs
tnotheis, line 1: See comments in the other tests. I will have a detailed look once those are fixed.

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Application/Metrics/MetricStatusesService.cs
daniel-almeida-konkconsulting: I need either an interface or the method to be virtual or abstract.
Out of the options available, figured the interface was the best fit since we even discussed having it on the MetricStatusesService task.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/QuotaCreatedForTierIntegrationEventHandlerTests.cs
daniel-almeida-konkconsulting: So that each test doesn't have to FakeIt and then pass it to the Handler creation while also allowing for tests that want to use it for assertions, setup, etc. to be able to freely do it.

Modules/Quotas/src/Quotas.Application/Metrics/MetricStatusesService.cs
tnotheis: Okay, agree

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/QuotaCreatedForTierIntegrationEventHandlerTests.cs
tnotheis: You can simply make it optional in the Create method.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/QuotaCreatedForTierIntegrationEventHandlerTests.cs
```diff
@@ -14,29 +14,24 @@ namespace Backbone.Modules.Quotas.Application.Tests.Tests.Identities;
...
+        var tierQuotaDefinition = new TierQuotaDefinition(MetricKey.NumberOfSentMessages, 5, QuotaPeriod.Month);
         var tierQuotaDefinitionsRepository = new FindTierQuotaDefinitionsStubRepository(tierQuotaDefinition);
 
         var firstIdentity = new Identity("some-identity-address-one", tierId);
         var secondIdentity = new Identity("some-identity-address-two", tierId);
         var identities = new List<Identity> { firstIdentity, secondIdentity };
         var identitiesRepository = A.Fake<IIdentitiesRepository>();
         A.CallTo(() => identitiesRepository.FindWithTier(tierId, CancellationToken.None, true)).Returns(identities);
-        var handler = CreateHandler(identitiesRepository, tierQuotaDefinitionsRepository);
+        var handler = CreateHandler(identitiesRepository, tierQuotaDefinitionsRepository, null);
```
tnotheis, line 31: Oh come on... Just make the parameter optional instead of only nullable... ;)

Same for the other file

tnotheis (changes requested):

PR #271, opened 2023-09-05
Default Tier for Clients

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.


JLSRKonk:
The pull request is still missing the fix for the missing TierIds in the cns-transport tests.

JLSRKonk:
**Screenshots**

image

image

image

tnotheis:
The textbox in the UI should be labeled with "Default Tier" instead of just "Tier". We don't assign Tiers to Clients.
Also, the hint below the textbox should explain what the Default Tier is used for.

JLSRKonk:
image

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/ClientDetailsStepDefinitions.cs
```diff
@@ -0,0 +1,62 @@
...
+        _clientId = string.Empty;
+    }
+
+    [Given(@"a Client c")]
+    public async Task GivenAClient()
+    {
+        var createClientRequest = new CreateClientRequest
+        {
+            ClientId = string.Empty,
+            DisplayName = "a-client-display-name",
```
tnotheis, line 24: You should pass as little data as necessary, in order to reduce noise in the tests. The display name isn't required, so you shouldn't pass it.

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/ClientsStepDefinitions.cs
```diff
@@ -53,6 +59,28 @@ public void GivenANonExistentClientC()
         _clientId = "some-non-existent-client-id";
     }
 
+    [Given(@"a Tier t")]
+    public async Task GivenATier()
+    {
+        var createTierQuotaRequest = new CreateTierRequest
```
tnotheis, line 65:
```suggestion
        var createTierRequest = new CreateTierRequest
```

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/ClientsStepDefinitions.cs
```diff
@@ -124,6 +152,60 @@ public async Task WhenAPatchRequestIsSentToTheClientsChangeSecretEndpointForAnIn
         _changeClientSecretResponse.Content.Should().NotBeNull();
     }
 
+    [When(@"a PATCH request is sent to the /Clients/{c.ClientId}/Update endpoint with the tier id t.Id")]
+    public async Task WhenAPatchRequestIsSentToTheClientsUpdateEndpointWithATierId()
+    {
+        var changeClientSecretRequest = new UpdateClientRequest()
```
tnotheis, line 158:
```suggestion
        var updateClientRequest = new UpdateClientRequest()
```

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/ClientsStepDefinitions.cs
```diff
@@ -124,6 +152,60 @@ public async Task WhenAPatchRequestIsSentToTheClientsChangeSecretEndpointForAnIn
...
+        _updateClientResponse = await _clientsApi.UpdateClient(_clientId, requestConfiguration);
+
+        _updateClientResponse.Should().NotBeNull();
+        _updateClientResponse.Content.Should().NotBeNull();
+    }
+
+    [When(@"a PATCH request is sent to the /Clients/{c.ClientId}/Update endpoint with a non-existent tier id")]
+    public async Task WhenAPatchRequestIsSentToTheClientsUpdateEndpointWithAnInexistentTierId()
+    {
+        var changeClientSecretRequest = new UpdateClientRequest()
```
tnotheis, line 176:
```suggestion
        var updateClientRequest = new UpdateClientRequest()
```

AdminUi/test/AdminUi.Tests.Integration/StepDefinitions/ClientsStepDefinitions.cs
```diff
@@ -124,6 +152,60 @@ public async Task WhenAPatchRequestIsSentToTheClientsChangeSecretEndpointForAnIn
...
+        _updateClientResponse = await _clientsApi.UpdateClient(_clientId, requestConfiguration);
+
+        _updateClientResponse.Should().NotBeNull();
+        _updateClientResponse.Content.Should().NotBeNull();
+    }
+
+    [When(@"a PATCH request is sent to the /Clients/{c.ClientId}/Update endpoint")]
+    public async Task WhenAPatchRequestIsSentToTheClientsUpdateEndpointForAnInexistentClient()
+    {
+        var changeClientSecretRequest = new UpdateClientRequest()
```
tnotheis, line 194:
```suggestion
        var updateClientRequest = new UpdateClientRequest()
```

Modules/Devices/src/Devices.Application/Clients/Commands/CreateClients/Handler.cs
```diff
@@ -22,12 +27,27 @@ public async Task<CreateClientResponse> Handle(CreateClientCommand request, Canc
                 throw new ApplicationException(ApplicationErrors.Devices.ClientIdAlreadyExists());
         }
 
+        if (!string.IsNullOrEmpty(request.TierId))
```
tnotheis, line 30: As discussed in chat: the DefaultTier property in the request should never be empty. **Add a validator to validate that.**

Modules/Devices/src/Devices.Application/Clients/Commands/CreateClients/Handler.cs
```diff
@@ -22,12 +27,27 @@ public async Task<CreateClientResponse> Handle(CreateClientCommand request, Canc
...
+        {
+            var tierIdResult = TierId.Create(request.TierId);
+            if (tierIdResult.IsFailure)
+                throw new ApplicationException(ApplicationErrors.Devices.InvalidTierId());
+
+            _ = await _tiersRepository.FindById(tierIdResult.Value, cancellationToken) ?? throw new ApplicationException(GenericApplicationErrors.NotFound(nameof(Tier)));
```
tnotheis, line 36: This is so ugly...
1. You should always request as little data as possible from the database. In this case, all you need is true or false (does the Tier exist or not)
2. Assigning the result of the Find method to a variable just so you can use the `??` operator is a hack and should be avoided.

Modules/Devices/src/Devices.Application/Clients/Commands/CreateClients/Handler.cs
```diff
@@ -22,12 +27,27 @@ public async Task<CreateClientResponse> Handle(CreateClientCommand request, Canc
...
+        else
+        {
+            var basicTier = await _tiersRepository.GetBasicTierAsync(cancellationToken);
+            request.TierId = basicTier.Id.Value;
+        }
+
         var clientSecret = string.IsNullOrEmpty(request.ClientSecret) ? PasswordGenerator.Generate(30) : request.ClientSecret;
         var clientId = string.IsNullOrEmpty(request.ClientId) ? ClientIdGenerator.Generate() : request.ClientId;
         var displayName = string.IsNullOrEmpty(request.DisplayName) ? clientId : request.DisplayName;
+        var tierId = request.TierId;
```
tnotheis, line 47: Why this variable?

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -35,22 +36,20 @@ public async Task<bool> Exists(string clientId, CancellationToken cancellationTo
         return client != null;
     }
 
-    public async Task Add(string clientId, string displayName, string clientSecret, CancellationToken cancellationToken)
+    public async Task Add(string clientId, string displayName, string clientSecret, string tierId, CancellationToken cancellationToken)
```
tnotheis, line 39:
```suggestion
    public async Task Add(string clientId, string displayName, string clientSecret, TierId tierId, CancellationToken cancellationToken)
```

Modules/Devices/src/Devices.Application/Clients/Commands/UpdateClient/Handler.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            _ = await _tiersRepository.FindById(tierIdResult.Value, cancellationToken) ?? throw new ApplicationException(GenericApplicationErrors.NotFound(nameof(Tier)));
+        }
+        else
+        {
+            var basicTier = await _tiersRepository.GetBasicTierAsync(cancellationToken);
+            request.TierId = basicTier.Id.Value;
+        }
+
+        client.TierId = request.TierId;
```
tnotheis, line 37:
```suggestion
        client.ChangeTierId(request.DefaultTier);
```

Modules/Devices/src/Devices.Application/Clients/Commands/UpdateClient/Handler.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        _oAuthClientsRepository = oAuthClientsRepository;
+        _tiersRepository = tiersRepository;
+    }
+
+    public async Task<UpdateClientResponse> Handle(UpdateClientCommand request, CancellationToken cancellationToken)
+    {
+        var client = await _oAuthClientsRepository.Find(request.ClientId, cancellationToken);
+
+        if (!string.IsNullOrEmpty(request.TierId))
```
tnotheis, line 23: `TierId` should be required. **Add a Validator to validate that.**

Modules/Devices/src/Devices.Application/Clients/Commands/UpdateClient/Handler.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            var tierIdResult = TierId.Create(request.TierId);
+            if (tierIdResult.IsFailure)
+                throw new ApplicationException(ApplicationErrors.Devices.InvalidTierId());
+
+            _ = await _tiersRepository.FindById(tierIdResult.Value, cancellationToken) ?? throw new ApplicationException(GenericApplicationErrors.NotFound(nameof(Tier)));
```
tnotheis, line 29: See CreateClient Handler

Modules/Devices/src/Devices.Application/Clients/Commands/CreateClients/Handler.cs
tnotheis, line 1: Just noticed that the folder name is wrong. Should be `CreateClient` instead of `CreateClients`

Modules/Devices/src/Devices.Application/Identities/Commands/CreateIdentity/CreateIdentityCommand.cs
```diff
@@ -6,6 +6,7 @@ namespace Backbone.Modules.Devices.Application.Identities.Commands.CreateIdentit
 public class CreateIdentityCommand : IRequest<CreateIdentityResponse>
 {
     public string ClientId { get; set; }
+    public string TierId { get; set; }
```
tnotheis, line 9: Did you really think it's a good idea that an Identity can decide which Tier it wants to be in?

Modules/Devices/src/Devices.Application/Infrastructure/Persistence/Repository/IOAuthClientsRepository.cs
```diff
@@ -1,14 +1,14 @@
...
-using OpenIddict.EntityFrameworkCore.Models;
+using Backbone.Modules.Devices.Domain.OpenIddict;
 
 namespace Backbone.Modules.Devices.Application.Infrastructure.Persistence.Repository;
 
 public interface IOAuthClientsRepository
 {
     Task<IEnumerable<OAuthClient>> FindAll(CancellationToken cancellationToken);
-    Task<OpenIddictEntityFrameworkCoreApplication> Find(string clientId, CancellationToken cancellationToken);
+    Task<CustomOpenIddictEntityFrameworkCoreApplication> Find(string clientId, CancellationToken cancellationToken);
```
tnotheis, line 9: Why do you now start to return a CustomOpenIddictEntityFrameworkCoreApplication when we return an `OAuthClient` in the other methods?

Modules/Devices/src/Devices.Application/Infrastructure/Persistence/Repository/IOAuthClientsRepository.cs
```diff
@@ -1,14 +1,14 @@
...
     Task<bool> Exists(string clientId, CancellationToken cancellationToken);
-    Task Add(string clientId, string displayName, string clientSecret, CancellationToken cancellationToken);
-    Task Update(OpenIddictEntityFrameworkCoreApplication client, CancellationToken cancellationToken);
+    Task Add(string clientId, string displayName, string clientSecret, string tierId, CancellationToken cancellationToken);
+    Task Update(CustomOpenIddictEntityFrameworkCoreApplication client, CancellationToken cancellationToken);
```
tnotheis, line 12: See above

JLSRKonk:
It seems I didn't commit the fix for the change client secret bug, committed it now.

Modules/Devices/src/Devices.Application/Identities/Commands/CreateIdentity/Handler.cs
```diff
@@ -46,9 +47,10 @@ public async Task<CreateIdentityResponse> Handle(CreateIdentityCommand command,
         if (existingIdentity != null)
             throw new OperationFailedException(ApplicationErrors.Devices.AddressAlreadyExists());
 
-        var basicTier = await _tiersRepository.GetBasicTierAsync(cancellationToken);
+        var tierId = TierId.Create(command.TierId).Value;
+        var tier = await _tiersRepository.FindById(tierId, cancellationToken) ?? throw new OperationFailedException(ApplicationErrors.Devices.InvalidTierId());
```
tnotheis, line 51: Instead of using the passed id, you should use the default tier that was configured for the client. Why else would we have a default tier? 😮

Modules/Devices/src/Devices.Domain/OpenIddict/CustomOpenIddictEntityFrameworkCoreApplication.cs
tnotheis, line 1: I just realized that, if you actually return the OAuthClient from the repository, then you can move the `CustomOpenIddict*` classes to the Devices.Infrastructure project.

tnotheis (changes requested):

Modules/Devices/src/Devices.Application/Infrastructure/Persistence/Repository/IOAuthClientsRepository.cs
JLSRKonk: When we want to update a Client, we need the complete `CustomOpenIddictEntityFrameworkCoreApplication` object for the `_applicationManager.UpdateAsync` method.
We could technically change it so it returns an `OAuthClient`, but then we would need to call the `_applicationManager.FindByClientIdAsync` inside the `Update` and `ChangeClientSecret` methods to get the full object.

Modules/Devices/src/Devices.Application/Clients/Commands/UpdateClient/Handler.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    public Handler(IOAuthClientsRepository oAuthClientsRepository, ITiersRepository tiersRepository)
+    {
+        _oAuthClientsRepository = oAuthClientsRepository;
+        _tiersRepository = tiersRepository;
+    }
+
+    public async Task<UpdateClientResponse> Handle(UpdateClientCommand request, CancellationToken cancellationToken)
+    {
+        var client = await _oAuthClientsRepository.Find(request.ClientId, cancellationToken);
```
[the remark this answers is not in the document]
JLSRKonk: This repository uses the following for changes to clients:
`OpenIddictApplicationManager<CustomOpenIddictEntityFrameworkCoreApplication> _applicationManager`

I don't really think we can add tracking for this.

Modules/Devices/src/Devices.Application/Infrastructure/Persistence/Repository/IOAuthClientsRepository.cs
tnotheis: Another alternative would be implementing your own tracking mechanism. So: in each `Find` method, save the `CustomOpenIddictEntityFrameworkCoreApplication`s in a list. After that you can map them to `OAuthClient`s and return them. 
Then, when `Update` is called, check if the list contains a `CustomOpenIddictEntityFrameworkCoreApplication`. For the unlikely case it doesn't contain one, you can read it using the application manager.

Modules/Devices/src/Devices.Application/Clients/Commands/UpdateClient/Handler.cs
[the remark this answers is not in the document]
tnotheis: True. But in order to stay consistent, we should still add the parameter, even though we cannot use it. Because the caller of the repository shouldn't know about the actual implementation. It may as well be that there is an implementation that DOES support tracking.

Modules/Devices/src/Devices.Domain/OpenIddict/CustomOpenIddictEntityFrameworkCoreScope.cs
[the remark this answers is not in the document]
JLSRKonk: All examples I found of switching the Application class to a custom one also included the creation of all four custom classes we have in this folder, even if only the Application class was the one with a custom field. Should I try removing all the custom classes except the Application and see if it works using default ones instead?

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20230901155527_AddTierIdToClients.cs
```diff
@@ -0,0 +1,65 @@
...
+
+#nullable disable
+
+namespace Backbone.Modules.Devices.Infrastructure.Database.Postgres.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddTierIdToClients : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
```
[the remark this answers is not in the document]
JLSRKonk: Isn't applying the query below enough?
image
I have a version of this query both on the Sql and the Postgres migrations.

Modules/Devices/src/Devices.Application/Clients/Commands/UpdateClient/Handler.cs
```diff
@@ -18,23 +18,17 @@ public Handler(IOAuthClientsRepository oAuthClientsRepository, ITiersRepository
...
+
+        var tierIdResult = TierId.Create(request.DefaultTier);
+        if (tierIdResult.IsFailure)
+            throw new ApplicationException(ApplicationErrors.Devices.InvalidTierIdOrDoesNotExist());
+
+        var tierExists = await _tiersRepository.ExistsWithId(tierIdResult.Value, cancellationToken);
+        if (!tierExists)
+            throw new ApplicationException(ApplicationErrors.Devices.InvalidTierIdOrDoesNotExist());
+
+        client.DefaultTier = request.DefaultTier;
```
tnotheis, line 31: Once again: never add public setters to domain classes.

```suggestion
        client.ChangeDefaultTier(request.DefaultTier);
```

Modules/Devices/src/Devices.Domain/Entities/OAuthClient.cs
```diff
@@ -1,14 +1,14 @@
...
         ClientId = clientId;
         DisplayName = displayName;
-        TierId = tierId;
+        DefaultTier = defaultTier;
     }
 
     public string ClientId { get; }
     public string DisplayName { get; }
-    public string TierId { get; }
+    public string DefaultTier { get; set; }
```
tnotheis, line 13: Always use value objects in domain classes
```suggestion
    public ClientId DefaultTier { get; }
```

Modules/Devices/src/Devices.Application/Identities/Commands/CreateIdentity/Handler.cs
```diff
@@ -47,10 +47,11 @@ public async Task<CreateIdentityResponse> Handle(CreateIdentityCommand command,
...
             throw new OperationFailedException(ApplicationErrors.Devices.AddressAlreadyExists());
 
-        var tierId = TierId.Create(command.TierId).Value;
-        var tier = await _tiersRepository.FindById(tierId, cancellationToken) ?? throw new OperationFailedException(ApplicationErrors.Devices.InvalidTierId());
+        var client = await _oAuthClientsRepository.Find(command.ClientId, cancellationToken);
 
-        var newIdentity = new Identity(command.ClientId, address, command.IdentityPublicKey, tier.Id, command.IdentityVersion);
+        var tierId = TierId.Create(client.DefaultTier).Value;
+
+        var newIdentity = new Identity(command.ClientId, address, command.IdentityPublicKey, tierId, command.IdentityVersion);
```
tnotheis, line 54:
```suggestion
        var newIdentity = new Identity(command.ClientId, address, command.IdentityPublicKey, client.DefaultTier, command.IdentityVersion);
```

Modules/Devices/src/Devices.Application/Identities/Commands/CreateIdentity/Handler.cs
```diff
@@ -60,7 +61,7 @@ public async Task<CreateIdentityResponse> Handle(CreateIdentityCommand command,
 
         _eventBus.Publish(new IdentityCreatedIntegrationEvent(newIdentity));
 
-        _logger.LogTrace($"Successfully published IdentityCreatedIntegrationEvent. Identity Address: {newIdentity.Address}, Tier: {tier.Name}");
+        _logger.LogTrace($"Successfully published IdentityCreatedIntegrationEvent. Identity Address: {newIdentity.Address}, Tier: {tierId}");
```
tnotheis, line 64:
```suggestion
        _logger.LogTrace($"Successfully published IdentityCreatedIntegrationEvent. Identity Address: '{newIdentity.Address}', Tier: '{tierId}'");
```

Modules/Devices/src/Devices.Application/Infrastructure/Persistence/Repository/IOAuthClientsRepository.cs
```diff
@@ -1,14 +1,15 @@
...
 {
     Task<IEnumerable<OAuthClient>> FindAll(CancellationToken cancellationToken);
-    Task<CustomOpenIddictEntityFrameworkCoreApplication> Find(string clientId, CancellationToken cancellationToken);
+    Task<OAuthClient> Find(string clientId, CancellationToken cancellationToken, bool track = false);
     Task<bool> Exists(string clientId, CancellationToken cancellationToken);
-    Task Add(string clientId, string displayName, string clientSecret, string tierId, CancellationToken cancellationToken);
-    Task Update(CustomOpenIddictEntityFrameworkCoreApplication client, CancellationToken cancellationToken);
+    Task Add(string clientId, string displayName, string clientSecret, TierId tierId, CancellationToken cancellationToken);
+    Task Update(OAuthClient client, CancellationToken cancellationToken);
+    Task ChangeClientSecret(OAuthClient client, string clientSecret, CancellationToken cancellationToken);
```
tnotheis, line 13:
```suggestion
    Task ChangeClientSecret(OAuthClient client, string newSecret, CancellationToken cancellationToken);
```

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -13,21 +14,38 @@ namespace Backbone.Modules.Devices.Infrastructure.Persistence.Repository;
 public class OAuthClientsRepository : IOAuthClientsRepository
 {
     private readonly OpenIddictApplicationManager<CustomOpenIddictEntityFrameworkCoreApplication> _applicationManager;
+    private List<CustomOpenIddictEntityFrameworkCoreApplication> _cachedApplications;
```
tnotheis, line 17: You should use a dictionary instead, so that the read has a better performance. It's also easier to use.

tnotheis, line 17: Call it `_trackedApplications`

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -13,21 +14,38 @@ namespace Backbone.Modules.Devices.Infrastructure.Persistence.Repository;
...
+        if (cachedApplicationIndex != -1)
+        {
+            _cachedApplications[cachedApplicationIndex] = application;
+        }
+        else
+        {
+            _cachedApplications.Add(application);
+        }
+
+        return new OAuthClient(application.ClientId!, application.DisplayName!, application.DefaultTier);
```
tnotheis, line 48: You shouldn't update the tracked entity. Once it is tracked. this entity should always be used. Also, you should check the `track` paramter.

```suggestion
        if(_trackedApplications.ContainsKey(clientId)) {
            return _trackedApplications[clientId];
        
        var application = await _applicationManager.FindByClientIdAsync(clientId, cancellationToken);
        
        if(track)
            _trackedApplications[clientId]= application;

        return new OAuthClient(application.ClientId!, application.DisplayName!, application.DefaultTier);
```

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -13,21 +14,38 @@ namespace Backbone.Modules.Devices.Infrastructure.Persistence.Repository;
...
+        _cachedApplications = new List<CustomOpenIddictEntityFrameworkCoreApplication>();
     }
 
     public async Task<IEnumerable<OAuthClient>> FindAll(CancellationToken cancellationToken)
     {
-        var clients = await _applicationManager.ListAsync(applications => applications.Select(c => new OAuthClient(c.ClientId, c.DisplayName, c.TierId)), cancellationToken).ToListAsync(cancellationToken);
-        return clients;
+        var applications = await _applicationManager.ListAsync(cancellationToken: cancellationToken).ToListAsync(cancellationToken);
+
+        _cachedApplications = applications;
```
tnotheis, line 29: Add a `track` parameter and only track the applications if it is set to true.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -13,21 +14,38 @@ namespace Backbone.Modules.Devices.Infrastructure.Persistence.Repository;
...
+
+        return applications.Select(client => new OAuthClient(client.ClientId!, client.DisplayName!, client.DefaultTier)).ToList();
     }
 
-    public async Task<CustomOpenIddictEntityFrameworkCoreApplication> Find(string clientId, CancellationToken cancellationToken)
+    public async Task<OAuthClient> Find(string clientId, CancellationToken cancellationToken, bool track = false)
     {
-        return await _applicationManager.FindByClientIdAsync(clientId, cancellationToken) ?? throw new NotFoundException(nameof(OAuthClient));
+        var application = await _applicationManager.FindByClientIdAsync(clientId, cancellationToken) ?? throw new NotFoundException(nameof(OAuthClient));
```
tnotheis, line 36: Don't throw here. That's the job of the Handler. Just return null. And I know that we (and I include myself here) did that wrong in other places.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -87,4 +94,47 @@ private static string GetPermissions()
...
+        if (cachedApplication != null)
+        {
+            cachedApplication.DefaultTier = client.DefaultTier;
+            await _applicationManager.UpdateAsync(cachedApplication, cancellationToken);
+            return;
+        }
+
+        var application = await _applicationManager.FindByIdAsync(client.ClientId, cancellationToken) ?? throw new NotFoundException(nameof(OAuthClient));
+        application.DefaultTier = client.DefaultTier;
+        await _applicationManager.UpdateAsync(application, cancellationToken);
```
tnotheis, line 110: Don't you notice when you're duplicating code??

```suggestion
        var application = _cachedApplications.FirstOrDefault(ca => ca.ClientId == client.ClientId);
        if (application == null)
        {
            application = await _applicationManager.FindByIdAsync(client.ClientId, cancellationToken)?? throw new NotFoundException(nameof(OAuthClient));
        }

        application.DefaultTier = client.DefaultTier;
        await _applicationManager.UpdateAsync(application, cancellationToken);
```

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -87,4 +94,47 @@ private static string GetPermissions()
...
+        await _applicationManager.DeleteAsync(client, cancellationToken);
+    }
+
+    public async Task ChangeClientSecret(OAuthClient client, string clientSecret, CancellationToken cancellationToken)
+    {
+        var cachedApplication = _cachedApplications.FirstOrDefault(ca => ca.ClientId == client.ClientId);
+        if (cachedApplication != null)
+        {
+            await _applicationManager.UpdateAsync(cachedApplication, clientSecret, cancellationToken);
+            _cachedApplications.Remove(cachedApplication);
```
tnotheis, line 133: Why do you remove it after you changed the secret?

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
```diff
@@ -87,4 +94,47 @@ private static string GetPermissions()
...
+            return;
+        }
+
+        var application = await _applicationManager.FindByIdAsync(client.ClientId, cancellationToken) ?? throw new NotFoundException(nameof(OAuthClient));
+        await _applicationManager.UpdateAsync(application, clientSecret, cancellationToken);
```
tnotheis, line 138: Again: code duplication (see comment on the `Update` method.

Modules/Devices/src/Devices.Application/Clients/Commands/ChangeClientSecret/Handler.cs
```diff
@@ -16,10 +16,9 @@ public async Task<ChangeClientSecretResponse> Handle(ChangeClientSecretCommand r
         var client = await _oAuthClientsRepository.Find(request.ClientId, cancellationToken);
 
         var clientSecret = string.IsNullOrEmpty(request.NewSecret) ? PasswordGenerator.Generate(30) : request.NewSecret;
-        client.ClientSecret = clientSecret;
 
-        await _oAuthClientsRepository.Update(client, cancellationToken);
+        await _oAuthClientsRepository.ChangeClientSecret(client, clientSecret, cancellationToken);
```
tnotheis, line 20: I'd suggest that we change the `ChangeClientSecret` method to accept clientId and the new secret. Then we don't need to read the client only to pass it back to the repository again. This seems kind of unnecessary.

Modules/Devices/test/Devices.Application.Tests/Tests/Clients/Commands/ChangeClientSecret/HandlerTests.cs
```diff
@@ -11,27 +11,22 @@ public class HandlerTests
...
-            TierId = "Some-tier-id"
-        };
+        var client = new OAuthClient("some-client-id", string.Empty, string.Empty);
+
         var newClientSecret = "New-client-secret";
         var command = new ChangeClientSecretCommand(client.ClientId, newClientSecret);
 
         var oAuthClientsRepository = A.Fake<IOAuthClientsRepository>();
-        A.CallTo(() => oAuthClientsRepository.Find(client.ClientId, A<CancellationToken>._)).Returns(client);
+        A.CallTo(() => oAuthClientsRepository.Find(client.ClientId, A<CancellationToken>._, false)).Returns(client);
```
tnotheis, line 20: Remember to delete this after you changed the way we change the client secret.

Modules/Devices/test/Devices.Application.Tests/Tests/Clients/Commands/UpdateClient/HandlerTests.cs
```diff
@@ -1,44 +1,38 @@
...
 
-        var newTier = new Tier(TierName.Create("new-tier-name").Value);
+        var newDefaultTier = new Tier(TierName.Create("new-default-tier").Value);
 
-        var command = new UpdateClientCommand(client.ClientId, newTier.Id);
+        var command = new UpdateClientCommand(client.ClientId, newDefaultTier.Id);
 
         var oAuthClientsRepository = A.Fake<IOAuthClientsRepository>();
-        A.CallTo(() => oAuthClientsRepository.Find(client.ClientId, A<CancellationToken>._)).Returns(client);
+        A.CallTo(() => oAuthClientsRepository.Find(client.ClientId, A<CancellationToken>._, true)).Returns(client);
```
tnotheis, line 22:
```suggestion
        A.CallTo(() => oAuthClientsRepository.Find(client.ClientId, A<CancellationToken>._, A<bool>._)).Returns(client);
```

Modules/Devices/test/Devices.Application.Tests/Tests/Clients/Commands/UpdateClient/HandlerTests.cs
```diff
@@ -1,44 +1,38 @@
...
+using Backbone.Modules.Devices.Domain.Entities;
 using FakeItEasy;
 using Xunit;
 
 namespace Backbone.Modules.Devices.Application.Tests.Tests.Clients.Commands.UpdateClient;
 public class HandlerTests
 {
     [Fact]
-    public async Task Change_Client_Tier_Id()
+    public async Task Change_Default_Tier()
```
tnotheis, line 12: Tests for
* an inexistent tier
* an inexistent client

are missing

Modules/Devices/src/Devices.Infrastructure/OpenIddict/CustomOpenIddictEntityFrameworkCoreApplication.cs
```diff
@@ -1,8 +1,8 @@
 ﻿using OpenIddict.EntityFrameworkCore.Models;
 
-namespace Backbone.Modules.Devices.Domain.OpenIddict;
+namespace Backbone.Modules.Devices.Infrastructure.OpenIddict;
 public class CustomOpenIddictEntityFrameworkCoreApplication : OpenIddictEntityFrameworkCoreApplication<string, CustomOpenIddictEntityFrameworkCoreAuthorization, CustomOpenIddictEntityFrameworkCoreToken>
 {
-    public string TierId { get; set; }
+    public string DefaultTier { get; set; }
```
tnotheis, line 6:
```suggestion
    public TierId DefaultTier { get; set; }
```

You should regenerate the migrations after this. This should change the datatype from varchar to char.

tnotheis (changes requested):

Modules/Devices/src/Devices.Application/Clients/Commands/ChangeClientSecret/Handler.cs
JLSRKonk: We need the client for the ChangeClientSecretResponse. Should the handler only return the ClientId and ClientSecret instead of all the OAuthClient fields?

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
JLSRKonk: What about the Update, Delete and ChangeClientSecret methods? Should they also have their errors moved to their handlers?

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-edit/client-edit.component.ts
```diff
@@ -144,16 +163,30 @@ export class ClientEditComponent {
...
-        let newTierId = "";
+        let defaultTier = "";
         this.tierList.forEach((tier) => {
-            if (tier.id == this.client.tierId) {
-                newTierId = tier.id;
+            if (tier.id == this.client.defaultTier) {
+                defaultTier = tier.id;
                 return;
             }
         });
```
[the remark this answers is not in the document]
JLSRKonk: So I had this scenario in my head, where if we had an invalid DefaultTier associated with a Client (if for example the Tier no longer exists), then the select default tier would appear to have no tier selected as there is no match for this particular Tier. The user would then press the save button, hoping it would use the Basic Tier, but would be met with an error.

This logic behind this loop is the following:
If there isn't a Tier that matches the DefaultTier in the Tier list, then it discards the DefaultTier and uses the empty string defaultTier.
Seeing the defaultTier empty, the loop below would assume no Tier was selected and would use the Basic Tier.
If it does find the DefaultTier in the list, it uses that instead.

I tested this by deleting a Tier and attempting to change the DefaultTier of a Client.
Normally this behaviour shouldn't happen, but it seems the EntityTypeConfiguration does not assume the DefaultTier as a ForeignKey. Should I add the DefaultTier as a ForeignKey to the CustomOpenIddictCoreApplication?

And what's the expected behaviour when deleting a Tier? Should the deletion be denied or should the deletion go through and the Basic Tier become the new DefaultTier for those Applications?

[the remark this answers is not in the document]
tnotheis: 1. If I understand you correctly, the problem you describe can only occur when the following steps are executed in order: 
    1. User1 selects a new default Tier for a client
    2. User2 (or User1 in another tab) deletes the selected Tier
    3. User1 clicks on save
    This will happen once in a lifetime. So we can definitely ignore this in the frontend. Of course, the backend should (and I think already does) handle the case of an inexistent Tier.
2. Yes, you should establish a foreign key relationship on Client.DefaultTier and Tier.Id.
3. It should not be possible to delete a Tier that is used as a default Tier somewhere.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/OAuthClientsRepository.cs
tnotheis: For now we can leave it as it is there. We would need some more refactoring in order to change it there.

Modules/Devices/src/Devices.Application/Clients/Commands/ChangeClientSecret/Handler.cs
tnotheis: Hmm. Okay, ignore my comment. :)

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20230901155527_AddTierIdToClients.cs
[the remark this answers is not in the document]
tnotheis: Sorry, I must have overlooked that.

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-edit/client-edit.component.ts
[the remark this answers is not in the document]
JLSRKonk: Understood.
I added a ForeignKey restraint on the last commit, with a DeleteBehavior.Restrict , so it doesn't allow queries that attempt to remove a Tier with a relation to a Client.
In addition to this, I added a number of clients verification to the Delete Client handler, with a corresponding error.

Modules/Devices/src/Devices.Domain/Entities/OAuthClient.cs
```diff
@@ -1,12 +1,21 @@
...
+        DefaultTier = defaultTier;
     }
 
     public string ClientId { get; }
     public string DisplayName { get; }
+    public TierId DefaultTier { get; private set; }
+
+    public void ChangeDefaultTier(TierId newDefaultTier)
+    {
+        DefaultTier = newDefaultTier;
```
tnotheis, line 19: We should throw a DomainException if the new Tier is equal to the old one.

tnotheis (changes requested):

Modules/Devices/test/Devices.Domain.Tests/OAuthClient/OAuthClientChangeDefaultTier.cs
tnotheis, line 1: The filename doesn't match our naming schema. It should be "OAuthClientTests"

Modules/Devices/test/Devices.Domain.Tests/OAuthClient/OAuthClientChangeDefaultTier.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    [Fact]
+    public void Default_tier_cannot_be_changed_to_same_tier()
+    {
+        // Arrange
+        var tierId = TierId.Generate();
+        var client = new Entities.OAuthClient(string.Empty, string.Empty, tierId);
+
+        // Act
+        var result = client.ChangeDefaultTier(tierId);
```
tnotheis, line 16:
```suggestion
        var error = client.ChangeDefaultTier(tierId);
```

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierCanBeDeleted.cs
```diff
@@ -30,7 +30,7 @@ public void Tier_with_related_identities_cannot_be_deleted()
 
         // Assert
         result.Should().NotBeNull();
-        result.Should().BeEquivalentTo(DomainErrors.CannotDeleteUsedTier(1));
+        result.Should().BeEquivalentTo(DomainErrors.CannotDeleteUsedTier($"The Tier is assigned to one or more Identities. A Tier cannot be deleted if there are Identities assigned to it ({1} found)."));
```
tnotheis, line 33:
```suggestion
        result.Should().BeEquivalentTo(DomainErrors.CannotDeleteUsedTier(""));
        result.Message.Should().Contain("Tier is assigned to one or more Identities");
```

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierCanBeDeleted.cs
```diff
@@ -44,6 +44,6 @@ public void Tier_with_related_clients_cannot_be_deleted()
 
         // Assert
         result.Should().NotBeNull();
-        result.Should().BeEquivalentTo(DomainErrors.CannotDeleteUsedDefaultTier(1));
+        result.Should().BeEquivalentTo(DomainErrors.CannotDeleteUsedTier($"The Tier is used as the default Tier by one or more clients. A Tier cannot be deleted if it is the default Tier of a Client ({1} found)."));
```
tnotheis, line 47: see above

tnotheis (changes requested):

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierCanBeDeleted.cs
JLSRKonk: If we change this, wouldn't we get an error since the tier.CanBeDeleted call still passes a message argument and as such, the result will not be equivalent?
image

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-edit/client-edit.component.ts
```diff
@@ -101,6 +101,9 @@ export class ClientEditComponent {
         this.tierService.getTiers().subscribe({
             next: (data: PagedHttpResponseEnvelope<TierOverview>) => {
                 this.tierList = data.result;
+                if (!this.editMode) {
+                    this.client.defaultTier = this.getDefaultTier();
```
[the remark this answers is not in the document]
JLSRKonk: Makes sense. I have removed the default "Default Tier" and added a disabled check to the save button, so we can only save after a Default Tier is selected by the user.

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-edit/client-edit.component.ts
```diff
@@ -183,23 +176,22 @@ export class ClientEditComponent {
...
+        if (basicTier) {
+            return basicTier.id;
+        }
+
+        this.snackBar.open("Basic Tier not found", "Dismiss", {
+            verticalPosition: "top",
+            horizontalPosition: "center"
+        });
+
+        this.disabled = true;
```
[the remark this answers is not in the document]
JLSRKonk: The GetTiers endpoint should always return a list that includes the Basic Tier. If it couldn't find the Basic Tier in that list, then there was an error somewhere, so I decided it would be best to disable the page so the user doesn't waste their time trying to fill the missing fields.
After the changes above this is no longer relevant, and has been removed.

Modules/Devices/test/Devices.Domain.Tests/Tiers/TierCanBeDeleted.cs
tnotheis: You're right. I thought we already had an Equals methods on the DomainError. I will add it.

tnotheis: Done. When you now use

```csharp
result.Should().Be(DomainErrors.CannotDeleteUsedTier(""));
```
(BeEquivalentTo -> Be) 

it should work.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/TiersRepository.cs
```diff
@@ -83,6 +83,6 @@ public async Task<Tier> GetBasicTierAsync(CancellationToken cancellationToken)
 
     public async Task<IEnumerable<Tier>> FindByIds(IEnumerable<TierId> tierIds, CancellationToken cancellationToken)
     {
-        return await _tiersDbSet.Where(t => tierIds.Contains(t.Id)).ToListAsync(cancellationToken) ?? throw new NotFoundException(nameof(Tier));
+        return await _tiers.Where(t => tierIds.Contains(t.Id)).ToListAsync(cancellationToken) ?? throw new NotFoundException(nameof(Tier));
```
tnotheis, line 86: Just noticed: `ToListAsync` never returns null. So you can remove the exception.

```suggestion
        return await _tiers.Where(t => tierIds.Contains(t.Id)).ToListAsync(cancellationToken) ?? throw new NotFoundException(nameof(Tier));
```

tnotheis (changes requested):

tnotheis (approved):

PR #273, opened 2023-09-06
Assignment of Tier to Identities

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

When the Identity page is loaded, a new Tiers block is available:
image
The currently selected Tier is highlighted.

Changing the pre-selected Tier enables the save button:
image


daniel-almeida-konkconsulting:
Please add some screenshots for this PR.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
```diff
@@ -59,6 +71,16 @@ export class IdentityEditComponent {
         this.getIdentity();
     }
 
+    getAdmissibleTiers() {
+        this.tierService.getTiers().subscribe({
+            next: (tiers) => {
+                this.tiers = tiers.result;
+                this.tier = this.tiers.filter((t) => t.id === this.identity.tierId)[0];
```
daniel-almeida-konkconsulting, line 78: Is this operation safe? Shouldn't we check the result/length of the collection post filter statement?

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
```diff
@@ -99,6 +126,32 @@ export class IdentityEditComponent {
...
+                        horizontalPosition: "center"
+                    });
+
+                    this.loading = true;
+                    setTimeout(() => {
+                        this.getIdentity();
+                    }, 2000);
+                },
+                error: (err: any) => {
+                    this.loading = false;
```
daniel-almeida-konkconsulting, line 144: When is the loading animation turned on?
I see it for the waiting mechanism while it's "reloading" but not before the initial request.
Shouldn't we have `this.loading = true` as the first statement in the current method?

AdminUi/src/AdminUi/ClientApp/src/app/services/identity-service/identity.service.ts
```diff
@@ -5,6 +5,7 @@ import { HttpResponseEnvelope } from "src/app/utils/http-response-envelope";
 import { PagedHttpResponseEnvelope } from "src/app/utils/paged-http-response-envelope";
 import { environment } from "src/environments/environment";
 import { Quota } from "../quotas-service/quotas.service";
+import { Tier } from "../tier-service/tier.service";
```
daniel-almeida-konkconsulting, line 8: Could probably avoid this reference by just passing `tier.id` to the `updateIdentityTier` method.

Modules/Quotas/test/Quotas.Application.Tests/Tests/IntegrationEvents/TierDeleted/TierOfIdentityChangedIntegrationEventHandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using Xunit;
+
+namespace Backbone.Modules.Quotas.Application.Tests.Tests.IntegrationEvents.TierDeleted;
+public class TierOfIdentityChangedIntegrationEventHandlerTests
+{
+    private readonly ILogger<TierOfIdentityChangedIntegrationEventHandler> _logger;
+
+    public TierOfIdentityChangedIntegrationEventHandlerTests()
+    {
+        _logger = A.Fake<ILogger<TierOfIdentityChangedIntegrationEventHandler>>();
```
daniel-almeida-konkconsulting, line 20: Is this used?

daniel-almeida-konkconsulting (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
daniel-almeida-konkconsulting: In that case I would simply use `this.tiers.find((t) => t.id === this.identity.tierId)!`

Modules/Quotas/test/Quotas.Application.Tests/Tests/IntegrationEvents/TierDeleted/TierOfIdentityChangedIntegrationEventHandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            NewTier = newTier.Id,
+            IdentityAddress = identity.Address
+
+        };
+
+        // Act
+        await handler.Handle(tierDeletedIntegrationEvent);
+
+        // Assert
+        A.CallTo(metricStatusesService).Where(x => x.Method.Name == nameof(metricStatusesService.RecalculateMetricStatuses)).WithAnyArguments().MustHaveHappened();
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: I would still change it since, despite the overloads, you are 100% sure on what method is called and can afford to be more specific due to that. How are we restricting the service by restricting a specific test?

daniel-almeida-konkconsulting:
> # Readiness checklist
> * [ ]  I added/updated unit tests.
> * [ ]  I added/updated integration tests.
> * [x]  I ensured that the PR title is good enough for the changelog.
> * [x]  I labeled the PR.
> 
> When the Identity page is loaded, a new Tiers block is available: image The currently selected Tier is highlighted.
> 
> Changing the pre-selected Tier enables the save button: image

I would probably place the tier selection under the Created At field, in the details section.

Modules/Quotas/test/Quotas.Application.Tests/Tests/IntegrationEvents/TierDeleted/TierOfIdentityChangedIntegrationEventHandlerTests.cs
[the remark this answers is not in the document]
Dannyps: A developer could change the service to pass the MetricKeys as strings. It's really not that relevant,, I suppose.

daniel-almeida-konkconsulting (approved):

tnotheis:
> I would probably place the tier selection under the Created At field, in the details section.

I agree.

tnotheis:
And you should use the enmeshed-orange instead of the yellow-ish color on the active Tier

tnotheis:
And: you didn't select that you added/updated integration tests, even though you actually added tests.

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            throw new ApplicationException(new ApplicationError(newTierId.Error.Code, newTierId.Error.Message));
+        }
+
+        var identity = await _identitiesRepository.FindByAddress(request.Address, cancellationToken) ?? throw new NotFoundException(nameof(Identity));
+
+        var oldTier = await _tiersRepository.FindById(identity.TierId, cancellationToken) ?? throw new NotFoundException(nameof(Tier));
+        var newTier = await _tiersRepository.FindById(newTierId.Value, cancellationToken) ?? throw new NotFoundException(nameof(Tier));
+
+        identity.TierId = newTier.Id;
```
tnotheis, line 37: Use a domain method. No accessible setters in Domain classes!

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/UpdateIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        A.CallTo(() => tiersRepository.FindById(newTier.Id, A<CancellationToken>._)).Returns(newTier);
+
+        var handler = CreateHandler(identitiesRepository, tiersRepository, eventBus);
+        var request = MakeRequest(newTier, identity);
+
+        // Act
+        var result = await handler.Handle(request, CancellationToken.None);
+
+        // Assert
+        result.TierId.Should().Be(newTier.Id);
```
tnotheis, line 40: Assert that the Update method is called with the correct tier id

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/UpdateIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,138 @@
...
+using Enmeshed.UnitTestTools.Extensions;
+using FakeItEasy;
+using FluentAssertions;
+using Xunit;
+
+namespace Backbone.Modules.Devices.Application.Tests.Tests.Identities.Commands.UpdateIdentity;
+public class HandlerTests
+{
+    [Fact]
+    public async void Command_runs_successfully_updates_identity()
```
tnotheis, line 17:
```suggestion
    public async void Updates_the_identity_in_the_database()
```

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/UpdateIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        // Act
+        var result = await handler.Handle(request, CancellationToken.None);
+
+        // Assert
+        result.TierId.Should().Be(newTier.Id);
+    }
+
+    [Fact]
+    public async void Command_runs_successfully_publishes_IntegrationEvent()
```
tnotheis, line 44:
```suggestion
    public async void Publishes_TierOfIdentityChangedIntegrationEvent()
```

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/UpdateIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        // Act
+        var result = await handler.Handle(request, CancellationToken.None);
+
+        // Assert
+        A.CallTo(() => eventBus.Publish(A<TierOfIdentityChangedIntegrationEvent>._)).MustHaveHappened();
+    }
+
+    [Fact]
+    public async void Command_fails_when_identity_is_missing()
```
tnotheis, line 71:
```suggestion
    public async void Fails_when_identity_does_not_exist()
```

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/UpdateIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        // Arrange
+        var identitiesRepository = A.Fake<IIdentitiesRepository>();
+        var tiersRepository = A.Fake<ITiersRepository>();
+        var eventBus = A.Fake<IEventBus>();
+
+        var oldTier = new Tier(TierName.Create("Tier name").Value);
+        var newTier = new Tier(TierName.Create("Tier name").Value);
+
+        var identity = new Identity(TestDataGenerator.CreateRandomDeviceId(), TestDataGenerator.CreateRandomIdentityAddress(), new byte[] { 1, 1, 1, 1, 1 }, oldTier.Id, 1);
+
```
tnotheis, line 82: I would make the return value of the identitiesRepository explicit.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/UpdateIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        Func<Task> acting = async () => await handler.Handle(request, CancellationToken.None);
+
+        // Assert
+        var exception = acting.Should().AwaitThrowAsync<NotFoundException>().Which;
+        exception.Message.Should().StartWith("Identity");
+        exception.Code.Should().Be("error.platform.recordNotFound");
+    }
+
+    [Fact]
+    public async void Command_fails_when_tier_is_missing()
```
tnotheis, line 99:
```suggestion
    public async void Fails_when_tier_does_not_exist()
```

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/UpdateIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,138 @@
...
+    {
+        // Arrange
+        var identitiesRepository = A.Fake<IIdentitiesRepository>();
+        var tiersRepository = A.Fake<ITiersRepository>();
+        var eventBus = A.Fake<IEventBus>();
+
+        var oldTier = new Tier(TierName.Create("Tier name").Value);
+        var newTier = new Tier(TierName.Create("Tier name").Value);
+
```
tnotheis, line 108: I would make the return value of the tiersRepository explicit.

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierOfIdentityChanged/TierOfIdentityChangedIntegrationEventHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    public async Task Handle(TierOfIdentityChangedIntegrationEvent @event)
+    {
+        var identity = await _identitiesRepository.Find(@event.IdentityAddress, CancellationToken.None, track: true);
+        var newTier = await _tiersRepository.Find(@event.NewTier, CancellationToken.None, track: true);
+        identity.ChangeTier(newTier);
+        await _identitiesRepository.Update(identity, CancellationToken.None);
+        await _metricStatusesService.RecalculateMetricStatuses(
+            new List<string>() { identity.Address },
+            identity.MetricStatuses.Select(x => x.MetricKey).ToList(),
```
tnotheis, line 27: What if the new Tier has a Quota on a Metric the old Tier didn't have a Metric on? I think in that case there won't be an existing `MetricStatus`.
So you should rather don't pass anything, which should result in all metric statuses being recalculated.

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierOfIdentityChanged/TierOfIdentityChangedIntegrationEventHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+            CancellationToken.None
+            );
```
tnotheis, line 29: Formatting

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierOfIdentityChanged/TierOfIdentityChangedIntegrationEventHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+        _identitiesRepository = identitiesRepository;
+        _tiersRepository = tiersRepository;
+        _metricStatusesService = metricStatusesService;
+    }
+
+    public async Task Handle(TierOfIdentityChangedIntegrationEvent @event)
+    {
+        var identity = await _identitiesRepository.Find(@event.IdentityAddress, CancellationToken.None, track: true);
+        var newTier = await _tiersRepository.Find(@event.NewTier, CancellationToken.None, track: true);
+        identity.ChangeTier(newTier);
```
tnotheis, line 23: I know you implemented it as I requested, but I have a suggestion:

How about this:

```csharp
await identity.ChangeTier(newTier, metricCalculatorFactory, CancellationToken.None);
```
?

The Identity should then automatically recalculate the MetricStatuses (without the MetricStatusesService)

Modules/Quotas/test/Quotas.Application.Tests/Tests/IntegrationEvents/TierDeleted/TierOfIdentityChangedIntegrationEventHandlerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+using Enmeshed.UnitTestTools.Data;
+using FakeItEasy;
+using FluentAssertions;
+using Xunit;
+
+namespace Backbone.Modules.Quotas.Application.Tests.Tests.IntegrationEvents.TierDeleted;
+public class TierOfIdentityChangedIntegrationEventHandlerTests
+{
+    [Fact]
+    public async Task Handler_updates_identity_with_new_tier()
```
tnotheis, line 16: These are Handler specific tests. So who should update the identity if not the Handler?
```suggestion
    public async Task Updates_identity_with_new_tier()
```

Modules/Quotas/test/Quotas.Application.Tests/Tests/IntegrationEvents/TierDeleted/TierOfIdentityChangedIntegrationEventHandlerTests.cs
tnotheis, line 1: Remember to update the tests after you extracted the domain logic to the Identity class.

tnotheis (changes requested):

Dannyps:
> And you should use the enmeshed-orange instead of the yellow-ish color on the active Tier

I assumed the whole colour theme had been set up... I'm adding the accent colour, but I'm not sure about the rest (warning, error, etc.)

image

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
```diff
@@ -0,0 +1,44 @@
...
+        var newTierId = TierId.Create(request.TierId);
+        if (!newTierId.IsSuccess)
+        {
+            throw new ApplicationException(new ApplicationError(newTierId.Error.Code, newTierId.Error.Message));
+        }
+
+        var identity = await _identitiesRepository.FindByAddress(request.Address, cancellationToken) ?? throw new NotFoundException(nameof(Identity));
+
+        var oldTier = await _tiersRepository.FindById(identity.TierId, cancellationToken) ?? throw new NotFoundException(nameof(Tier));
+        var newTier = await _tiersRepository.FindById(newTierId.Value, cancellationToken) ?? throw new NotFoundException(nameof(Tier));
```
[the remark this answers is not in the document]
Dannyps: I did this change, but I'm not very happy with the result. Please advise.

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierOfIdentityChanged/TierOfIdentityChangedIntegrationEventHandler.cs
Dannyps: Did you mean `metricCalculatorFactory` or `metricStatusesService`?

Dannyps: I believe I interpreted that correctly.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
```diff
@@ -68,6 +90,7 @@ export class IdentityEditComponent {
                     this.identity = data.result;
                     this.groupQuotasByMetricForTable();
                     this.devicesTableData = this.identity.devices;
+                    this.getAdmissibleTiers();
```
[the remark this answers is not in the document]
tnotheis: I don't see anything that changed here.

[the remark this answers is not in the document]
tnotheis: Oh, just saw that you renamed it to `loadAdmissibleTiers.`. But no. 2 still remains true: it fetches available/admissible Tiers as well as the current tier of the Identity.

AdminUi/src/AdminUi/ClientApp/src/app/services/identity-service/identity.service.ts
```diff
@@ -67,3 +67,7 @@ export interface IdentityOverview {
     identityVersion: string;
     numberOfDevices: number;
 }
+
+export interface UpdateTierDTO {
```
tnotheis, line 71:
```suggestion
export interface UpdateTierRequest {
```

AdminUi/src/AdminUi/Controllers/IdentitiesController.cs
```diff
@@ -103,10 +100,10 @@ public async Task<IActionResult> GetIdentityByAddress([FromRoute] string address
     [HttpPut("{identityAddress}")]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesError(StatusCodes.Status404NotFound)]
-    public async Task<IActionResult> PutIdentityTierByAddress([FromRoute] string identityAddress, [FromBody] UpdateIdentityTierRequest request, CancellationToken cancellationToken)
+    public async Task<IActionResult> PutIdentityByAddress([FromRoute] string identityAddress, [FromBody] UpdateIdentityTierRequest request, CancellationToken cancellationToken)
```
tnotheis, line 103:
```suggestion
    public async Task<IActionResult> UpdateIdentity([FromRoute] string identityAddress, [FromBody] UpdateIdentityTierRequest request, CancellationToken cancellationToken)
```

AdminUi/src/AdminUi/Controllers/IdentitiesController.cs
```diff
@@ -103,10 +100,10 @@ public async Task<IActionResult> GetIdentityByAddress([FromRoute] string address
...
     {
         var command = new UpdateIdentityCommand() { Address = identityAddress, TierId = request.TierId };
-        await _mediator.Send<Backbone.Modules.Devices.Domain.Entities.Identity>(command, cancellationToken);
+        await _mediator.Send<Identity>(command, cancellationToken);
```
tnotheis, line 106: Just noticed: Why the hell do you return an Identity from the command??? We NEVER EVER return domain objects from the application layer! That's one of the most important rules.
And you don't even use the return value here anyway.

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
```diff
@@ -0,0 +1,48 @@
...
+    public Handler(IIdentitiesRepository identitiesRepository, ITiersRepository tiersRepository, IEventBus eventBus)
+    {
+        _identitiesRepository = identitiesRepository;
+        _tiersRepository = tiersRepository;
+        _eventBus = eventBus;
+    }
+
+    public async Task<Identity> Handle(UpdateIdentityCommand request, CancellationToken cancellationToken)
+    {
+        var newTierId = TierId.Create(request.TierId);
```
tnotheis, line 27: This is a result, not a tier id...
```suggestion
        var newTierIdResult = TierId.Create(request.TierId);
```

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        var identity = await _identitiesRepository.FindByAddress(request.Address, cancellationToken) ?? throw new NotFoundException(nameof(Identity));
+
+        var tiers = await _tiersRepository.FindByIds(new List<TierId>() { identity.TierId, newTierId.Value }, cancellationToken);
+
+        var oldTier = tiers.SingleOrDefault(t => t.Id == identity.TierId) ?? throw new NotFoundException(nameof(Tier));
+        var newTier = tiers.SingleOrDefault(t => t.Id == newTierId.Value) ?? throw new NotFoundException(nameof(Tier));
+
+        identity.SetTier(newTier.Id);
```
tnotheis, line 40:
```suggestion
        identity.ChangeTier(newTier.Id);
```

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
```diff
@@ -0,0 +1,48 @@
...
+        if (!newTierId.IsSuccess)
+        {
+            throw new ApplicationException(new ApplicationError(newTierId.Error.Code, newTierId.Error.Message));
+        }
+
+        var identity = await _identitiesRepository.FindByAddress(request.Address, cancellationToken) ?? throw new NotFoundException(nameof(Identity));
+
+        var tiers = await _tiersRepository.FindByIds(new List<TierId>() { identity.TierId, newTierId.Value }, cancellationToken);
+
+        var oldTier = tiers.SingleOrDefault(t => t.Id == identity.TierId) ?? throw new NotFoundException(nameof(Tier));
```
tnotheis, line 37: You should call `Single` here. The Tier of an Identity should always exist. And if it doesn't, it is actually an "internal server error" and the resulting 500 status code is correct.
If you return a 404, the caller assumes that it is his own fault (due to the 4XX status range)

Modules/Devices/src/Devices.Application/Infrastructure/Persistence/Repository/ITiersRepository.cs
```diff
@@ -13,4 +13,5 @@ public interface ITiersRepository
     Task<Tier> FindById(TierId tierId, CancellationToken cancellationToken);
     Task Remove(Tier tier);
     Task<int> GetNumberOfIdentitiesAssignedToTier(Tier tier, CancellationToken cancellationToken);
+    Task<IEnumerable<Tier>> FindByIds(IEnumerable<TierId> tiers, CancellationToken cancellationToken);
```
tnotheis, line 16:
```suggestion
    Task<IEnumerable<Tier>> FindByIds(IEnumerable<TierId> ids, CancellationToken cancellationToken);
```

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
```diff
@@ -0,0 +1,48 @@
...
+        var newTier = tiers.SingleOrDefault(t => t.Id == newTierId.Value) ?? throw new NotFoundException(nameof(Tier));
+
+        identity.SetTier(newTier.Id);
+
+        await _identitiesRepository.Update(identity, cancellationToken);
+
+        _eventBus.Publish(new TierOfIdentityChangedIntegrationEvent(identity, oldTier, newTier));
```
tnotheis, line 44: Only trigger the event if the Tier actually changed (oldTier != newTier). 
Add a test for this.

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierOfIdentityChanged/TierOfIdentityChangedIntegrationEventHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+        _tiersRepository = tiersRepository;
+        _metricCalculatorFactory = metricCalculatorFactory;
+    }
+
+    public async Task Handle(TierOfIdentityChangedIntegrationEvent @event)
+    {
+        var identity = await _identitiesRepository.Find(@event.IdentityAddress, CancellationToken.None, track: true);
+        var newTier = await _tiersRepository.Find(@event.NewTier, CancellationToken.None, track: true);
+
+        await identity.ChangeTierAsync(newTier, _metricCalculatorFactory, CancellationToken.None);
```
tnotheis, line 24: Don't. Add. The. Async. Suffix!

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -129,4 +129,18 @@ private IEnumerable<Quota> GetAppliedQuotasForMetric(MetricKey metric)
         var appliedQuotas = allQuotasOfMetric.Where(q => q.Weight == highestWeight).ToArray();
         return appliedQuotas;
     }
+
+    public async Task ChangeTierAsync(Tier newTier, MetricCalculatorFactory metricCalculatorFactory, CancellationToken cancellationToken)
+    {
+        _tierQuotas.RemoveAll(_ => true);
+        _metricStatuses.RemoveAll(_ => true);
```
tnotheis, line 136: Wtf
```suggestion
        _tierQuotas.Clear();
        _metricStatuses.Clear();
```

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -129,4 +129,18 @@ private IEnumerable<Quota> GetAppliedQuotasForMetric(MetricKey metric)
         var appliedQuotas = allQuotasOfMetric.Where(q => q.Weight == highestWeight).ToArray();
         return appliedQuotas;
     }
+
+    public async Task ChangeTierAsync(Tier newTier, MetricCalculatorFactory metricCalculatorFactory, CancellationToken cancellationToken)
+    {
```
tnotheis, line 134: To make sure you don't do any unnecessary work, you should check if the newTier actually differs from the current one and return if they are the same.

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -129,4 +129,18 @@ private IEnumerable<Quota> GetAppliedQuotasForMetric(MetricKey metric)
...
+        _tierQuotas.RemoveAll(_ => true);
+        _metricStatuses.RemoveAll(_ => true);
+
+        TierId = newTier.Id;
+        foreach (var tierQuotaDefinition in newTier.Quotas)
+        {
+            AssignTierQuotaFromDefinition(tierQuotaDefinition);
+        }
+        var metricKeys = newTier.Quotas.Select(q => q.MetricKey);
```
tnotheis, line 143: What if there are multiple quotas on the same metric?
=> the same MetricStatus is updated multiple times
=> unnecessary work

Modules/Quotas/test/Quotas.Application.Tests/Tests/IntegrationEvents/TierDeleted/TierOfIdentityChangedIntegrationEventHandlerTests.cs
```diff
@@ -0,0 +1,142 @@
...
+            IdentityAddress = identity.Address
+        };
+
+        // Act
+        await handler.Handle(tierDeletedIntegrationEvent);
+
+        // Assert
+        identity.MetricStatuses.Should().HaveCount(1);
+        identity.MetricStatuses.First().MetricKey.Should().Be(newTier.Quotas.First().MetricKey);
+    }
```
tnotheis, line 136: These should be tests of the `Identity` class.

tnotheis (changes requested):

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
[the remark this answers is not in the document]
tnotheis: I'm okay with it. Besides the minor comments I wrote as part of the review.

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -129,4 +129,18 @@ private IEnumerable<Quota> GetAppliedQuotasForMetric(MetricKey metric)
...
+        await UpdateMetricStatuses(metricKeys, metricCalculatorFactory, cancellationToken);
```
tnotheis, line 144: As I said in a comment of the previous review, you should recalculate ALL metric statuses. Because what if the old tier defined a quota that doesn't exist in the new tier? Then the MetricStatus may be exhausted even though it shouldn't be.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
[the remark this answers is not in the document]
Dannyps: Number 2 is true for `loadIdentityAndTier`. I have renamed it to `loadIdentityAndTiers`.

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
Dannyps: If they are the same, should the method simply return, or should we thrown an exception?

tnotheis: I think we should throw. Just to make sure the caller notices that there was something wrong. It may be that they selected the wrong tier by mistake.

Dannyps: It won't get to the domain code because we're checking in the handler before. I was thinking more about future code using this function.

tnotheis: You should remove the check in the handler. Always put as much logic as possible into Domain

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateIdentity/Handler.cs
tnotheis: Right

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
tnotheis: Didn't realize that you clear the metric statuses. This fixes the issue I described before.

But it introduces another one: you don't take individual Quotas into account. Because imagine an individual quota on Metric M was exhausted before the Tier change. Now you change the Tier, which deletes all metric statuses. Then you only recalculate the ones from the Tier. But if the new Tier doesn't have a Quota on Metric M, it will be unexhausted after the save.

Dannyps: Well spotted. What do you think about a method like this?

```c#
private async Task UpdateAllMetricStatuses(MetricCalculatorFactory factory,CancellationToken cancellationToken)
{
    var metricKeys = _tierQuotas.Select(q => q.MetricKey).Union(_individualQuotas.Select(q => q.MetricKey)).Distinct();
    await UpdateMetricStatuses(metricKeys, factory, cancellationToken);
}
```

tnotheis: I don't think it's ideal, but we can go with it as an intermediate solution.  You should clear the metric keys in here though.

tnotheis (changes requested):

tnotheis (approved):

PR #275, opened 2023-09-06
First request on each route has a very long response time

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.


JLSRKonk:
In order to create and update the CompiledModels, we need to run the following commands from inside the AdminUi project folder:

`dotnet ef dbcontext optimize -c AdminUiDbContext -o ..\AdminUi.Infrastructure\CompiledModels -n AdminUi.Infrastructure.CompiledModels`

`dotnet ef dbcontext optimize -c ChallengesDbContext -o ..\..\..\Modules\Challenges\src\Challenges.Infrastructure\CompiledModels -n Challenges.Infrastructure.CompiledModels`

Then the CompiledModels will be stored in the Infrastructure project of the DbContext.

tnotheis:
> In order to create and update the CompiledModels, we need to run the following commands from inside the AdminUi project folder:
> 
> `dotnet ef dbcontext optimize -c AdminUiDbContext -o ..\AdminUi.Infrastructure\CompiledModels -n AdminUi.Infrastructure.CompiledModels`
> 
> `dotnet ef dbcontext optimize -c ChallengesDbContext -o ..\..\..\Modules\Challenges\src\Challenges.Infrastructure\CompiledModels -n Challenges.Infrastructure.CompiledModels`
> 
> Then the CompiledModels will be stored in the Infrastructure project of the DbContext.

You should add the command to the script for adding migrations, so that the compiled models are regenerated every time a migration is added.

scripts/windows/efcore/add_migration.ps1
```diff
@@ -8,6 +8,15 @@ $environment="dbmigrations-" + $provider.ToLower()
 $repoRoot = git rev-parse --show-toplevel
 $dbContextName = "${moduleName}DbContext"
 $startupProject = "$repoRoot\ConsumerApi"
+$adminUiProject = "$repoRoot\AdminUi\src\AdminUi"
+
+function CompileModelsAdminUi {
+    $cmdAdminUiDbContext = "dotnet ef dbcontext optimize --project $adminUiProject --context AdminUiDbContext --output-dir $repoRoot\AdminUi\src\AdminUi.Infrastructure\CompiledModels --namespace AdminUi.Infrastructure.CompiledModels"
+    $cmdDevicesDbContext = "dotnet ef dbcontext optimize --project $adminUiProject --context DevicesDbContext --output-dir $repoRoot\Modules\Devices\src\Devices.Infrastructure\CompiledModels --namespace Devices.Infrastructure.CompiledModels"
+    Write-Host "Compiling models for '$adminUiProject'..."
```
tnotheis, line 16: Pleaaaase, stop wasting my time with such **unnecessary** mistakes...! The log message says that you're compiling models for the Admin UI project. But then you compile them for both Admin UI AND the Devices module... Same for the function name.

However. The change was unnecessary anyway. I don't understand why you only compile those two DbContexts now. We need this for all DbContexts. And please don't start copying the command X times. Instead, do it only for the db context for which there is currently a migration being generated.

tnotheis (changes requested):

scripts/windows/efcore/add_migration.ps1
JLSRKonk: From what I understood, we should only need to run the optimize command on the project we are having issues (in this case the adminUi project). In a normal situation, we would only need to run the following command, without specifying the DbContext:

`dotnet ef dbcontext optimize --project $adminUiProject --output-dir $repoRoot\AdminUi\src\AdminUi.Infrastructure\CompiledModels --namespace AdminUi.Infrastructure.CompiledModels`

But after running the above command, dotnet will say that there are currently two DbContexts in use for the adminUi project, and that we should specify which one we want to optimize. The two DbContexts being the AdminUiDbContext and the DevicesDbContext. The reason for this is that we use the DevicesDbContext for OpenIddict in the adminUi project.

Seeing as the adminUi project does not use the other DbContexts, I'm not sure if we could even run the optimize command for them in the adminUi project, and if it would even make a diference in the performance.

As for the naming, we are only compiling the DbContexts used by the adminUi project so maybe we can change the function name to 'CompileModelsUsedByAdminUi' and the log message to 'Compiling models used by $adminUiProject...'.

tnotheis: The problem isn't just with the Admin UI. It's similar for the Consumer API - although less extreme there. But the first request takes about 10-20 times as long as subsequent requests.

So having compiled models for everything seems like it was worth a try.

scripts/windows/efcore/add_migration.ps1
```diff
@@ -1,27 +1,64 @@
...
         $provider
     )
 
-    New-Item env:"Modules__${moduleName}__Infrastructure__SqlDatabase__Provider" -Value $provider -Force | Out-Null
+    switch($moduleName){
+        "AdminUi" {
+            New-Item env:"${moduleName}__Infrastructure__SqlDatabase__Provider" -Value $provider -Force | Out-Null
 
-    $migrationProject = "$repoRoot\Modules\$moduleName\src\$moduleName.Infrastructure.Database.$provider"
+            $migrationProject = "$repoRoot\$moduleName\src\$moduleName.Infrastructure.Database.$provider"
```
tnotheis, line 22: I would hardcode the Admin UI folder names here:

```suggestion
            $migrationProject = "$repoRoot\AdminUi\src\AdminUi.Infrastructure.Database.$provider"
```

scripts/windows/efcore/remove_migration.ps1
tnotheis, line 1: We need to recreate the compiled models after removing a migration, don't we? 
But before you copy&paste the logic for the compiled model creation, you should probably extract the corresponding logic from the `CompileModels` function of the `add_migration.ps1` script into a separate script, which you then call from `add_migration.ps1` and `remove_migration.ps1`.

tnotheis (changes requested):
I didn't mean to approve

tnotheis (approved):

PR #300, opened 2023-09-18
Error on concurrent registrations to Push Notifications

daniel-almeida-konkconsulting:
…rrent device registration creation

# Readiness checklist

-   [X] I labeled the PR.


Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/PnsRegistrationRepository.cs
```diff
@@ -20,8 +23,16 @@ public PnsRegistrationRepository(DevicesDbContext dbContext)
...
-        await _registrations.AddAsync(registration, cancellationToken);
-        await _dbContext.SaveChangesAsync(cancellationToken);
+        try
+        {
+            await _registrations.AddAsync(registration, cancellationToken);
+            await _dbContext.SaveChangesAsync(cancellationToken);
+        }
+        catch (DbUpdateException exception)
+        {
+            if (exception.GetBaseException() is SqlException { Number: 2627 }) // violation in unique key index
```
tnotheis, line 33: Is it the same error code for SQL Server and Postgres?

Modules/Devices/src/Devices.Infrastructure/InfrastructureErrors.cs
```diff
@@ -10,4 +10,10 @@ public static InfrastructureError InvalidPushNotificationConfiguration(List<stri
         return new InfrastructureError("error.platform.unsupportedAppId",
             $"The given app id is not supported. Supported app ids are: {supportedAppIdsString}.");
     }
+
+    public static InfrastructureError ConcurrentDeviceRegistrationCreationViolatesPrimaryKeyConstraint(string deviceId)
```
tnotheis, line 14: I would rather go with `UniqueKeyViolation` to make it reusable. And adapt the error message accordingly.

Modules/Devices/src/Devices.Infrastructure/PushNotifications/DirectPush/DirectPushService.cs
```diff
@@ -56,9 +57,15 @@ public async Task UpdateRegistration(IdentityAddress address, DeviceId deviceId,
...
 
-            await _pnsRegistrationRepository.Add(new PnsRegistration(address, deviceId, handle, appId), cancellationToken);
-
-            _logger.LogTrace("New device successfully registered.");
+            try
+            {
+                await _pnsRegistrationRepository.Add(new PnsRegistration(address, deviceId, handle, appId), cancellationToken);
+                _logger.LogTrace("New device successfully registered.");
+            }
+            catch (InfrastructureException exception)
```
tnotheis, line 65: What if at some point we decide to throw an InfrastructureException for another reason? Then we would just log it here instead of passing it to the caller. So you should check for the specific error.
And can you please override the `Equals` method on the `InfrastructureError` and make it return true if the `Code` property is the same? (remember you can use ReSharper to generate the Equality members)

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/PnsRegistrationRepository.cs
tnotheis, line 33: We have `DbUpdateExtensions` with a `HasReason` method in `Synchronization.Application`. You should move them to `BuildingBlocks.Application`, add a reason for this exception, update the implementation of `HasReason`, and use the extension method here.

tnotheis (changes requested):

Modules/Devices/src/Devices.Infrastructure/InfrastructureErrors.cs
```diff
@@ -11,9 +11,9 @@ public static InfrastructureError InvalidPushNotificationConfiguration(List<stri
             $"The given app id is not supported. Supported app ids are: {supportedAppIdsString}.");
     }
 
-    public static InfrastructureError ConcurrentDeviceRegistrationCreationViolatesPrimaryKeyConstraint(string deviceId)
+    public static InfrastructureError UniqueKeyViolation(string key = "")
     {
-        return new InfrastructureError("error.platform.concurrentDeviceRegistrationCreation",
-            $"Device registration with id {deviceId} already created.");
+        return new InfrastructureError("error.platform.uniqueKeyViolation",
+            $"Entity with key {key} already exists");
```
tnotheis, line 17: In the Rocket Chat message I sent you, I had single quotes around the placeholder...
```suggestion
            $"Entity with key '{key}' already exists");
```

tnotheis (changes requested):

tnotheis (approved):

PR #301, opened 2023-09-19
Admin UI Code Quality Checks

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.

This PR changed most files in the AdminUi angular project, but most changes are similar.
The most unique changes were on these files:

`runChecks.mjs`
- Added the zx-script for the quality checks.
- The tests can be run from the ClientApp folder with the following command `zx .\.ci\runChecks.mjs`

`logger-server.service.ts`
- Added eslint naming exceptions for "X-API-KEY" and "X-XSFR-TOKEN".
- Added missing "X-XSFR-TOKEN" for the logger service (remember the logger bypasses interceptors).

`logger-writer.service.ts`
- Added eslint console exceptions for the console logging logic for the logger.

`api-key.interceptor.ts` `xsfr.interceptor.ts`
- Added eslint naming exceptions for "X-API-KEY" and "X-XSFR-TOKEN".

Besides these, most changes were simply adding return types, making variables public or private and awaiting promises.


AdminUi/src/AdminUi/ClientApp/.ci/runChecks.mjs
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env zx
+
+process.env.FORCE_COLOR = 3;
```
tnotheis, line 3: What's that?

AdminUi/src/AdminUi/ClientApp/.ci/runChecks.mjs
tnotheis, line 1: You can create an sh Script instead of mjs. I'm starting to get annoyed by `zx`.

AdminUi/src/AdminUi/ClientApp/.eslintrc
```diff
@@ -0,0 +1,15 @@
+{
+    "ignorePatterns": ["/*", "!/src", "!/test"],
```
tnotheis, line 2: Since we don't have a `test` folder and will probably never have, you can remove it from the list of ignorePatterns

AdminUi/src/AdminUi/ClientApp/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+    "extends": ["@js-soft/ts", "@js-soft/ts/mocha"],
+    "parserOptions": {
+        "project": ["tsconfig.json", "test/tsconfig.json"]
```
tnotheis, line 5: You can remove "test/tsconfig.json" from this list

AdminUi/src/AdminUi/ClientApp/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+    },
+    "rules": {
+        "jest/expect-expect": [
+            "error",
+            {
+                "assertFunctionNames": ["expect", "verify", "expectValid*", "*.expectThrows*"]
+            }
+        ]
+    }
```
tnotheis, line 14: You can remove this.

AdminUi/src/AdminUi/ClientApp/src/main.ts
```diff
@@ -4,4 +4,4 @@ import { AppModule } from "./app/app.module";
 
 platformBrowserDynamic()
     .bootstrapModule(AppModule)
-    .catch((err) => console.error(err));
+    .catch((err) => console.error(err)); // eslint-disable-line no-console
```
tnotheis, line 7: Didn't we (you) introduce a logging library lately? Why don't you simply use that one instead of `console`? Same for all other occurrences of console.*

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/.ci/runChecks.mjs
JLSRKonk: zx output is all in the same color. This line preserves the error and information colors in the command line.

AdminUi/src/AdminUi/ClientApp/src/main.ts
JLSRKonk: This is an extension of that library. If you don't remember, we needed to include it in the logging library PR in order to make the local console logs have the same appearance as the endpoint logs.

AdminUi/src/AdminUi/ClientApp/.ci/runChecks.sh
```diff
@@ -0,0 +1,9 @@
+set -e
+set -x
+
+npm ci
+npx eslint --ext ts ./src
+npx prettier --check .
+npx license-check --ignorePackages adminui@0.0.0
```
tnotheis, line 7: You should be able to skip the version

```suggestion
npx license-check --ignorePackages adminui
```

tnotheis (changes requested):

tnotheis:
Oh, and you should run the script from the test pipeline.

AdminUi/src/AdminUi/ClientApp/.ci/runChecks.sh
JLSRKonk: I tried it without the version, but it isn't ignoring the package without the version.

AdminUi/src/AdminUi/ClientApp/src/main.ts
tnotheis: Hmm, I wasn't aware of that. Isn't it possible to actually use the logger of the library?

tnotheis: Ahh, okay, then I'm fine with leaving it as it is.

.github/workflows/test.yml
```diff
@@ -8,6 +8,17 @@ on:
       - "**"
 
 jobs:
+  run-adminUi-checks:
+    runs-on: ubuntu-latest
+    name: AdminUi Run Checks
```
tnotheis, line 13:
```suggestion
    name: Run Admin UI Checks
```

.github/workflows/test.yml
```diff
@@ -8,6 +8,17 @@ on:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install script dependencies
+        run: npm install --prefix ./.ci
```
tnotheis, line 18: Since you don't use zx, you don't need this

.github/workflows/test.yml
```diff
@@ -8,6 +8,17 @@ on:
...
+      - name: Run Checks
+        run: bash AdminUi/src/AdminUi/ClientApp/.ci/runChecks.sh
```
tnotheis, line 20: I would probably move the script to the .ci folder, call it runAdminUiChecks. Then, in the first line of the script, save the current working directory (`$INITIAL_DIR=$pwd`), switch to the directory you need to be in, run the commands, and then switch back to the `INITIAL_DIR`.

tnotheis (changes requested):

.ci/runAdminUiChecks.sh
```diff
@@ -0,0 +1,11 @@
+set -e
+set -x
+
+INITIAL_DIR=$pwd
```
tnotheis, line 4: I forgot the parantheses:

```suggestion
INITIAL_DIR=$(pwd)
```

.github/workflows/test.yml
tnotheis: You only fixed this partially.

tnotheis (changes requested):
.

PR #307, opened 2023-09-26
Performance logging for calls to external services

NikolaVetnic:
# Readiness checklist

-   [X] I labeled the PR.

# Description

## 1 Blob Storage

Added `Microsoft.Extensions.Logging.Abstractions` and `System.Linq.Async` dependencies to `BuildingBlocks/src/Tooling`. 

Wrapped all the "atomic" awaited calls in appropriate `_logger` statements, e.g.:
```csharp
            await _logger.TraceTime(async () =>
            {
                await _storageClient.DownloadObjectAsync(folder, blobId, stream);
            }, nameof(FindAsync));
```

The following is an example of a "non-atomic" awaited call, which consists of several awaited calls which are all logged separately:
```csharp
            await UploadChangedBlobs();
```

Created custom method `ILoggerExtensions.TraceTimeAsyncEnumeration` to log the time needed to complete the `FindAllAsync` methods in `AzureStorageAccount` and `GoogleCloudStorage`. This method captures the items from the source, logs the elapsed time once the enumeration completes, and then returns the captured items as a new `IAsyncEnumerable<T>`. This solution might be memory inefficient, and I cannot determine if that impacts its usability in our case. I could also not figure out how to trigger this by shooting endpoints so it should be thoroughly checked.

Sample output of one of the time tracing `_logger` calls:
```JSON
{
   "@t":"2023-09-26T15:34:14.2190956Z",
   "@mt":"Executed action method \\{ActionName}, returned result \\{ActionResult} in \\{ElapsedMilliseconds}ms.",
   "ActionName":"Backbone.Modules.Challenges.ConsumerApi.Controllers.ChallengesController.Create (Backbone.Modules.Challenges.ConsumerApi)",
   "ActionResult":"Microsoft.AspNetCore.Mvc.CreatedAtActionResult",
   "ElapsedMilliseconds":7859.3592,
   "EventId":{
      "Id":103,
      "Name":"ActionMethodExecuted"
   },
   "SourceContext":"Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker",
   "ActionId":"7d1d0a78-a964-4434-95cb-eb9fbdd5a4bc",
   "RequestId":"0HMTUKNT91LTF:00000002",
   "RequestPath":"/api/v1/Challenges",
   "ConnectionId":"0HMTUKNT91LTF",
   "identityAddress":"",
   "deviceId":"",
   "CorrelationId":"79d10112-ff93-44ff-bb2a-89da78cbbecf"
}
```

## 2 `EventBus` Method Calls

Wrapped all the "atomic" awaited calls in appropriate `_logger` statements here as well.

## 3 Tracing Time for Executing DB Commands

As searching both docs and the GitHub repo yielded no results, I tried the third approach stated. Appropriate statement in `Backbone.Modules.Synchronization.Application.Datawallets.Commands.PushDatawalletModifications.Handler` was wrapped thusly:
```csharp
            await _logger.TraceTime(async () =>
            {
                await _dbContext.SaveChangesAsync(_cancellationToken);
                Console.WriteLine("OBSERVED_METHOD:: " + nameof(_dbContext.SaveChangesAsync));
            }, nameof(_dbContext.SaveChangesAsync));
```

The output resulting from hitting the `/Datawallet/Modifications` URL with a `GET` request is attached: [ConsumerApi_WalletModificationLogOutputSample.txt](https://github.com/nmshd/backbone/files/12730412/ConsumerApi_WalletModificationLogOutputSample.txt)

Search for `PLACE_OF_INTEREST` tags in the file for quick navigation to places of interest. 

In short, the time tracked by the `Executed DbCommand` statements in this case was 20ms, while our custom time tracer records ~2720ms. It appears to me that the `Executed DbCommand` does not take the latency into account.


BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/RabbitMQ/EventBusRabbitMQ.cs
```diff
@@ -181,7 +181,9 @@ private async Task ProcessEvent(string eventName, string message)
 
         if (_subsManager.HasSubscriptionsForEvent(eventName))
         {
-            await using var scope = _autofac.BeginLifetimeScope(AUTOFAC_SCOPE_NAME);
+            var scope = await _logger.TraceTime(
```
tnotheis, line 184: You can remove this. It doesn't take much time.

BuildingBlocks/src/Tooling/Extensions/ILoggerExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+            return await action();
+        }
+        finally
+        {
+            watch.Stop();
+            logger.LogDebug(EVENT_ID_EXECUTION_TIME, "'{action}' took {elapsedMilliseconds} ms.", actionName ?? "Action", watch.ElapsedMilliseconds);
+        }
+    }
+
+    public static async Task<IAsyncEnumerable<T>> TraceTimeAsyncEnumeration<T>(this ILogger logger, IAsyncEnumerable<T> source, string? actionName = null)
```
tnotheis, line 49: This method defeats the whole purpose of the AsyncEnumerable, because it fetches all items before returning them. So the caller will receive all items at once.
Before following such an approach, we should rather not log the time in this case at all.

Modules/Synchronization/src/Synchronization.Application/Datawallets/Commands/PushDatawalletModifications/Handler.cs
```diff
@@ -138,7 +142,11 @@ private async Task Save(DatawalletModification[] modifications)
 
         try
         {
-            await _dbContext.SaveChangesAsync(_cancellationToken);
+            await _logger.TraceTime(async () =>
+            {
+                await _dbContext.SaveChangesAsync(_cancellationToken);
+                Console.WriteLine("OBSERVED_METHOD:: " + nameof(_dbContext.SaveChangesAsync));
+            }, nameof(_dbContext.SaveChangesAsync));
```
tnotheis, line 149: Did you forget to remove this code?

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorage.cs
```diff
@@ -102,13 +107,17 @@ private async Task UploadChangedBlobs()
...
 
             try
             {
                 _logger.LogTrace("Uploading blob with key '{blobName}'...", blob.Name);
-                await _storageClient.UploadObjectAsync(blob.Folder, blob.Name, null,
-                    memoryStream);
+
+                await _logger.TraceTime(async () =>
+                {
+                    await _storageClient.UploadObjectAsync(blob.Folder, blob.Name, null, memoryStream);
```
tnotheis, line 118: Can you also log the time for the whole `UploadChangedBlobs` method please? (in `SaveAsync`, wrap the call of the method)

Modules/Synchronization/src/Synchronization.Application/Datawallets/Commands/PushDatawalletModifications/Handler.cs
NikolaVetnic: Yes, removed it just now.

BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/RabbitMQ/EventBusRabbitMQ.cs
NikolaVetnic: Removed the unnecessary `TraceTime()` statement.

BuildingBlocks/src/Tooling/Extensions/ILoggerExtensions.cs
NikolaVetnic: Removed `TraceTimeAsyncEnumeration()` method and its calls.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorage.cs
NikolaVetnic: Time logged for the whole `UploadChangedBlobs()` method.

NikolaVetnic:
An attempt was made to log the entire time elapsed while executing the `SaveChangesAsync()` method utilizing DB interceptors and `SavingChangesAsync` and `SavedChangesAsync` hooks.

DB interceptor class used (path `\Modules\Synchronization\src\Synchronization.Application\Infrastructure\SaveChangesInterceptor.cs`):
```csharp
using Microsoft.EntityFrameworkCore.Diagnostics;
using Microsoft.Extensions.Logging;

public class SaveChangesTimeInterceptor : SaveChangesInterceptor
{
    private readonly ILogger<SaveChangesTimeInterceptor> _logger;
    private System.Diagnostics.Stopwatch _stopwatch;

    public SaveChangesTimeInterceptor(ILogger<SaveChangesTimeInterceptor> logger)
    {
        _logger = logger;
    }

    public override ValueTask<InterceptionResult<int>> SavingChangesAsync(
        DbContextEventData eventData,
        InterceptionResult<int> result,
        CancellationToken cancellationToken = default)
    {
        _stopwatch = System.Diagnostics.Stopwatch.StartNew();
        return base.SavingChangesAsync(eventData, result, cancellationToken);
    }

    public override async ValueTask<int> SavedChangesAsync(
        SaveChangesCompletedEventData eventData,
        int result,
        CancellationToken cancellationToken = default)
    {
        _stopwatch.Stop();
        _logger.LogInformation($"SaveChangesAsync took {_stopwatch.ElapsedMilliseconds} ms.");
        return await base.SavedChangesAsync(eventData, result, cancellationToken);
    }
}
```

Registering the interceptor in `\Modules\Synchronization\src\Synchronization.Infrastructure\Persistence\Database\IServiceCollectionExtensions.cs`:
```csharp
    .AddInterceptors(new SaveChangesTimeInterceptor(
        services.BuildServiceProvider().GetRequiredService<ILogger<SaveChangesTimeInterceptor>>()    
    ))
```

Official docs on using interceptors in a scenario such as this one: [https://learn.microsoft.com/en-us/ef/core/logging-events-diagnostics/interceptors](https://learn.microsoft.com/en-us/ef/core/logging-events-diagnostics/interceptors)

Elapsed time measured varies, the greatest value being when the ConsumerApi is first started after which it decreases with every request. It may increase again after a period of inactivity. Example of the times measured in one case (each request was POST to `/Datawallet/Modifications` with 20 modifications): 195ms, 27ms, 20ms, 20ms, pause ~2min, 32ms, 18ms, pause ~15min, 18ms, 21ms, 21ms. The time logged by the `Executed DbCommand` statement is much lower, i.e. usually under 10ms.

All this info is given in order to scrutinize the interceptor itself as it's not yet 100% proven it does what it should do.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/AzureStorageAccount/AzureStorageAccount.cs
```diff
@@ -78,7 +83,7 @@ public Task<IAsyncEnumerable<string>> FindAllAsync(string folder, string? prefix
 
     public async Task SaveAsync()
     {
-        await UploadChangedBlobs();
+        await _logger.TraceTime(async () => UploadChangedBlobs(), nameof(UploadChangedBlobs));
         await DeleteRemovedBlobs();
```
tnotheis, line 87: For the sake of completeness, can you please add time logging for this call as well?

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/AzureStorageAccount/AzureStorageAccount.cs
```diff
@@ -95,10 +100,15 @@ private async Task UploadChangedBlobs()
         var changedBlobs = new Dictionary<BlobClient, byte[]>(_changedBlobs);
         foreach (var (cloudBlockBlob, bytes) in changedBlobs)
         {
-            await using var memoryStream = new MemoryStream(bytes);
+            var memoryStream = await _logger.TraceTime(async () => new MemoryStream(bytes), nameof(MemoryStream));
```
tnotheis, line 103: No need to log this call. It shouldn't even take 1ms.

Also, you removed the `using` as part of your changes, which isn't good.

Same for the other blob storage implementations.

BuildingBlocks/src/Tooling/Extensions/ILoggerExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Microsoft.Extensions.Logging;
+
+public static class ILoggerExtensions
+{
+    private const string DEBUG_PERFORMANCE_ENV_VAR = "DEBUG_PERFORMANCE";
+    private static readonly EventId EVENT_ID_EXECUTION_TIME = new(1000, "ExecutionTime");
+
+    public static async Task TraceTime(this ILogger logger, Func<Task> action, string? actionName = null)
+    {
+        if (Environment.GetEnvironmentVariable(DEBUG_PERFORMANCE_ENV_VAR).IsNullOrEmpty())
```
tnotheis, line 14: Can you extract this into a static class + variable please, so that we can use it in other places as well?

And remember to replace all usages of this with the new class.

```c#
public static class EnvironmentVariables
{
    public static readonly string DEBUG_PERFORMANCE = Environment.GetEnvironmentVariable("DEBUG_PERFORMANCE");
}

File: Modules/Synchronization/src/Synchronization.Infrastructure/Persistence/Database/IServiceCollectionExtensions.cs [hunk 8]
```diff
@@ -31,6 +32,9 @@ public static void AddDatabase(this IServiceCollection services, DbOptions optio
                         sqlOptions.MigrationsAssembly(SQLSERVER_MIGRATIONS_ASSEMBLY);
                         sqlOptions.EnableRetryOnFailure(options.RetryOptions.MaxRetryCount, TimeSpan.FromSeconds(options.RetryOptions.MaxRetryDelayInSeconds), null);
                     })
+                    .AddInterceptors(new SaveChangesTimeInterceptor(
+                        services.BuildServiceProvider().GetRequiredService<ILogger<SaveChangesTimeInterceptor>>()
+                    ))
```
tnotheis, line 37: Only add the interceptor if the env variable is set.

Modules/Synchronization/src/Synchronization.Infrastructure/Persistence/Database/IServiceCollectionExtensions.cs
```diff
@@ -41,7 +45,9 @@ public static void AddDatabase(this IServiceCollection services, DbOptions optio
                         sqlOptions.MigrationsAssembly(POSTGRES_MIGRATIONS_ASSEMBLY);
                         sqlOptions.EnableRetryOnFailure(options.RetryOptions.MaxRetryCount, TimeSpan.FromSeconds(options.RetryOptions.MaxRetryDelayInSeconds), null);
                     })
-
+                    .AddInterceptors(new SaveChangesTimeInterceptor(
+                        services.BuildServiceProvider().GetRequiredService<ILogger<SaveChangesTimeInterceptor>>()
+                    ))
```
tnotheis, line 50: Only add the interceptor if the env variable is set.

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/AzureStorageAccount/AzureStorageAccount.cs
NikolaVetnic: Hmm, that is already done on my end... I'll check again once I'm done with all the other changes.

tnotheis: And don't forget to do it for both implementations.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/Database/SaveChangesInterceptor.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Enmeshed.Tooling.Extensions;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.Extensions.Logging;
+
+namespace Enmeshed.BuildingBlocks.Infrastructure.Persistence.Database;
+
+public class SaveChangesTimeInterceptor : SaveChangesInterceptor
```
tnotheis, line 7: I wonder if this class is thread safe. So: is there a separate stopwatch for each save operation?

In other words: does EF Core create a new object of this class every time SaveChangesAsync is called?

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorage.cs
```diff
@@ -88,7 +93,7 @@ private void EliminateNotFound(Exception ex, string blobId)
 
     public async Task SaveAsync()
     {
-        await UploadChangedBlobs();
+        await _logger.TraceTime(async () => UploadChangedBlobs(), nameof(UploadChangedBlobs));
         await DeleteRemovedBlobs();
```
tnotheis, line 97: There is still no time logged for this call... Same for AzureStorageAccount

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/Database/SaveChangesInterceptor.cs
NikolaVetnic: It is actually thread-safe. I've implemented the static counter which is incremented in the constructor whenever an interceptor is instantiated. The value increases for every request made.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorage.cs
NikolaVetnic: I just checked my local files - it's definitely there, it also seems to be committed and pushed. Can you please check again?

tnotheis: https://github.com/nmshd/backbone/blob/c2289f60f52249335a75c405ac723cfba71176ea/BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorage.cs#L97

https://github.com/nmshd/backbone/blob/c2289f60f52249335a75c405ac723cfba71176ea/BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/AzureStorageAccount/AzureStorageAccount.cs#L87

This is the current state on your branch.

NikolaVetnic: Ok, I thought it refers to `UploadChangedBlobs()`. All done now.

tnotheis (dismissed):

tnotheis (approved):

PR #308, opened 2023-09-27
Enhance the log context with username

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description
Currently we log the identityAddress and the deviceId associated with a certain request for all log calls associated with that request. There is a special kind of requests, however. Requests to the `/connect/token` endpoint should somehow log what user did them, but the properties above are not set because there isn't a JWT yet. Thus, we open an exception for this scenario, logging the username directly from the request payload.

# Example
### Output:

```json
{
   "@t":"2023-09-27T08:35:07.6889720Z",
   "@mt":"The event {EventName} was marked as handled by {HandlerName}.",
   "@l":"Debug",
   "EventName":"OpenIddict.Server.OpenIddictServerEvents+ProcessSignInContext",
   "HandlerName":"OpenIddict.Server.OpenIddictServerHandlers+Exchange+ApplyTokenResponse`1[[OpenIddict.Server.OpenIddictServerEvents+ProcessSignInContext, OpenIddict.Server, Version=4.3.0.0, Culture=neutral, PublicKeyToken=****]]",
   "SourceContext":"OpenIddict.Server.OpenIddictServerDispatcher",
   "ActionId":"62feabce-cf4b-4eea-854d-e1106ff91ca1",
   "ActionName":"ConsumerApi.Controllers.AuthorizationController.Exchange (ConsumerApi)",
   "RequestId":"0HMTV6IFT6EFP:00000001",
   "RequestPath":"/connect/token",
   "ConnectionId":"0HMTV6IFT6EFP",
   "username":"USRa",
   "identityAddress":"",
   "deviceId":"",
   "CorrelationId":"f88153e7-2200-4d6f-9b5f-4eeb54aead13"
}
```

### DIFF
--- a/Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
+++ b/Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
@@ -12,6 +12,7 @@ public class AspNetCoreUserContext : IUserContext
     private const string ADDRESS_CLAIM = "address";
     private const string DEVICE_ID_CLAIM = "device_id";
     private const string USER_ID_CLAIM = "sub";
+    private const string USERNAME_CLAIM = "name";
     private const string SUBSCRIPTION_PLAN_CLAIM = "subscription_plan";
     private readonly IHttpContextAccessor _context;
 
@@ -66,7 +67,7 @@ public class AspNetCoreUserContext : IUserContext
 
     public string GetUsernameOrNull()
     {
-        var username = _context.HttpContext.User.Identities.First().Name;
+        var username = _context.HttpContext.User.FindFirstValue(USERNAME_CLAIM);
         return username;
     }
 
--- a/Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
+++ b/Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
@@ -59,6 +59,17 @@ public class AspNetCoreUserContext : IUserContext
         return userId;
     }
 
+    public string GetUsername()
+    {
+        return GetUsernameOrNull() ?? throw new NotFoundException();
+    }
+
+    public string GetUsernameOrNull()
+    {
+        var username = _context.HttpContext.User.Identities.First().Name;
+        return username;
+    }
+
     public IEnumerable<string> GetRoles()
     {
         var rolesClaim = _context.HttpContext.User.FindFirstValue(ClaimTypes.Role);
--- a/Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
+++ b/Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
@@ -12,7 +12,6 @@ public class AspNetCoreUserContext : IUserContext
     private const string ADDRESS_CLAIM = "address";
     private const string DEVICE_ID_CLAIM = "device_id";
     private const string USER_ID_CLAIM = "sub";
-    private const string USERNAME_CLAIM = "name";
     private const string SUBSCRIPTION_PLAN_CLAIM = "subscription_plan";
     private readonly IHttpContextAccessor _context;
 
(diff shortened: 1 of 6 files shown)


Dannyps:
@tnotheis I just noticed the case inconsistency between the `username`, `identityAddress` and `deviceId` attribute names and the rest of the attributes. Should these start with an uppercase letter?

ConsumerApi/Mvc/Middleware/UserDataLoggingMiddleware.cs
tnotheis, line 1: You know what: write the username into the logs no matter what. You can get it via the `IUserContext.GetUserIdOrNull` method. If this method returns null, read it via `context.GetOpenIddictServerRequest`

tnotheis:
> @tnotheis I just noticed the case inconsistency between the `username`, `identityAddress` and `deviceId` attribute names and the rest of the attributes. Should these start with an uppercase letter?

Valid point. But this would mean that we would have to update all log statements, because if there is something like

```csharp
_logger.LogDebug("{placeholder}", value);
```

then the JSON logs contain a property `placeholder`. I think that's not worth the effort. But don't get me wrong: it triggers me just as much as it does you. ;)

ConsumerApi/Mvc/Middleware/UserDataLoggingMiddleware.cs
Dannyps: I intercepted a call with a valid JWT token and printed the `_userContext.GetUserIdOrNull()` value: `5cda1a4b-08cd-4866-aa2a-d5641905dded`.

Somehow, I don't feel that what we expected. Can I proceed with fixing the `GetUserId(OrNull)` methods?

tnotheis: My bad. The UserId is actually the database id and not the username. You can add a GetUsername method to the `IUserContext` and in the `AspNetUserContext` read the `name` claim from the JWT. This should do the trick.

Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
```diff
@@ -59,6 +59,17 @@ public string GetUserIdOrNull()
...
     }
 
+    public string GetUsername()
+    {
+        return GetUsernameOrNull() ?? throw new NotFoundException();
+    }
+
+    public string GetUsernameOrNull()
+    {
+        var username = _context.HttpContext.User.Identities.First().Name;
```
tnotheis, line 69: 1. First() throws an exception if there is no element.
2. Why don't you use the same way to retrieve the value as in the other methods? (`_context.HttpContext.User.FindFirstValue(USERNAME_CLAIM)`)?

ConsumerApi/Mvc/Middleware/UserDataLoggingMiddleware.cs
```diff
@@ -24,7 +25,8 @@ public async Task Invoke(HttpContext context)
         ILogEventEnricher[] enrichers =
         {
             new PropertyEnricher("deviceId", deviceId ?? ""),
-            new PropertyEnricher("identityAddress", identityAddress ?? "")
+            new PropertyEnricher("identityAddress", identityAddress ?? ""),
+            new PropertyEnricher("username", _userContext.GetUsernameOrNull() ?? context.GetOpenIddictServerRequest()?.Username)
```
tnotheis, line 29: You forgot to handle null. Also, you should extract the username into a variable, just like deviceId and identityAddress
```suggestion
            var username = _userContext.GetUsernameOrNull() ?? context.GetOpenIddictServerRequest()?.Username;
            ...
            new PropertyEnricher("username", username ?? "")
```

tnotheis (changes requested):

Backbone.Infrastructure/UserContext/AspNetCoreUserContext.cs
Dannyps: 2. I considered it might be more interesting to use the typed attribute, but using the `name` claim works too.

tnotheis: Hm, yeah. If it's actually the correct value, why not

tnotheis (approved):

PR #312, opened 2023-10-02
Admin UI: Show number of Identities in Clients overview page

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.

# Screenshots
image


AdminUi/src/AdminUi.Infrastructure.Database.SqlServer/Migrations/20230929144733_ClientsOverview.cs
```diff
@@ -0,0 +1,33 @@
...
+                CREATE OR ALTER VIEW ClientOverviews AS
+                    SELECT
+                        CLIENTS.ClientId,
+                        CLIENTS.DisplayName,
+        	            CLIENTS.DefaultTier,
+                        COUNT (IDENTITIES.ClientId) AS NumberOfIdentities
+                    FROM Devices.OpenIddictApplications CLIENTS
+                    LEFT JOIN Devices.Identities IDENTITIES
+                    ON IDENTITIES.ClientId = CLIENTS.ClientId
+                    GROUP BY CLIENTS.ClientId, CLIENTS.DisplayName, CLIENTS.DefaultTier
```
tnotheis, line 23: Isn't the following easier/better?

```sql
SELECT
    CLIENTS.ClientId,
    CLIENTS.DisplayName,
    CLIENTS.DefaultTier,
    (
		SELECT COUNT(ClientId) 
		FROM Devices.Identities
		WHERE ClientId = CLIENTS.ClientId
	) AS NumberOfIdentities
FROM Devices.OpenIddictApplications CLIENTS
```
(honest question, because I'm not an SQL expert by any means. ;)

AdminUi/src/AdminUi.Infrastructure.Database.Postgres/Migrations/20230929151518_ClientsOverview.cs
```diff
@@ -0,0 +1,33 @@
...
+                CREATE OR REPLACE VIEW "ClientOverviews" AS
+                    SELECT
+                        CLIENTS."ClientId" AS "ClientId",
+                        CLIENTS."DisplayName" AS "DisplayName",
+                        CLIENTS."DefaultTier" AS "DefaultTier",
+                        COUNT (IDENTITIES."ClientId") AS "NumberOfIdentities"
+                    FROM "Devices"."OpenIddictApplications" CLIENTS
+                    LEFT JOIN "Devices"."Identities" IDENTITIES
+                    ON IDENTITIES."ClientId" = CLIENTS."ClientId"
+                    GROUP BY CLIENTS."ClientId", CLIENTS."DisplayName", CLIENTS."DefaultTier"
```
tnotheis, line 23: See comment on SQL Server.

AdminUi/src/AdminUi.Infrastructure.Database.Postgres/Migrations/20230929151518_ClientsOverview.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace AdminUi.Infrastructure.Database.Postgres.Migrations
+{
+    /// <inheritdoc />
+    public partial class ClientsOverview : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql("""
+                CREATE OR REPLACE VIEW "ClientOverviews" AS
```
tnotheis, line 14: I know it's similar in case of the IdentityOverview, but we should probably rather use `CREATE VIEW` instead of `CREATE OR REPLACE VIEW`. Just to make sure we don't replace anything by accident. Same for SQL Server.

```suggestion
                CREATE VIEW "ClientOverviews" AS
```

tnotheis (changes requested):

AdminUi/src/AdminUi.Infrastructure.Database.SqlServer/Migrations/20230929144733_ClientsOverview.cs
JLSRKonk: Yeah it seems much better, the one I had before was overly complicated. I tested it and it seems like there should be no issues.

tnotheis (approved):

PR #315, opened 2023-10-04
Save all datawallet modification payloads of a push in a single blob

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

The reason for this change is that we currently observe some performance issues in the customer's environment. This is presumably due to a relatively slow blob storage. With this change, instead of saving one blob per datawallet modification, there is only one blob, which contains the payloads of all modifications in a JSON-serialized Dictionary, where the Index of the the modification is the key of the dictionary.


Dannyps (commented):
Do we ensure retro-compatibility with old DataWalletModifications? ANd if not, should we?

Modules/Synchronization/src/Synchronization.Application/Datawallets/Queries/GetModifications/Handler.cs
```diff
@@ -39,27 +45,60 @@ public async Task<GetModificationsResponse> Handle(GetModificationsQuery request
...
+        var blobs = await Task.WhenAll(blobReferences.Select(r =>
+        {
+            try
+            {
+                return _blobStorage.FindAsync(_blobOptions.RootFolder, r);
+            }
+            catch (NotFoundException)
+            {
+                _logger.LogWarning("Blob with reference '{blobReference}' not found.", r);
+                throw new Exception($"Blob with reference '{r}' not found.");
```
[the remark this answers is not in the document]
tnotheis: Yeah, thought so too when I looked at it a few minutes ago. ;-)

tnotheis:
> Datawallets/Queries/GetModifications/Handler.cs

We do: https://github.com/nmshd/backbone/blob/2dc9c17232120266a37cbf33eb133a4256fd7e6c/Modules/Synchronization/src/Synchronization.Application/Datawallets/Queries/GetModifications/Handler.cs#L95-L99

Btw: the solution I implemented is far from ideal. Especially the fact that the logic for creating modifications is duplicated across two files. But since we need to refactor the persistence of the Synchronization module anyway, I didn't spend a lot of time finding a better solution.

Dannyps (approved):

Modules/Synchronization/src/Synchronization.Infrastructure.Database.Postgres/Migrations/20231004081756_AddBlobReferencePropertyToDatawalletModification.cs
```diff
@@ -0,0 +1,32 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "BlobReference",
+                table: "DatawalletModifications",
+                type: "character(32)",
+                unicode: false,
+                fixedLength: true,
+                maxLength: 32,
+                nullable: false,
```
daniel-almeida-konkconsulting, line 20: Shouldn't it be nullable?

Modules/Synchronization/src/Synchronization.Domain/Entities/DatawalletModification.cs
```diff
@@ -6,10 +6,13 @@ namespace Backbone.Modules.Synchronization.Domain.Entities;
 public class DatawalletModification
 {
 #pragma warning disable CS8618
-    private DatawalletModification() { }
+    private DatawalletModification(string blobReference)
```
daniel-almeida-konkconsulting, line 9: Why the new parameter in the private construtor when it's only used by EF Core to generate the migrations?

Modules/Synchronization/src/Synchronization.Application/Datawallets/Commands/PushDatawalletModifications/Handler.cs
```diff
@@ -109,7 +112,7 @@ private DatawalletModification CreateModification(PushDatawalletModificationItem
             modificationDto.ObjectIdentifier,
             modificationDto.PayloadCategory,
             modificationDto.EncryptedPayload,
-            _activeDevice
+            _activeDevice, blobReference
```
daniel-almeida-konkconsulting, line 115: Nitpicking, but I would place the parameter in a new line

Modules/Synchronization/src/Synchronization.Application/Datawallets/Queries/GetModifications/Handler.cs
```diff
@@ -15,15 +19,17 @@ public class Handler : IRequestHandler<GetModificationsQuery, GetModificationsRe
...
     private readonly IMapper _mapper;
     private readonly BlobOptions _blobOptions;
 
-    public Handler(ISynchronizationDbContext dbContext, IMapper mapper, IUserContext userContext, IBlobStorage blobStorage, IOptions<BlobOptions> blobOptions)
+    public Handler(ISynchronizationDbContext dbContext, IMapper mapper, IUserContext userContext, IBlobStorage blobStorage, IOptions<BlobOptions> blobOptions, ILogger<Handler> logger)
     {
         _dbContext = dbContext;
         _mapper = mapper;
         _blobStorage = blobStorage;
+        _logger = logger;
```
daniel-almeida-konkconsulting, line 32: Should probably remove the _logger since we no longer use it after daniel silvas' suggestion

Modules/Synchronization/src/Synchronization.Infrastructure.Database.Postgres/Migrations/20231004081756_AddBlobReferencePropertyToDatawalletModification.cs
tnotheis: No. Because new modifications should always specify a value.

New modifications will receive the default value (=> an empty string)

Modules/Synchronization/src/Synchronization.Domain/Entities/DatawalletModification.cs
tnotheis: Oh, it seems ReSharper added it to all constructors and not just the "real" one

Modules/Synchronization/src/Synchronization.Infrastructure.Database.Postgres/Migrations/20231004081756_AddBlobReferencePropertyToDatawalletModification.cs
daniel-almeida-konkconsulting: Did you mean to write "Old modifications will receive the default value (=> an empty string)" in your comment?
If yes, then I understand (y)

tnotheis: Ahm, yeah, of course I did. ;-)

tnotheis:
Sorry for the mess in the PR. Wanted to fix everything before we deploy the new version. But doing something quickly is not always quick. ;)

daniel-almeida-konkconsulting (approved):

PR #318, opened 2023-10-04
Repositories which provide Count methods with time frames must use those parameters

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description
It has become clear that we need to ensure that the repositories used by the implementations of the IMetricCalulator interface make use of the `from` and `to` parameters. Unit tests are a valid approach to fill this need.
What has not become clear, as of yet, is if we should extensively test this, or if testing one of the repositories is enough. Moreover, the approach to this testing is highly unorthodox as we cannot create an in-memory database from the `QuotasDbContext`, having to do it using the original entity's DbContext, and then re-using the connection.

# Decide
We have to decide:
1. Whether the current approach is the best.
2. If we should replicate the same test for all other repositories (or maybe just a select few). And how.


tnotheis (changes requested):
I don't like that you return yet another value from the CreateDbContexts method. Instead, in cases where you need the same connection for multiple instances, you should create the sqlite connection outside of the method and pass it to the first call already.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+    {
+        AssertionScope.Current.FormattingOptions.MaxLines = 1000;
+
+        var connection = FakeDbContextFactory.CreateDbConnection();
+        (_messagesArrangeContext, _, _) = FakeDbContextFactory.CreateDbContexts<MessagesDbContext>(connection);
+        (_, _, _actContext) = FakeDbContextFactory.CreateDbContexts<QuotasDbContext>(connection);
+    }
+
+    [Fact]
+    public async void Counts_entities_within_timeframe_hour_quotaPeriod()
```
tnotheis, line 39: You should change the return type of the tests to `Task`.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+using Backbone.Modules.Quotas.Infrastructure.Persistence.Database;
+using Backbone.Modules.Quotas.Infrastructure.Persistence.Repository;
+using Enmeshed.DevelopmentKit.Identity.ValueObjects;
+using Enmeshed.Tooling;
+using Enmeshed.UnitTestTools.Data;
+using Enmeshed.UnitTestTools.TestDoubles.Fakes;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Xunit;
+using MessageEntity = Backbone.Modules.Messages.Domain.Entities.Message;
```
tnotheis, line 13: Why do you rename it?

Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
```diff
@@ -0,0 +1,146 @@
...
+        return new(
+            identityAddress,
+            DeviceId.New(),
+            null,
+            Array.Empty<byte>(),
+            new List<Attachment>(),
+            new List<RecipientInformation>()
+            )
+        {
+            CreatedAt = createdAt
```
tnotheis, line 143: You can get rid of the `InternalsVisibleTo` attribute and the setter of the `CreatedAt` property by changing the implementation of the `CreateMessage` method to something like this:

```csharp
SystemTime.Set(createdAt);

var message = new(
            identityAddress,
            DeviceId.New(),
            null,
            Array.Empty<byte>(),
            new List<Attachment>(),
            new List<RecipientInformation>()
            );

SystemTime.Reset();

return message;
```

Cases like this are the reason I introduced `SystemTime`.

tnotheis (changes requested):
All in all I'm quite happy with the tests. 👍

Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
Dannyps: It was necessary at some point, but not anymore.

Modules/Messages/src/Messages.Domain/Entities/Message.cs
```diff
@@ -25,7 +25,7 @@ public Message(IdentityAddress createdBy, DeviceId createdByDevice, DateTime? do
 
     public MessageId Id { get; }
 
-    public DateTime CreatedAt { get; internal set; }
+    public DateTime CreatedAt { get; set; }
```
tnotheis, line 28: By removing the `internal` you just made it worse. Remove the whole setter.

tnotheis (changes requested):

Modules/Messages/src/Messages.Domain/Entities/Message.cs
Dannyps: I was sleeping, for sure. My bad

tnotheis (approved):

PR #325, opened 2023-10-09
Integration events get lost when they are in the queue during startup

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description

When launching the ConsumerAPI or the AdminAPI, the current code starts consuming events before the event handlers are registered. This results in events being lost if they're sent before the handler registration, e.g. while the app is down. This PR fixes that by adding a way for both APIs to only start consuming events after having registered all handlers.

# Testing

Testing can be achieved by:
1. Stopping the ConsumerAPI
2. Creating a new Tier on the AdminUI (for example, any other triggering event action would be okay)
3. Launching the Consumer API
4. Check that the created Tier exists on the devices schema.

We have run tests for the checked implementations:
 - [x] Azure Event Bus
 - [x] RabbitMq
 - [x] Google PubSub

# Remarks
 - This PR includes a change to the AdminUi's `appsettings.override.json` file in order to make use of Postgres instead of MS SQL Server for local dev.


BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/AzureServiceBus/EventBusAzureServiceBus.cs
```diff
@@ -106,6 +104,11 @@ public async void Publish(IntegrationEvent @event)
         _subscriptionManager.AddSubscription<T, TH>();
     }
 
+    public void StartConsuming()
+    {
+        RegisterSubscriptionClientMessageHandlerAsync().GetAwaiter().GetResult();
```
tnotheis, line 109: I think we don't actually need to await this, do we? It just finishes at some time in the future and starts consuming.

AdminUi/src/AdminUi/Program.cs
```diff
@@ -155,6 +156,9 @@ static void Configure(WebApplication app)
     {
         ResponseWriter = HealthCheckWriter.WriteResponse
     });
+
+    var eventBus = app.Services.GetRequiredService<IEventBus>();
+    eventBus.StartConsuming();
```
tnotheis, line 161: Actually, we don't register for push notifications in the Admin UI. So maybe we should remove this call, in order to avoid an open, unused connection to the event bus.

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/RabbitMQ/EventBusRabbitMQ.cs
```diff
@@ -50,6 +51,14 @@ public void Dispose()
         _subsManager.Clear();
     }
 
+    public void StartConsuming()
+    {
+        if(_consumer != null)
```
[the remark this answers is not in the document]
Dannyps: We need to call `CreateConsumerChannel` before because its side-effects are needed to register event handlers.

Regarding that flaw, what do you think about throwing an Exception on the else?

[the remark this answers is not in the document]
tnotheis: > Regarding that flaw, what do you think about throwing an Exception on the else?

That's fine.

AdminUi/src/AdminUi/Program.cs
tnotheis: That's true... 🤔

Though I think that we will rather create a completely new service that sits next to Consumer API and Admin UI which is responsible for processing all integration events. Listening to events in the Admin UI would probably make things very complicated.

tnotheis: But: props for bringing this up :)

BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/AzureServiceBus/EventBusAzureServiceBus.cs
tnotheis: Don't we have the same warning in the following line: https://github.com/nmshd/backbone/blob/f5dfb9a75c4dcdc974db106b685ee36a6a3b4908/BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/GoogleCloudPubSub/EventBusGoogleCloudPubSub.cs#L84 then?

tnotheis (approved):

PR #329, opened 2023-10-10
Admin UI: Default Tier in Clients Overview page shows Id instead of Display Name

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.

# Screenshots
image


AdminUi/src/AdminUi.Infrastructure/DTOs/ClientOverview.cs
```diff
@@ -3,7 +3,8 @@ public class ClientOverview
 {
     public string ClientId { get; set; }
     public string DisplayName { get; set; }
-    public string DefaultTier { get; set; }
+    public string DefaultTierId { get; set; }
+    public string DefaultTierName { get; set; }
```
tnotheis, line 7:
```suggestion
    public TierDTO DefaultTier { get; set; }
```

Should be like that in all places, of course. I won't comment all of them.

tnotheis (changes requested):

AdminUi/src/AdminUi.Infrastructure/DTOs/ClientOverview.cs
JLSRKonk: Are these changes only for the ClientOverview view?
The IdentityOverview view also has this format:
image

Should I change it there as well?

AdminUi/src/AdminUi.Infrastructure/Persistence/Database/EntityTypeConfigurations/ClientOverviewEntityTypeConfiguration.cs
```diff
@@ -8,6 +8,13 @@ public class ClientOverviewEntityTypeConfiguration : IEntityTypeConfiguration<Cl
...
         builder.ToView("ClientOverviews");
-        builder.HasNoKey();
+        builder.OwnsOne(
+            c => c.DefaultTier,
+            dt =>
+            {
+                dt.Property(p => p.Id).HasColumnName("DefaultTierId");
+                dt.Property(p => p.Name).HasColumnName("DefaultTierName");
+            });
+        builder.HasKey(c => c.ClientId);
```
tnotheis, line 18: I don't know whether it is a good idea to define a key here.
1. Afaik a view can't have a PK.
2. The ClientId isn't part of the PK in the Clients table. Maybe this will cause problems?

Did you run your code on SQL Server as well as Postgres?

AdminUi/src/AdminUi.Infrastructure/Persistence/Database/EntityTypeConfigurations/IdentityOverviewEntityTypeConfiguration.cs
```diff
@@ -9,7 +9,14 @@ public class IdentityOverviewEntityTypeConfiguration : IEntityTypeConfiguration<
...
         builder.ToView("IdentityOverviews");
-        builder.HasNoKey();
+        builder.OwnsOne(
+            c => c.Tier,
+            dt =>
+            {
+                dt.Property(p => p.Id).HasColumnName("TierId");
+                dt.Property(p => p.Name).HasColumnName("TierName");
+            });
+        builder.HasKey(c => c.Address);
```
tnotheis, line 19: I don't know whether it is a good idea to define a key here. Afaik a view can't have a PK.

Did you run your code on SQL Server as well as Postgres?

tnotheis (changes requested):

AdminUi/src/AdminUi.Infrastructure/Persistence/Database/EntityTypeConfigurations/ClientOverviewEntityTypeConfiguration.cs
JLSRKonk: I ran this both on SQL Server and Postgres and everything seemed to be working properly, I didn't find any issues.

Regarding the first two questions, when I first implemented this `.OwnsOne()`, I had a `.HasNoKey()` but was running into this error when creating migrations:
MicrosoftTeams-image
After troubleshooting with Daniel Almeida, we tried adding the clientId as key and it worked.

Looking at the migration code, at first glance, I don't think there should be any problems in the future, but I do agree that it's weird having to assign a key to a view.

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-list/client-list.component.ts
```diff
@@ -97,7 +97,7 @@ export class ClientListComponent {
         this.clients = this.serverClients;
         if (this.filter.clientId) this.clients = this.clients.filter((c) => c.clientId.toUpperCase().includes(this.filter.clientId!.toUpperCase()));
         if (this.filter.displayName) this.clients = this.clients.filter((c) => c.displayName.toUpperCase().includes(this.filter.displayName!.toUpperCase()));
-        if (this.filter.tiers && this.filter.tiers.length > 0) this.clients = this.clients.filter((c) => this.filter.tiers!.includes(c.defaultTier));
+        if (this.filter.tiers && this.filter.tiers.length > 0) this.clients = this.clients.filter((c) => this.filter.tiers!.includes(c.defaultTier.id));
```
tnotheis, line 100: Am I getting this right: the user has to type the ID of the client? That's not good. How should he know the ID? He only sees the name in the UI.

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-list/client-list.component.ts
JLSRKonk: As far as I am aware, the Tiers filter currently uses a dropdown without search:
image
So the user doesn't interact with the IDs directly, only with the Tier names.

tnotheis (approved):

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-list/client-list.component.ts
tnotheis: Ah, sure.

PR #348, opened 2023-10-18
Simplify aspnet logging

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

This PR includes the following:
* It mainly reduces the noise of ASP.NET Core default logs by reducing the log level to Warning and at the same time adding the `SerilogRequestLogging` middleware, which summarizes each request.
* I also added "Seq" for local development. Seq is a tool to view structured logs. I added it as a container to the docker-compose stack, so you can now access the logs on `locahost:8081`. 
* I also increased the default log level from Error to Warning, while overriding the default log level for our backbone internal logs to Information
* Finally, I surrounded the startup logic with a try catch block, and added some log statements to `catch` and `finally`, as this seems to be best practice.

### DIFF
--- a/docker-compose/docker-compose.yml
+++ b/docker-compose/docker-compose.yml
@@ -89,12 +89,23 @@ services:
     volumes:
       - rabbitmq-volume:/var/lib/rabbitmq # default data dir
 
+  seq:
+    image: datalust/seq:latest
+    ports:
+      - "8081:80"
+      - "5341:5341"
+    volumes:
+      - seq-data:/data
+    environment:
+      - ACCEPT_EULA=Y
+    restart: unless-stopped
+
   #elasticsearch:
   #  image: elasticsearch:${ELK_VERSION}
   #  container_name: elasticsearch
   #  restart: always
   #  volumes:
-  #    - elastic_data:/usr/share/elasticsearch/data/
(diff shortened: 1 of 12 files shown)


docker-compose/docker-compose.yml
```diff
@@ -89,12 +89,23 @@ services:
     volumes:
       - rabbitmq-volume:/var/lib/rabbitmq # default data dir
 
+  seq:
```
jkoenig134, line 92: nothing depends on this

jkoenig134 (approved):

docker-compose/docker-compose.yml
tnotheis: I think that's okay. If the server is not available yet, I think our logging library waits until it is. And if it doesn't it's okay as well, as the startup logs are not that important.
Waiting would increase the startup time though, which would be worse.

tnotheis: Though... Now that I think about it: it would be nice if the seq service would be started automatically when you start the consumer api 🤔

tnotheis: But as I don't have time now to wait for the pipelines to run again, I will do it as part of my next PR.

jkoenig134 (approved):

PR #350, opened 2023-10-18
AdminUI: Filtering and ordering for overview tables

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


tnotheis:
Where are the screenshots?

AdminUi/src/AdminUi/ClientApp/src/app/services/identity-service/identity.service.ts
```diff
@@ -11,23 +15,172 @@ import { Quota } from "../quotas-service/quotas.service";
...
+            }
+        }
+
+        let filterParameter = "";
+        if (odataFilter.toString() !== "") filterParameter = `?$filter=${odataFilter.toString()}`;
+        if (filterParameter === "") {
+            filterParameter += `?${paginationFilter}`;
+        } else {
+            filterParameter += `&${paginationFilter}`;
+        }
```
tnotheis, line 179: Wouldn't this work:

```typescript
const filterComponents = [odataFilter.toString(), paginationFilter];
const filterParameter = `?${filterComponents.join("&")}`;
```
Since the `paginationFilter` seems to always be set, this should work, and would be way easier to understand imo.

AdminUi/src/AdminUi/ClientApp/src/app/utils/number-filter.ts
```diff
@@ -0,0 +1,4 @@
+export interface NumberFilter {
+    operator?: string;
+    value?: number | null;
```
tnotheis, line 3: Why is it optional AND can be null?

AdminUi/src/AdminUi/ClientApp/src/app/utils/odata-response.ts
```diff
@@ -0,0 +1,3 @@
+export interface ODataResponse<Type> {
+    value: Type;
```
tnotheis, line 2: Should be same as in the "normal" API.
```suggestion
    result: Type;
```

AdminUi/test/AdminUi.Tests.Integration/Models/ODataResponseContent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Newtonsoft.Json;
+
+namespace AdminUi.Tests.Integration.Models;
+
+public class ODataResponseContent<T>
+{
+    public T? Value { get; set; }
+
+    [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+    public Error? Error { get; set; }
```
tnotheis, line 10: Why does the response have an `Error` property here, but not in the UI?

AdminUi/test/AdminUi.Tests.Integration/Models/RequestConfiguration.cs
```diff
@@ -8,6 +8,7 @@ public class RequestConfiguration
     public string? ContentType { get; set; }
     public string? AcceptHeader { get; set; }
     public string? Content { get; set; }
+    public bool IsODataRequest { get; set; }
```
tnotheis, line 11: Why is this property necessary?

tnotheis: It isn't even used, as far as I can see.

tnotheis (changes requested):

daniel-almeida-konkconsulting:
image

daniel-almeida-konkconsulting:
image

AdminUi/src/AdminUi/ClientApp/src/app/components/client/client-list/client-list.component.ts
```diff
@@ -16,46 +20,64 @@ import { ChangeSecretDialogComponent } from "../change-secret-dialog/change-secr
...
+    public operators: string[] = ["=", "<", ">", ">=", "<="];
     public displayedColumns: string[] = ["select", "clientId", "displayName", "defaultTier", "numberOfIdentities", "createdAt", "actions"];
 
     public constructor(
         private readonly router: Router,
         private readonly dialog: MatDialog,
         private readonly snackBar: MatSnackBar,
-        private readonly clientService: ClientServiceService
+        private readonly tierService: TierService,
+        private readonly clientService: ClientServiceService,
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: I have no clue, the client service as been created for quite some time now, I'm simply using it.
I'll rename it.

AdminUi/src/AdminUi/ClientApp/src/app/utils/number-filter.ts
daniel-almeida-konkconsulting: Field is optional due to the usage of ngIf directives (meaning it's undefined when the filter is not rendered).
Can be null because <input> fields set the value to null when their content is deleted.

tnotheis: > Can be null because fields set the value to null when their content is deleted.

Why don't they just set the value to undefined?

daniel-almeida-konkconsulting: I'm not sure, that was my initial thought, especially when considering I had the field set as optional only.
But the reality is, even when value was defined as `value?: number;`, it would be set as null by the input component.

AdminUi/src/AdminUi/ClientApp/src/app/services/identity-service/identity.service.ts
daniel-almeida-konkconsulting: that approach is missing the $filter anottation based on whether we are applying a filter or just the pagination.

tnotheis (approved):

PR #351, opened 2023-10-19
Remove Azure Notification Hub support

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

Since Azure Notification Hub does not support the full feature set we currently need (especially having multiple apps with different PNS keys) we remove the implementation.


tnotheis:
Ah, by the way, I also removed some other things we don't use anymore (namely AzureAppConfiguration and ApplicationInsights).

daniel-almeida-konkconsulting (approved):

PR #357, opened 2023-10-25
Admin UI: List Identities in the Client Details page

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.

# Screenshots
## Client Details
image
_note: Created with Client column and filter not shown in the Client Details page_

## List Identities
image
_note: Created with Client column and filter are shown in the List Identities page_


daniel-almeida-konkconsulting:
Can we center the table content with the filters?
I think it would look a lot cleaner in both pages.
A simples `text-align: center` should be sufficient.

daniel-almeida-konkconsulting:
From the screenshots, it appears the table (when in the identity overview page) is not occupying the page in its entirety.
Wasn't this the case before promoting it to a separate component?
I'd say we need as much screen real estate as we have available for this page.

JLSRKonk:
> Can we center the table content with the filters? I think it would look a lot cleaner in both pages. A simple `text-align: center` should be sufficient.

Here is a screenshot with the text centered:
image
@tnotheis Do you prefer the table content to be centered, or should we leave it as it was before?



> From the screenshots, it appears the table (when in the identity overview page) is not occupying the page in its entirety. Wasn't this the case before promoting it to a separate component? I'd say we need as much screen real estate as we have available for this page.

For the screenshots, I zoomed out the page to 50%, so maybe that is why the width may look a bit odd.
I tested it in the main branch and it looks the same, so the width should still be the same as before.
For reference, here is a normal 100% zoom screenshot:
image
There is a scrollbar below for the rest of the table.

tnotheis:
I kind of dislike both. ;) But since I don't have a better idea, let's go with the centered approach.

tnotheis (approved):

PR #359, opened 2023-10-26
Listing quotas for identity

NikolaVetnic:
# Readiness checklist

-   [X] I labeled the PR.

# Description

Listing all individual and tier quotas for the identity whose address is passed in as a `[FromRoute]` parameter. The quotas related to the identity whose address is passed in the request are mapped to `QuotaDTO` objects, paginated and then returned to the controller. Response caching is used to cache the results as the operation is quite expensive.


Modules/Quotas/src/Quotas.Application/DTOs/IndividualQuotaDTO.cs
```diff
@@ -16,4 +24,34 @@ public IndividualQuotaDTO(string id, MetricDTO metric, int max, QuotaPeriod peri
...
+    private readonly IMetricsRepository _metricRepository;
+
+    public MetricToMetricDtoResolver(IMetricsRepository metricRepository)
+    {
+        _metricRepository = metricRepository;
+    }
+
+    public MetricDTO Resolve(IndividualQuota source, IndividualQuotaDTO destination, MetricDTO destMember, ResolutionContext context)
+    {
+        var metric = _metricRepository.FindAllWithKeys(new List<MetricKey> { source.MetricKey }, CancellationToken.None);
```
tnotheis, line 54: `FindAllWithKeys` is an async method. Calling an async method in a non-async context is really bad in 99% of all cases, because it blocks the current thread. This can lead to really bad performance.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIndividualQuotas/Handler.cs
```diff
@@ -0,0 +1,55 @@
...
+using Google.Api;
+using MediatR;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace Backbone.Modules.Quotas.Application.Identities.Queries.GetIndividualQuotas;
+
+public class Handler : IRequestHandler<ListIndividualQuotasQuery, ListIndividualQuotasResponse>
+{
+    private readonly IMapper _mapper;
+    private readonly IMemoryCache _cache;
```
tnotheis, line 17: Why do you use a cache?

tnotheis (changes requested):
In general, if you didn't finish a task yet, you shouldn't create a PR. At least not a "real" one. You can however create a PR in status "Draft". Also, keep in mind that the PR description will be what people read when they want to know more about a feature. So having sentences like "This is a work in progress." are bad.

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
```diff
@@ -0,0 +1,40 @@
...
+    private readonly ApplicationOptions _options;
+
+    public QuotasController(IMediator mediator, IOptions<ApplicationOptions> options) : base(mediator)
+    {
+        _options = options.Value;
+    }
+
+    [HttpGet("{address}")]
+    [ResponseCache(Duration = 1800, Location = ResponseCacheLocation.Any, VaryByQueryKeys = new string[] { "address" })]
+    [ProducesResponseType(typeof(PagedHttpResponseEnvelope<List<QuotaDTO>>), StatusCodes.Status200OK)]
```
tnotheis, line 28: Are you aware of what you did here? Now every identity can get the quota of each other identity. 😲 I mean... it's no confidential data, but we still shouldn't do that.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/Handler.cs
```diff
@@ -0,0 +1,85 @@
...
+    public Handler(IMemoryCache cache, IIdentitiesRepository identitiesRepository, MetricCalculatorFactory metricCalculatorFactory)
+    {
+        _cache = cache;
+        _identitiesRepository = identitiesRepository;
+        _metricCalculatorFactory = metricCalculatorFactory;
+    }
+
+    public async Task<ListQuotasForIdentityResponse> Handle(ListQuotasForIdentityQuery request, CancellationToken cancellationToken)
+    {
+        var identityList = await _identitiesRepository.FindByAddresses(new List<string> { request.Address }.AsReadOnly(), cancellationToken, true);
```
tnotheis, line 29: 1. Why do you call `AsReadOnly`?
2. Do you know what the last parameter of the `FindByAddresses` method does?
3. Why do you call the method that searches multiple addresses instead of the one that returns just a single one?

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/Handler.cs
tnotheis, line 1: Where are the tests?

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/Handler.cs
```diff
@@ -0,0 +1,85 @@
...
+        var identity = identityList.First();
```
tnotheis, line 30: This method throws an exception if the list is empty. This results in an http status 500.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/ListQuotasForIdentityQuery.cs
```diff
@@ -0,0 +1,16 @@
...
+
+public class ListQuotasForIdentityQuery : IRequest<ListQuotasForIdentityResponse>
+{
+    public ListQuotasForIdentityQuery(PaginationFilter paginationFilter, string address)
+    {
+        PaginationFilter = paginationFilter;
+        Address = address;
+    }
+
+    public PaginationFilter PaginationFilter { get; set; }
```
tnotheis, line 14: Pagination is only necessary for requests where we expect there to be a lot of response items. Which is not the case for quotas. I guess there will be at most 10.

Modules/Quotas/src/Quotas.ConsumerApi/QuotasModule.cs
```diff
@@ -30,6 +30,10 @@ public override void ConfigureServices(IServiceCollection services, IConfigurati
         });
 
         services.AddSqlDatabaseHealthCheck(Name, parsedConfiguration.Infrastructure.SqlDatabase.Provider, parsedConfiguration.Infrastructure.SqlDatabase.ConnectionString);
+
+        services.AddMemoryCache();
```
tnotheis, line 34: Remove this

tnotheis (changes requested):

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
NikolaVetnic: Yes, stupid mistake. Resolved.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/Handler.cs
NikolaVetnic: I added one I could think of on my own, if more are required please let me know.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/Handler.cs
tnotheis: 1. You should always question the suggestions of Copilot. Don't just take it as granted.
2. Why did you remove it if you don't know what it does? ;) 
In this case it's actually correct to remove it though. It an be used to control the tracking behavior (https://learn.microsoft.com/en-us/ef/core/querying/tracking). So in this case, since it's a readonly query, we don't need tracking.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/Handler.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        var tierQuotasForIdentity = await Task.WhenAll(tierQuotasForIdentityTasks);
+
+        var quotasForIdentity = individualQuotasForIdentity.Concat(tierQuotasForIdentity).ToList();
+
+        return new ListQuotasForIdentityResponse(quotasForIdentity);
+    }
+}
+
+public static class PaginationHelper
```
tnotheis, line 66: Remove

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/ListQuotasForIdentityResponse.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Backbone.BuildingBlocks.Application.Pagination;
+using Backbone.Modules.Quotas.Application.DTOs;
+
+namespace Backbone.Modules.Quotas.Application.Identities.Queries.GetQuotasForIdentity;
+
+public class ListQuotasForIdentityResponse
+{
+    public IEnumerable<QuotaDTO> Items { get; }
```
tnotheis, line 8: Properties should be placed below the constructors

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/ListQuotasForIdentityResponse.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Backbone.BuildingBlocks.Application.Pagination;
+using Backbone.Modules.Quotas.Application.DTOs;
+
+namespace Backbone.Modules.Quotas.Application.Identities.Queries.GetQuotasForIdentity;
+
+public class ListQuotasForIdentityResponse
```
tnotheis, line 6: Given we have this response, the resulting HTTP response content looks as follows:

```json
{
    "result": {
        "items": [...]
    }
}
```

But what we want is:

```json
{
    "result": [...]
}
```

Have a look at other routes that return multiple items to see how it's done.

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
```diff
@@ -0,0 +1,37 @@
...
+public class QuotasController : ApiControllerBase
+{
+    private readonly ApplicationOptions _options;
+
+    public QuotasController(IMediator mediator, IOptions<ApplicationOptions> options) : base(mediator)
+    {
+        _options = options.Value;
+    }
+
+    [HttpGet("ListIndividualQuotas")]
```
tnotheis, line 28: Hmm, this results in a route like `/api/v1/Quotas/ListIndividualQuotas`. This is wrong for multiple reasons:
1. It doesn't conform to REST conventions (routes should describe resources, and not actions). While this may not always be possible, in this case it is.
2. The task description says clearly states what the route should look like (except for the "api/v1" prefix, which I usually take as granted in the tasks.
3. The route says it lists individual quota, even though it lists individual quota AND tier quota.

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
```diff
@@ -0,0 +1,37 @@
...
+    [ResponseCache(Duration = 1800, Location = ResponseCacheLocation.Any, VaryByQueryKeys = new string[] { "address" })]
```
tnotheis, line 29: If I understand the ResponseCache correctly (and it's possible that I don't, since I never used it), then it is configured wrong here. I assume that this now leads to the same response for every user as long as the cache is valid. Because you set `VaryByQueryKeys` to `address`, but there is no such property in the query. So this means that the cache is the same for all requests.

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
```diff
@@ -0,0 +1,37 @@
...
+    [ProducesResponseType(typeof(List<QuotaDTO>), StatusCodes.Status200OK)]
```
tnotheis, line 30: Did you copy this from somewhere? If so, can you please tell me from where? Because it's incorrect. Look at the `MessagesController` for example, to see how it should look like.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/GetQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        var handler = new Handler(fakeUserContext, fakeIdentitiesRepository, dummyMetricCalculationFactory);
+
+        // Act
+        var quotas = (await handler.Handle(new ListQuotasForIdentityQuery(), CancellationToken.None)).Items.ToList();
+
+        // Assert
+        quotas.Should().HaveCount(2);
+
+        quotas[0].Id.Should().Be(identity.IndividualQuotas.First().Id);
```
tnotheis, line 53: Here you make an assumption about the implementation, which can easily break the test even though the implementation is still correct. Because imagine that we decide to first add the tier quota and then add the individual quota. 
=> Your test expects that the first entry is a tier quota, so it will fail.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/GetQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using FakeItEasy;
+using FluentAssertions;
+using Xunit;
+using static Backbone.UnitTestTools.Data.TestDataGenerator;
+
+namespace Backbone.Modules.Quotas.Application.Tests.Tests.Identities.GetQuotasForIdentity;
+public class HandlerTests
+{
+    [Fact]
+    public async Task Returns_individual_and_tier_quotas_for_identity()
```
tnotheis, line 25: This test doesn't test the probably most complex and most important logic: the calculation of the current usage.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetQuotasForIdentity/Handler.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        var individualQuotasForIdentityTasks = identity.IndividualQuotas
+            .Select(async q =>
+            {
+                var calculator = _metricCalculatorFactory.CreateFor(q.MetricKey);
+                var usage = await calculator.CalculateUsage(q.Period.CalculateBegin(), q.Period.CalculateEnd(), identity.Address, cancellationToken);
+
+                return new QuotaDTO(
+                    q,
+                    new MetricDTO(new Metric(q.MetricKey, q.MetricKey.ToString()!)),
```
tnotheis, line 39: Why are you just setting the display name of the Metric to the same value as the key?

You can get the metrics (including their display name) from the MetricsRepository. So after you retrieved the Identity, you should get all Metrics, and then use the in memory list to get the metric you're looking for for each quota.

tnotheis (changes requested):

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
NikolaVetnic: I've consulted the docs you provided: [https://learn.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-8.0](https://learn.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-8.0)

There is this piece of code:
```csharp
[Route("api/[controller]/ms")]
[HttpGet]
[ResponseCache(Duration = 10, Location = ResponseCacheLocation.Any, NoStore = false)]
public ContentResult GetTimeMS() => Content(
                  DateTime.Now.Millisecond.ToString());
```

I increased the duration and added the `VaryByQueryKeys` parameter to distinguish between different users, though it wasn't working as expected initially (it has been fixed, or at least an attempt to fix it has been made now).

tnotheis: This comment was referring to the `ProducesResponseType` attribute. Not the `ResponseCache` one.

NikolaVetnic: I misunderstood decorations parameters like this:
```csharp
...typeof(HttpResponseEnvelopeResult<List<QuotaDTO>>)...
```

Fixed now.

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
```diff
@@ -0,0 +1,28 @@
...
+
+[Route("api/v1/[controller]")]
+[Authorize("OpenIddict.Validation.AspNetCore")]
+public class QuotasController : ApiControllerBase
+{
+    public QuotasController(IMediator mediator) : base(mediator) { }
+
+    [HttpGet()]
+    [ProducesResponseType(typeof(HttpResponseEnvelopeResult<List<QuotaDTO>>), StatusCodes.Status200OK)]
+    [ProducesError(StatusCodes.Status400BadRequest)]
```
tnotheis, line 21: In which case will there be a 400?

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/Handler.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        _identityAddress = userContext.GetAddress();
+        _identitiesRepository = identitiesRepository;
+        _metricsRepository = metricsRepository;
+        _metricCalculatorFactory = metricCalculatorFactory;
+    }
+
+    public async Task<ListQuotasForIdentityResponse> Handle(ListQuotasForIdentityQuery request, CancellationToken cancellationToken)
+    {
+        var identity = await _identitiesRepository.Find(_identityAddress, cancellationToken) ?? throw new NotFoundException(nameof(Identity));
```
tnotheis, line 29: I don't think we should throw a NotFoundException here, as this will result in a 404, meaning that it's the user's fault. But actually, since we don't make use of any user input except for the token (which is generated by us), the user can't do anything about it.
So you can just throw an Exception with a descriptive error message. This will result in a 500, which is much more fitting.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/FakeMetricCalculatorFactory.cs
```diff
@@ -0,0 +1,85 @@
...
+        _numberOfSentMessages = numberOfSentMessages;
+        _numberOfRelationships = numberOfRelationships;
+        _numberOfRelationshipTemplates = numberOfRelationshipTemplates;
+        _numberOfTokens = numberOfTokens;
+        _amountOfUsedFileStorageSpace = amountOfUsedFileStorageSpace;
+    }
+
+    protected override IMetricCalculator CreateNumberOfFilesMetricCalculator()
+    {
+        throw new NotImplementedException();
```
tnotheis, line 28: You should implement all methods. It's weird that you can pass a parameter, and then it's never used.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        identity.CreateIndividualQuota(metric2.Key, max, period);
+
+        var fakeUserContext = A.Fake<IUserContext>();
+        A.CallTo(() => fakeUserContext.GetAddress()).Returns(identity.Address);
+
+        var fakeIdentitiesRepository = A.Fake<IIdentitiesRepository>();
+        A.CallTo(() => fakeIdentitiesRepository.Find(identity.Address, A<CancellationToken>._, A<bool>._)).Returns(identity);
+
+        var fakeMetricsRepository = A.Fake<IMetricsRepository>();
+        A.CallTo(() => fakeMetricsRepository.FindAll(CancellationToken.None, false)).Returns(new List<Metric> { metric1, metric2 });
```
tnotheis, line 37:
```suggestion
        A.CallTo(() => fakeMetricsRepository.FindAll(A<CancellationToken>._, A<bool>._)).Returns(new List<Metric> { metric1, metric2 });
```

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        // Act
+        var quotas = (await handler.Handle(new ListQuotasForIdentityQuery(), CancellationToken.None)).ToList();
+
+        // Assert
+        quotas.Should().HaveCount(2);
+
+        var tierQuota = quotas.Single(q => q.Source == QuotaSource.Tier);
+        var individualQuota = quotas.Single(q => q.Source == QuotaSource.Individual);
+
+        tierQuota.Should().NotBeNull();
```
tnotheis, line 52: It cannot be null, because `Single` throws if there is no element matching the condition. So either use `SingleOrDefault` or remove this assertion. I think I would remove the assertion.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        tierQuota.Max.Should().Be(5);
+        tierQuota.Usage.Should().Be(1);
+        tierQuota.Period.Should().Be(identity.TierQuotas.First().Period.ToString());
```
tnotheis, line 55: Why not use `tierQuota` here?

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        tierQuota.Metric.Key.Should().Be(metric1.Key.Value);
+        tierQuota.Metric.DisplayName.Should().Be(metric1.DisplayName);
+
+        individualQuota.Should().NotBeNull();
+        individualQuota.Max.Should().Be(5);
+        individualQuota.Usage.Should().Be(1);
+        individualQuota.Period.Should().Be(identity.IndividualQuotas.First().Period.ToString());
```
tnotheis, line 62: Why not use `individualQuota` here?

tnotheis (changes requested):

NikolaVetnic:
For the quotas PR the following is needed:
```json
[{
	"metricKey": "NumberOfSentMessages",
	"quotas": [{
		"source": "Tier",
		"max": 100,
		"usage": 0,
		"period": "Hour"
	}]
}]
```

Changes are: 
* the results should be grouped by the metric key
* we don't need the display name of the metric
* the quotas inside of the metrics should be sorted by percental exhaustion (descending => highest exhaustion first)

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
NikolaVetnic: Removed, in no case should it return an error.

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/Handler.cs
```diff
@@ -0,0 +1,63 @@
...
+                var usage = await calculator.CalculateUsage(q.Period.CalculateBegin(), q.Period.CalculateEnd(), identity.Address, cancellationToken);
+
+                var metric = new MetricDTO(metrics.Single(m => m.Key == q.MetricKey));
+
+                return new QuotaDTO(q, metric, usage);
+            });
+
+        var tierQuotasForIdentity = await Task.WhenAll(tierQuotasForIdentityTasks);
+
+        var quotasForIdentity = individualQuotasForIdentity.Concat(tierQuotasForIdentity).ToList();
```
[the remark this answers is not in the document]
NikolaVetnic: I omitted this because it seems to cause this error:
```
System.InvalidOperationException: 'A key cannot be configured on 'IndividualQuota' because it is a derived type. The key must be configured on the root type 'Quota'. If you did not intend for 'Quota' to be included in the model, ensure that it is not referenced by a DbSet property on your context, referenced in a configuration call to ModelBuilder, or referenced from a navigation on a type that is included in the model.'
```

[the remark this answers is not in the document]
tnotheis: This is easy to fix. Just exclude the property in the EntityTypeConfiguration. Because by default EF Core maps every public property to a column of the database.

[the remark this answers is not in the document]
NikolaVetnic: I tried with `builder.Ignore(x => x.AllQuotas);` but it didn't work - the `MessagesRepositoryTests` failed with the same message as before.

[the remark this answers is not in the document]
tnotheis: I guess you didn't recreate the migrations after adding the `builder.Ignore`?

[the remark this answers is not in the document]
tnotheis: Or rather the compiled models. You can run the `compile_models.ps1` script directly.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using Backbone.Modules.Quotas.Domain.Aggregates.Metrics;
+using Backbone.Modules.Quotas.Domain.Aggregates.Tiers;
+using FakeItEasy;
+using FluentAssertions;
+using Xunit;
+using static Backbone.UnitTestTools.Data.TestDataGenerator;
+
+namespace Backbone.Modules.Quotas.Application.Tests.Tests.Identities.ListQuotasForIdentity;
+
+public class HandlerTests
```
tnotheis, line 14: You should add a test where you have multiple quotas in a single group.
I would be okay with updating the current test as well.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        var fakeMetricsRepository = A.Fake<IMetricsRepository>();
+        A.CallTo(() => fakeMetricsRepository.FindAll(A<CancellationToken>._, false)).Returns(new List<Metric> { metric1, metric2 });
+
+        var fakeMetricCalculationFactory = new FakeMetricCalculatorFactory(numberOfSentMessages: 1, numberOfTokens: 1);
+
+        var handler = new Handler(fakeUserContext, fakeIdentitiesRepository, fakeMetricsRepository, fakeMetricCalculationFactory);
+
+        // Act
+        var quotas = (await handler.Handle(new ListQuotasForIdentityQuery(), CancellationToken.None)).ToList();
```
tnotheis, line 44: Aren't these quotaGroups?

```suggestion
        var quotaGroups = (await handler.Handle(new ListQuotasForIdentityQuery(), CancellationToken.None)).ToList();
```

tnotheis (changes requested):

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/FakeMetricCalculatorFactory.cs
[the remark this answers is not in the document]
NikolaVetnic: Not sure how to do this due to usage of Quota module specific classes.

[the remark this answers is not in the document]
tnotheis: Other tests doesn't mean other modules.
Maybe I wasn't clear. The only problem is that it is inside of the "ListQuotasForIdentity" folder.

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/Handler.cs
tnotheis: Why did you resolve this comment? You didn't change anything.

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/Handler.cs
```diff
@@ -0,0 +1,70 @@
...
+            };
+        }));
+
+        return singleQuotaDTOs
+            .GroupBy(quota => quota.MetricKey)
+            .Select(group => new QuotaGroupDTO
+            {
+                MetricKey = group.Key,
+                Quotas = group.ToList()
+            });
```
tnotheis, line 68: This is looking way better than your previous implementation.

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/ListQuotasForIdentityResponse.cs
```diff
@@ -0,0 +1,29 @@
...
+
+public class ListQuotasForIdentityResponse
+{
+    public ListQuotasForIdentityResponse(IEnumerable<QuotaGroupDTO> items)
+    {
+        Items = items;
+    }
+
+    public IEnumerable<QuotaGroupDTO> Items { get; private set; }
+}
```
tnotheis, line 14: This leads to the following response:

```json
{
  "items": [
    {
      "metricKey": "metric1",
      "quotas": [...]
    }
  ]
}
```

But what we want is this:

```json
[
  {
    "metricKey": "metric1",
    "quotas": [...]
  }
]
```

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Backbone.Modules.Quotas.ConsumerApi.Controllers;
+
+[Route("api/v1/[controller]")]
+[Authorize("OpenIddict.Validation.AspNetCore")]
+public class QuotasController : ApiControllerBase
+{
+    public QuotasController(IMediator mediator) : base(mediator) { }
+
+    [HttpGet]
+    [ProducesResponseType(typeof(HttpResponseEnvelopeResult<IEnumerable<QuotaGroupDTO>>), StatusCodes.Status200OK)]
```
tnotheis, line 19: This is a lie. What's returned is a `ListQuotasForIdentityResponse`...

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/FakeMetricCalculatorFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    protected override IMetricCalculator CreateUsedFileStorageSpaceCalculator()
+    {
+        return amountOfUsedFileStorageSpace.HasValue ? new FakeMetricCalculator(amountOfUsedFileStorageSpace.Value) : throw new ArgumentNullException(nameof(amountOfUsedFileStorageSpace));
+    }
+}
+
+public class FakeMetricCalculator(int value) : IMetricCalculator
+{
+    private int Value { get; } = value;
```
tnotheis, line 43: Why did you create a property for this?

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
+using Backbone.Modules.Quotas.Domain.Aggregates.Identities;
+using Backbone.Modules.Quotas.Domain.Aggregates.Metrics;
+using Backbone.Modules.Quotas.Domain.Aggregates.Tiers;
+using FakeItEasy;
+using FluentAssertions;
+using Xunit;
+using static Backbone.UnitTestTools.Data.TestDataGenerator;
+
+// ReSharper disable InconsistentNaming
```
tnotheis, line 12: What's this for?

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        // Act
+        var quotaGroupDTOs = (await handler.Handle(new ListQuotasForIdentityQuery(), CancellationToken.None)).Items.ToList();
+
+        // Assert
+        quotaGroupDTOs.Should().HaveCount(2);
+
+        quotaGroupDTOs.Should().ContainSingle(qg => qg.MetricKey == metric1.Key.Value);
+        quotaGroupDTOs.Should().ContainSingle(qg => qg.MetricKey == metric2.Key.Value);
+
+        var singleQuotaDTOs = quotaGroupDTOs.SelectMany(group => group.Quotas).ToList();
```
tnotheis, line 52: You should check group by group instead of combining all quotas into a single list. Because what if, by some weird bug, all quotas are written into the first group, and none in the second one?

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/Handler.cs
NikolaVetnic: No idea to be honest :( It's now taken care of, hopefully message is of good quality as well.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/FakeMetricCalculatorFactory.cs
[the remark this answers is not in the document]
NikolaVetnic: Ok, got it now. Done.

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/Handler.cs
NikolaVetnic: Thanks, I think I'm getting the hang of how these things are handled in C#. One detail I was worried about is the long list of input parameters - is it a problem or not?

Modules/Quotas/src/Quotas.ConsumerApi/Controllers/QuotasController.cs
NikolaVetnic: Ok, true. Now I've changed it like so (as part of the request just above this one):
```csharp
        return Ok(response.Items);
```

The type of `Items` is `IEnumerable<QuotaGroupDTO>`. Should I then leave it as it is now?

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/FakeMetricCalculatorFactory.cs
NikolaVetnic: An overlook, you're right. Removed.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
NikolaVetnic: Names such as this one bother ReSharper: `quotaGroupDTOs` - it would like `quotaGroupDtos`.

tnotheis: Then just use the suggested name :)

tnotheis (changes requested):

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/ListQuotasForIdentity/HandlerTests.cs
NikolaVetnic: Done :)

tnotheis (approved):

PR #363, opened 2023-11-02
HalfOfMonth variable not being used

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description
When the `Counts_entities_within_timeframe_month_quotaPeriod` test was developed, we attempted to pivot all dates around the 15th of the current month so that we could use notions such as _Tomorrow_ or _Yesterday_ and be certain that those dates would fall on the same Month. Turns out that this was not done properly, which led to errors when running this test on the last/first day of each month.

trivia: The commit is dated from October 31st, but it was actually done about the same time as this PR. Changing the system date for debugging had such a side-effect, i.e., I committed my changes before setting the clock to the right date again.

### DIFF
--- a/Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
+++ b/Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
@@ -28,14 +28,7 @@ public class MessagesRepositoryTests
      * such as TOMORROW or YESTERDAY and be certain that those dates fall on the same Month.
      * </summary>
      */
-    private static readonly DateTime TODAY = new(
-        DateTime.UtcNow.Year,
-        DateTime.UtcNow.Month,
-        15,
-        DateTime.UtcNow.Hour,
-        DateTime.UtcNow.Minute,
-        DateTime.UtcNow.Second
-        );
+    private static readonly DateTime TODAY = new(2020, 02, 15, 10, 30, 00);
     private static readonly DateTime YESTERDAY = TODAY.AddDays(-1);
     private static readonly DateTime TOMORROW = TODAY.AddDays(1);
     private static readonly DateTime LAST_YEAR = TODAY.AddYears(-1);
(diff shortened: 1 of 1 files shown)


Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
```diff
@@ -14,16 +14,18 @@
 namespace Backbone.Modules.Quotas.Application.Tests.Tests.Repositories;
 public class MessagesRepositoryTests
 {
+    private static readonly DateTime HALF_OF_MONTH = new(DateTime.Now.Year, DateTime.Now.Month, 15);
```
tnotheis, line 17: Can't we just call this `TODAY`? And then I would also set `SystemTime` to this value in the constructor of the test.

Otherwise, nobody understands why YESTERDAY is set to HALF_OF_MONTH-1

tnotheis (changes requested):

Modules/Quotas/test/Quotas.Application.Tests/Tests/Repositories/MessagesRepositoryTests.cs
Dannyps: of course, that helps!

tnotheis (approved):

tnotheis (changes requested):
Wait

tnotheis:
Don't we have to set the SystemTime in the constructor?

Dannyps:
We're setting the time in the `CreateMessage` method because it is only required for the arrangement of the tests. The Act and Assert can run with any time.

Dannyps:
1. It would indeed make more sense to use TODAY, despite the tests being written in a way that works with the actual date. I'll make it so.
2. I see your point and even agree with it to some extent, but it should be noted that it was the fact that we did not hardcode a date that allowed us to find the bug we're now fixing, and I wonder what else may come out of that.
3. Of course.

tnotheis:
> it should be noted that it was the fact that we did not hardcode a date that allowed us to find the bug we're now fixing, and I wonder what else may come out of that.

But there is nothing worse than non-reproducible tests. 

If there is a bug your current tests don't "see", then you are missing a test. So maybe you should think about what edge cases could exist, and write tests for them.

Dannyps:
You're correct. Sadly, I'm pretty confident that there are no further edge cases.

tnotheis (approved):

PR #367, opened 2023-11-06
Show current quota usage on Identity View

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description
This PR makes it possible to check the usage of the Quotas set for a certain Identity when viewing it:
image

### DIFF
--- a/Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
+++ b/Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
@@ -19,12 +19,9 @@ public class GetIdentityResponse
         {
             var usage = await CalculateUsage(metricCalculatorFactory, quota, identityAddress, cancellationToken);
             quotasList.Add(new QuotaDTO(
-                    quota.Id,
-                    quota is TierQuota ? QuotaSource.Tier : QuotaSource.Individual,
+                    quota,
                     new MetricDTO(metrics.First(m => m.Key == quota.MetricKey)),
-                    usage,
-                    quota.Max,
-                    quota.Period.ToString()
+                    usage
                 ));
         }
 
--- a/Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
+++ b/Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
@@ -31,7 +31,7 @@ public class GetIdentityResponseBuilder
             var usage = await CalculateUsageAsync(_metricCalculatorFactory, quota, identityAddress, cancellationToken);
             quotasList.Add(new QuotaDTO(
                     quota.Id,
-                    QuotaSource.Individual,
+                    quota is TierQuota ? QuotaSource.Tier : QuotaSource.Individual,
                     new MetricDTO(metrics.First(m => m.Key == quota.MetricKey)),
                     usage,
                     quota.Max,
--- a/Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
+++ b/Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
@@ -8,18 +8,8 @@ public class GetIdentityResponse
 {
     public string Address { get; set; }
     public IEnumerable<QuotaDTO> Quotas { get; set; }
-}
 
-public class GetIdentityResponseBuilder
-{
-    private readonly MetricCalculatorFactory _metricCalculatorFactory;
-
-    public GetIdentityResponseBuilder(MetricCalculatorFactory metricCalculatorFactory)
-    {
-        _metricCalculatorFactory = metricCalculatorFactory;
-    }
-
-    public async Task<GetIdentityResponse> BuildResponse(string identityAddress, IEnumerable<TierQuota> tierQuotas, IEnumerable<IndividualQuota> individualQuotas, IEnumerable<Metric> metrics, CancellationToken cancellationToken)
+    public static async Task<GetIdentityResponse> Create(MetricCalculatorFactory metricCalculatorFactory, string identityAddress, IEnumerable<TierQuota> tierQuotas, IEnumerable<IndividualQuota> individualQuotas, IEnumerable<Metric> metrics, CancellationToken cancellationToken)
     {
         var quotasList = new List<QuotaDTO>();
 
@@ -27,7 +17,7 @@ public class GetIdentityResponseBuilder
 
         foreach (var quota in allQuotas)
         {
-            var usage = await CalculateUsageAsync(_metricCalculatorFactory, quota, identityAddress, cancellationToken);
+            var usage = await CalculateUsage(metricCalculatorFactory, quota, identityAddress, cancellationToken);
             quotasList.Add(new QuotaDTO(
                     quota.Id,
                     quota is TierQuota ? QuotaSource.Tier : QuotaSource.Individual,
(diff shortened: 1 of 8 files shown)


Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
```diff
@@ -1,36 +1,54 @@
...
-        Quotas = quotasList;
+
+        var allQuotas = (individualQuotas as IEnumerable<Quota>).Union(tierQuotas);
+
+        foreach (var quota in allQuotas)
+        {
+            var usage = await CalculateUsageAsync(_metricCalculatorFactory, quota, identityAddress, cancellationToken);
+            quotasList.Add(new QuotaDTO(
+                    quota.Id,
+                    QuotaSource.Individual,
```
JLSRKonk, line 34: Even though we have both individualQuotas and tierQuotas in the allQuotas list, we are setting all of them as Individual Quotas here.

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/Handler.cs
```diff
@@ -1,18 +1,21 @@
...
 
 namespace Backbone.Modules.Quotas.Application.Identities.Queries.GetIdentity;
 public class Handler : IRequestHandler<GetIdentityQuery, GetIdentityResponse>
 {
     private readonly IIdentitiesRepository _identitiesRepository;
     private readonly IMetricsRepository _metricsRepository;
+    private readonly MetricCalculatorFactory _metricCalculatorFactory;
 
-    public Handler(IIdentitiesRepository identitiesRepository, IMetricsRepository metricsRepository)
+    public Handler(IIdentitiesRepository identitiesRepository, IMetricsRepository metricsRepository, MetricCalculatorFactory factory)
```
JLSRKonk, line 14: We should probably rename this.
```suggestion
    public Handler(IIdentitiesRepository identitiesRepository, IMetricsRepository metricsRepository, MetricCalculatorFactory metricCalculatorFactory)
```

JLSRKonk (changes requested):

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
Dannyps: Nice catch, thanks

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
```diff
@@ -1,36 +1,53 @@
...
+        {
+            var usage = await CalculateUsageAsync(_metricCalculatorFactory, quota, identityAddress, cancellationToken);
+            quotasList.Add(new QuotaDTO(
+                    quota.Id,
+                    quota is TierQuota ? QuotaSource.Tier : QuotaSource.Individual,
+                    new MetricDTO(metrics.First(m => m.Key == quota.MetricKey)),
+                    usage,
+                    quota.Max,
+                    quota.Period.ToString()
+                ));
```
tnotheis, line 38: How about

```suggestion
            quotasList.Add(new QuotaDTO(
                    quota,
                    metrics.First(m => m.Key == quota.MetricKey),
                    usage,
                ));
```

This
1. makes the mapping of `Source` mapping reusable
2. simplifies the usage

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
Dannyps: This would also pollute the DTO with unwanted attributes. Perhaps we could use a `Create` method like before?

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/Handler.cs
```diff
@@ -21,7 +24,7 @@ public async Task<GetIdentityResponse> Handle(GetIdentityQuery request, Cancella
 
         var metricsKeys = identity.TierQuotas.Select(q => q.MetricKey).Union(identity.IndividualQuotas.Select(q => q.MetricKey));
         var metrics = await _metricsRepository.FindAllWithKeys(metricsKeys, cancellationToken);
-
-        return new GetIdentityResponse(identity.Address, identity.TierQuotas, identity.IndividualQuotas, metrics);
+        var identityResponseBuilder = new GetIdentityResponseBuilder(_metricCalculatorFactory);
+        return await identityResponseBuilder.BuildResponse(identity.Address, identity.TierQuotas, identity.IndividualQuotas, metrics, cancellationToken);
```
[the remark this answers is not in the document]
Dannyps: That's indeed much easier, ty

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
tnotheis: You don't map all the properties of the Quota. You can just choose the ones you like. Sure, you have access to all of them, but that's no problem in my opinion.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/GetIdentity/HandlerTests.cs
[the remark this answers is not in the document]
Dannyps: What is the scope here? File? Namespace?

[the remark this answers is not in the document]
tnotheis: File is enough for now. :)

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/GetIdentity/HandlerTests.cs
```diff
@@ -60,12 +59,11 @@ public async void Gets_identity_quotas_by_address()
...
-        var metricsRepository = A.Fake<IMetricsRepository>();
-        var identitiesRepository = A.Fake<IIdentitiesRepository>();
-        var metricCalculatorFactory = A.Fake<MetricCalculatorFactory>();
-        A.CallTo(() => identitiesRepository.Find(A<string>._, A<CancellationToken>._, A<bool>._)).Returns((Identity)null);
+        var dummyMetricsRepository = A.Dummy<IMetricsRepository>();
+        var stubIdentitiesRepository = A.Fake<IIdentitiesRepository>();
+        A.CallTo(() => stubIdentitiesRepository.Find(A<string>._, A<CancellationToken>._, A<bool>._)).Returns((Identity)null);
 
-        var handler = CreateHandler(identitiesRepository, metricsRepository, metricCalculatorFactory);
+        var handler = CreateHandler(stubIdentitiesRepository, dummyMetricsRepository);
```
tnotheis, line 66: In general, I would recommend to not pass dummies to the CreateHandler method and instead initialize them in the CreateHandler method instead. Since there is no configuration required, it's no problem and it simplifies the tests even further.

tnotheis (changes requested):

Modules/Quotas/test/Quotas.Application.Tests/Tests/Identities/GetIdentity/HandlerTests.cs
Dannyps: The thing is that on line 36 a stub is passed, i.e., there are cases where this is passed. Should I make it optional instead, and create it in the `CreateHandler` method if needed?

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
```diff
@@ -1,36 +1,40 @@
...
-            )
-        ));
 
-        Address = identityAddress;
-        Quotas = quotasList;
+        var allQuotas = (individualQuotas as IEnumerable<Quota>).Union(tierQuotas);
+
+        foreach (var quota in allQuotas)
+        {
+            var usage = await CalculateUsage(metricCalculatorFactory, quota, identityAddress, cancellationToken);
```
tnotheis, line 20: Oh, and I just noticed: you should run parallelize the calculation. Otherwise, for n Quotas the creation of the response takes n*dbResponseTIme.

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Application/Identities/Queries/GetIdentity/GetIdentityResponse.cs
tnotheis: Okay, since EF Core doesn't allow parallel execution of multiple db queries, we should postpone this.

tnotheis (approved):

JLSRKonk (approved):

PR #369, opened 2023-11-07
Tokens: save contents in the database instead of in the blob storage

NikolaVetnic:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Saving tokens content to blob storage feature substituted with persisting to a database table due to performance issues. A database migration adding the `Content` table to the `"Tokens"."Tokens"` field was added. Removed saving the content to the blob storage on token creation. Updated the checks on whether the token content is empty, in which case it is to be filled in from the blob storage for reason of supporting legacy tokens.

Unit test that checks whether the content is retrieved from the blob storage as expected is added.


ConsumerApi/TokensDbContextSeeder.cs
```diff
@@ -0,0 +1,44 @@
...
+        await FillContentColumnsFromBlobStorage(context);
+    }
+
+    private async Task FillContentColumnsFromBlobStorage(TokensDbContext context)
+    {
+        // _blobRootFolder is null when blob storage configuration is not provided, meaning the content of database entries should not be loaded from blob storage
+        if (_blobRootFolder == null)
+            return;
+
+        var tokensWithMissingContents = await context.Tokens.Where(t => t.Content == null).ToListAsync();
```
tnotheis, line 33:
```suggestion
        var tokensWithMissingContent = await context.Tokens.Where(t => t.Content == null).ToListAsync();
```

ConsumerApi/TokensDbContextSeeder.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        foreach (var token in tokensWithMissingContents)
+        {
+            var blobTokenContent = await _blobStorage!.FindAsync(_blobRootFolder!, token.Id);
+            token.Content = blobTokenContent;
```
tnotheis, line 38: Can you remove the setter from the Content property and add/use a `LoadContent` property instead please?

Modules/Tokens/src/Tokens.ConsumerApi/TokensModule.cs
```diff
@@ -27,12 +27,18 @@ public override void ConfigureServices(IServiceCollection services, IConfigurati
...
+            if (parsedConfiguration.Infrastructure.BlobStorage != null)
+            {
+                options.BlobStorageOptions = new()
+                {
+                    CloudProvider = parsedConfiguration.Infrastructure.BlobStorage.CloudProvider,
+                    ConnectionInfo = parsedConfiguration.Infrastructure.BlobStorage.ConnectionInfo,
+                    Container =
+                    parsedConfiguration.Infrastructure.BlobStorage.ContainerName.IsNullOrEmpty()
+                        ? "tokens"
+                        : parsedConfiguration.Infrastructure.BlobStorage.ContainerName
```
tnotheis, line 39: Wrong indentation

Modules/Tokens/src/Tokens.Infrastructure/Persistence/IServiceCollectionExtensions.cs
```diff
@@ -18,13 +18,17 @@ public static void AddPersistence(this IServiceCollection services, Action<Persi
     public static void AddPersistence(this IServiceCollection services, PersistenceOptions options)
     {
         services.AddDatabase(options.DbOptions);
-        services.AddBlobStorage(options.BlobStorageOptions);
-        services.AddRepositories(options.BlobStorageOptions);
+
+        if (options.BlobStorageOptions != null)
+        {
+            services.AddBlobStorage(options.BlobStorageOptions);
+            services.AddRepositories(options.BlobStorageOptions);
```
tnotheis, line 25: Now you only add the repositories if there is a blob storage configuration. So if we remove it, there will be a lot of errors.

tnotheis (changes requested):
You can delete the whole `TokensRepositoryOptions` class. It shouldn't be needed anymore.

Modules/Tokens/src/Tokens.Infrastructure/Persistence/Repository/TokensRepository.cs
```diff
@@ -82,28 +71,16 @@ private async Task<DbPaginationResult<Token>> Find(IdentityAddress owner, IEnume
...
-
     #region Write
 
     public async Task Add(Token token)
     {
         await _tokensDbSet.AddAsync(token);
-        _blobStorage.Add(_options.BlobRootFolder, token.Id, token.Content);
-        await _blobStorage.SaveAsync();
+        // _blobStorage.Add(_options.BlobRootFolder, token.Id, token.Content);
+        // await _blobStorage.SaveAsync();
```
tnotheis, line 83: No commented out code

Modules/Tokens/test/Tokens.Application.Tests/Tokens.Application.Tests.csproj
```diff
@@ -18,6 +18,8 @@
 	<ItemGroup>
 		<ProjectReference Include="..\..\..\..\BuildingBlocks\src\UnitTestTools\UnitTestTools.csproj" />
 		<ProjectReference Include="..\..\src\Tokens.Application\Tokens.Application.csproj" />
+        <ProjectReference Include="..\..\..\Devices\test\Devices.Application.Tests\Devices.Application.Tests.csproj" />
```
tnotheis, line 21: Don't reference other modules

Modules/Tokens/test/Tokens.Application.Tests/Tokens.Application.Tests.csproj
```diff
@@ -18,6 +18,8 @@
...
+        <ProjectReference Include="..\..\src\Tokens.Infrastructure\Tokens.Infrastructure.csproj" />
```
tnotheis, line 22: Wrong indentation

tnotheis (changes requested):
You can delete the whole `TokensRepositoryOptions` class. It shouldn't be needed anymore.

ConsumerApi/TokensDbContextSeeder.cs
NikolaVetnic: `TokensRepositoryOptions` is needed in the `TokensDbContextSeeder`:
```csharp
_blobRootFolder = serviceProvider.GetService<IOptions<TokensRepositoryOptions>>()!.Value.BlobRootFolder;
```

ConsumerApi/TokensDbContextSeeder.cs
NikolaVetnic: I made the setter private as it is done in `RelationshipChange` entity, if I should remove it altogether please let me know.

Modules/Tokens/src/Tokens.ConsumerApi/TokensModule.cs
NikolaVetnic: Fixed by moving line 37 to previous line, if it doesn't work I will make it exactly as shown here.

NikolaVetnic:
Some issues I've had with comments:
1. `TokensRepositoryOptions` is needed in the `TokensDbContextSeeder`:
```csharp
_blobRootFolder = serviceProvider.GetService<IOptions<TokensRepositoryOptions>>()!.Value.BlobRootFolder;
```
2. I made the setter private in the `Token` as it is done in RelationshipChange entity, if I should remove it altogether please let me know.

ConsumerApi/TokensDbContextSeeder.cs
tnotheis: That's perfect. Removing it is not possible, of course, because we need to to be able to set the property in the LoadContent method.
I hadn't thought of that.

tnotheis (approved):

tnotheis (approved):

PR #377, opened 2023-11-13
Admin UI: Show Relationships of Identity in Identity Details page

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.

# Screenshots
image


Dannyps (commented):
Did you consider creating a new component to hold the newly shown information? I feel the `IdentityEditComponent` is getting a bit too cluttered.

tnotheis (changes requested):
Instead of using the Relationships module to read the Relationships, you should use the AdminDbContext directly from within the controller.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-edit/identity-edit.component.ts
```diff
@@ -104,6 +130,31 @@ export class IdentityEditComponent {
         });
     }
 
+    public getRelationships(): void {
+        this.loading = true;
+        this.relationshipService.getRelationshipsByParticipantAddress(this.identityAddress!, this.relationshipsPageIndex, this.relationshipsPageSize).subscribe({
+            next: (data: PagedHttpResponseEnvelope<Relationship>) => {
+                this.relationshipsTableData = data.result;
+                this.relationshipsTotalRecords = data.pagination?.totalRecords ? data.pagination.totalRecords : data.result.length;
```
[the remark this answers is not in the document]
JLSRKonk: For entries with one or two relationships, the get response wasn't including the pagination object. From what I understand, this happens if the totalRecords is less than the size of the page, but I'm not really sure about this behavior.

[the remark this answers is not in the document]
tnotheis: Ouh, I think you're right. I was under the impression that we always send pagination data for routes that MAY contain pagination. Sorry, my bad. I will resolve the conversation.

AdminUi/src/AdminUi.Infrastructure.Database.Postgres/Migrations/20231114154951_RelationshipsOverview.cs
```diff
@@ -0,0 +1,41 @@
...
+                        RelationshipChanges."Req_CreatedByDevice" AS "CreatedByDevice",
+                        RelationshipChanges."Res_CreatedByDevice" AS "AnsweredByDevice"
+                    FROM
+                        "Relationships"."Relationships" AS Relationships
+                    LEFT JOIN
+                        "Relationships"."RelationshipChanges" AS RelationshipChanges
+                    ON
+                        Relationships."Id" = RelationshipChanges."RelationshipId"
+                    WHERE
+                        RelationshipChanges."Type" = 10
```
tnotheis, line 31: You should either use double quotes for the table/column/AS names or not, but the mixture is confusing. :)

AdminUi/src/AdminUi.Infrastructure/DTOs/RelationshipDTO.cs
```diff
@@ -0,0 +1,26 @@
+﻿using Backbone.DevelopmentKit.Identity.ValueObjects;
+
+namespace Backbone.AdminUi.Infrastructure.DTOs;
+public class RelationshipDTO
+{
+    public string Peer { get; set; }
+    public string RequestedBy { get; set; }
+    public string? TemplateId { get; set; }
```
tnotheis, line 8: Hmm, I wasn't aware that this column/property is actually nullable. Can you make the property not nullable and create a migration please?

AdminUi/src/AdminUi/ClientApp/src/app/app.module.ts
```diff
@@ -75,7 +76,8 @@ import { XSRFInterceptor } from "./shared/interceptors/xsrf.interceptor";
         ConfirmationDialogComponent,
         LoginComponent,
         ChangeSecretDialogComponent,
-        IdentitiesOverviewComponent
+        IdentitiesOverviewComponent,
+        IdentityEditRelationshipsComponent
```
tnotheis, line 80: Why is it called **Edit**Relationships? We don't want to edit them. It's a readonly view.

AdminUi/src/AdminUi.Infrastructure/DTOs/RelationshipDTO.cs
```diff
@@ -0,0 +1,26 @@
...
+    public RelationshipStatus Status { get; set; }
+    public DateTime CreationDate { get; set; }
+    public DateTime? AnsweredAt { get; set; }
+    public string CreatedByDevice { get; set; }
+    public string? AnsweredByDevice { get; set; }
+
+    public RelationshipDTO(IdentityAddress participantAddress, RelationshipOverview relationshipOverview)
```
tnotheis, line 15:
```suggestion
    public RelationshipDTO(IdentityAddress pointOfView, RelationshipOverview relationshipOverview)
```

AdminUi/src/AdminUi/Controllers/RelationshipsController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public RelationshipsController(IMediator mediator, IOptions<ApplicationOptions> options, AdminUiDbContext adminUiDbContext) : base(mediator)
+    {
+        _adminUiDbContext = adminUiDbContext;
+        _options = options.Value;
+    }
+
+    [HttpGet]
+    [ProducesResponseType(typeof(PagedHttpResponseEnvelope<RelationshipDTO>), StatusCodes.Status200OK)]
+    public async Task<IActionResult> GetAllRelationshipsByParticipantAddress([FromQuery] string participant, [FromQuery] PaginationFilter paginationFilter, CancellationToken cancellationToken)
```
tnotheis, line 33: I would just call this method `GetAllRelationships`. We may need some additional query parameters in the future, and I wouldn't create a separate route for all of them.

AdminUi/src/AdminUi/Controllers/RelationshipsController.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        paginationFilter.PageSize ??= _options.Pagination.DefaultPageSize;
+        if (paginationFilter.PageSize > _options.Pagination.MaxPageSize)
+            throw new ApplicationException(
+                GenericApplicationErrors.Validation.InvalidPageSize(_options.Pagination.MaxPageSize));
+
+        var relationshipOverviews = await _adminUiDbContext.RelationshipOverviews
+            .AsQueryable()
```
tnotheis, line 41: `.AsQueryable` shouldn't be necessary. A `DbSet` already implements `IQueryable`.

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/app.module.ts
JLSRKonk: The file structure in the AdminUi for Identities is currently like this:
image
So name would be better interpreted like IdentityEditComponent => RelationshipsComponent. As the RelationshipsComponent is part of the IdentityEditComponent.

Maybe we should instead rename the IdentityEditComponent to IdentityDetailsComponent, that way IdentityDetailsRelationships wouldn't look out of place.

tnotheis: I think I like this more, yes. In fact, this would apply to the other edit-components as well. In my opinion, this is much more fitting, because depending on where you are, there aren't even "edit" capabilities in this view. 

But you don't need to change all names right now. We can do this in another PR. Just the identity one is okay for now.

AdminUi/src/AdminUi.Infrastructure/DTOs/RelationshipDTO.cs
JLSRKonk: By making this column/property not nullable, are you refering just to the property on the DTO or also the property on the actual Relationship entity?

tnotheis: Since I said "and create a migration", I mean the entity of course. :)

AdminUi/src/AdminUi.Infrastructure.Database.SqlServer/Migrations/20231117101915_RelationshipsOverview.cs
```diff
@@ -0,0 +1,41 @@
...
+                        "RelationshipChanges"."Req_CreatedByDevice" AS "CreatedByDevice",
+                        "RelationshipChanges"."Res_CreatedByDevice" AS "AnsweredByDevice"
+                    FROM
+                        "Relationships"."Relationships" AS "Relationships"
+                    LEFT JOIN
+                        "Relationships"."RelationshipChanges" AS "RelationshipChanges"
+                    ON
+                        "Relationships"."Id" = "RelationshipChanges"."RelationshipId"
+                    WHERE
+                        "RelationshipChanges"."Type" = 10
```
tnotheis, line 31: Is this correct SQL Server syntax? I'm under the impression that SQL Server needs brackets (`[` and `]`) instead of double quotes.

Modules/Relationships/src/Relationships.Infrastructure/Persistence/Database/EntityTypeConfigurations/RelationshipTemplateEntityTypeConfiguration.cs
```diff
@@ -11,5 +11,12 @@ public void Configure(EntityTypeBuilder<RelationshipTemplate> builder)
         builder.HasIndex(x => x.CreatedBy);
         builder.HasIndex(x => x.DeletedAt);
         builder.HasIndex(x => x.ExpiresAt);
+
+        builder
```
tnotheis, line 15: I don't understand what you're doing here. All you should have done was removing the questionmark from the `Relationship.RelationshipTemplate` and `Relationship.RelationshipTemplateId` property.

tnotheis (changes requested):

Modules/Relationships/src/Relationships.Infrastructure/Persistence/Database/EntityTypeConfigurations/RelationshipTemplateEntityTypeConfiguration.cs
JLSRKonk: When I created migrations with only that change, the migrations had a OnDeleteCascade for the RelationshipTemplate. This was not the behavior we had before so I tried changing it so the deletion gets Restricted when there are Relationships using the current RelationshipTemplate.

To implement that in the Configuration, I had to explicitly write out the relation between the Relationship Template table and the Relationship table.

tnotheis: Ohh, I see. In that case: well done! :)

AdminUi/src/AdminUi.Infrastructure.Database.SqlServer/Migrations/20231117101915_RelationshipsOverview.cs
JLSRKonk: I just tried resetting the database and tested the migrations and they ran without any problems on my end.
Here is also a screenshot of the query with double quotes in the management studio:
image
So in terms of functionality it seems to be working, but I can change it to keep consistency with the previous migrations.

tnotheis: Sorry, I just realized that my "thumb up" wasn't clear: you should change it for consistency reasons, I guess.

tnotheis (approved):

AdminUi/src/AdminUi/ClientApp/src/app/services/identity-service/identity.service.ts
```diff
@@ -1,7 +1,7 @@
 import { HttpClient } from "@angular/common/http";
 import { Injectable } from "@angular/core";
 import { NGXLogger } from "ngx-logger";
-import ODataFilterBuilder from "odata-filter-builder";
+import { ODataFilterBuilder } from "odata-filter-builder";
```
tnotheis, line 4: Can you talk to Daniel A. about this? Because he fixed this in another way. 
And I don't think that your code does actually compile, does it?

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/services/identity-service/identity.service.ts
JLSRKonk: Sure, I will ask him. But with these changes the app does compile and is running normally on my side.

tnotheis (approved):

PR #379, opened 2023-11-15
Add Admin UI to docker compose to simplify development workflow

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description
1. The Admin ~UI~ API is automatically launched when running the `docker-compose` project on Visual Studio.
2. The Admin UI (Angular) is served from another container, which updates files on change by means of polling. It should be possible to remove the need to poll if devs on Windows use docker via WSL (and the shared volume is mounted from there), but it is not clear if it is a smart move to open a Visual Studio solution from a WSL directory. Performance constraints should also be evaluated.

# Other minor changes
1. Removed now unused `launchSettings.json` file.
2. Gave a `container_name` to the `seq` service.


AdminUi/src/AdminUi/ClientApp/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:20
+
+RUN echo "RUN echo"
```
daniel-almeida-konkconsulting, line 3: Remove this

AdminUi/src/AdminUi/ClientApp/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+WORKDIR /app
+
+COPY AdminUi/src/AdminUi/ClientApp/package*.json ./
+COPY AdminUi/src/AdminUi/ClientApp/tsconfig.json ./
+COPY AdminUi/src/AdminUi/ClientApp/tsconfig.app.json ./
+
+RUN npm install
+
+RUN npm install -g @angular/cli
```
daniel-almeida-konkconsulting, line 13: Why is this separated from the standard npm install call?

AdminUi/src/AdminUi/Dockerfile
```diff
@@ -1,61 +1,42 @@
...
-COPY ["Common/src/Common.Infrastructure/Common.Infrastructure.csproj", "Common/src/Common.Infrastructure/"]
+COPY ["AdminUi/src/AdminUi.Infrastructure.Database.Postgres/AdminUi.Infrastructure.Database.Postgres.csproj", "AdminUi/src/AdminUi.Infrastructure.Database.Postgres/"]
+COPY ["AdminUi/src/AdminUi.Infrastructure/AdminUi.Infrastructure.csproj", "AdminUi/src/AdminUi.Infrastructure/"]
+COPY ["AdminUi/src/AdminUi.Infrastructure.Database.SqlServer/AdminUi.Infrastructure.Database.SqlServer.csproj", "AdminUi/src/AdminUi.Infrastructure.Database.SqlServer/"]
 RUN dotnet restore "AdminUi/src/AdminUi/AdminUi.csproj"
 COPY . .
-WORKDIR /src/AdminUi/src/AdminUi
-RUN dotnet build /property:WarningLevel=0 --configuration Release --output /app/build --no-restore "AdminUi.csproj" 
+WORKDIR "/src/AdminUi/src/AdminUi"
+RUN dotnet build "AdminUi.csproj" -c Release -o /app/build
```
daniel-almeida-konkconsulting, line 34: Why remove the `/property:WarningLevel=0` and the `--no-restore` arguments?

AdminUi/src/AdminUi/Dockerfile
```diff
@@ -1,61 +1,42 @@
...
-    curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg && \
-    echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list && \
-    apt update && \
-    apt install --yes nodejs
-
-# install angular cli
-RUN npm i -g @angular/cli
-
-RUN dotnet publish /property:WarningLevel=0 /p:UseAppHost=false --no-restore --configuration Release --output /app/publish "AdminUi.csproj"
+RUN dotnet publish "AdminUi.csproj" -c Release -o /app/publish /p:UseAppHost=false
```
daniel-almeida-konkconsulting, line 37: Same as above.

docker-compose/docker-compose.yml
```diff
@@ -87,15 +85,50 @@ services:
...
+    build:
+      context: ..
+      dockerfile: AdminUi/src/AdminUi/Dockerfile.debug
+    ports:
+      - "5173:5000"
+    depends_on:
+      # - ms-sql-server
+      - rabbitmq
+      - postgres
+      - seq
```
daniel-almeida-konkconsulting, line 111: If the `adminApi` depends_on `seq`, shouldn't the `consumerApi` follow suit?

daniel-almeida-konkconsulting (changes requested):

AdminUi/src/AdminUi/Dockerfile
Dannyps: This was the file generated by Visual Studio when re-adding Docker support to the project. I'll restore those flags for consistency, however.

docker-compose/docker-compose.yml
Dannyps: good catch, ty

AdminUi/src/AdminUi/ClientApp/Dockerfile
Dannyps: It is not needed, true. I have removed it.

AdminUi/src/AdminUi/appsettings.json
```diff
@@ -1,5 +1,6 @@
 {
     "AllowedHosts": "*",
+    "Urls": "http://*:5000",
```
daniel-almeida-konkconsulting, line 3: I remember discussing this property with you, and you were unsure whether it was actually doing something.
Did you get a chance to validate that?

Dannyps: Yes! It affects the port where the API is served in the docker container.

daniel-almeida-konkconsulting (approved):

AdminUi/src/AdminUi/AdminUi.csproj
tnotheis, line 1: Doesn't it make sense to keep the things in this file in case someone want's to run it locally instead of in Docker?

AdminUi/src/AdminUi/ClientApp/package.json
```diff
@@ -4,7 +4,7 @@
     "license": "MIT",
     "scripts": {
         "ng": "ng",
-        "start": "ng serve --port 44480",
+        "start": "ng serve --host 0.0.0.0 --disable-host-check --poll 1000",
```
tnotheis, line 7: 1000 means every second? That sounds like a really short time. Did you experience any implications on performance?

AdminUi/src/AdminUi/Dockerfile
```diff
@@ -1,61 +1,42 @@
...
+
 FROM mcr.microsoft.com/dotnet/aspnet:7.0 AS base
 WORKDIR /app
 EXPOSE 80
 
 FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
-
 WORKDIR /src
 COPY ["Directory.Build.props", "."]
+COPY ["Modules/Directory.Build.props", "Modules/"]
```
tnotheis, line 10: Wtf. Was this actually missing? I wonder why it even built before.

AdminUi/src/AdminUi/Dockerfile
tnotheis, line 1: Did you try to run the docker image built with this Dockerfile to test if everything still works as expected? I'm surprised why you were able to remove that much stuff.

AdminUi/src/AdminUi/Properties/.gitignore
```diff
@@ -0,0 +1 @@
+launchSettings.json
```
tnotheis, line 1: Why do you gitignore this file?

AdminUi/src/AdminUi/appsettings.json
tnotheis, line 3: This should be set via `ASPNETCORE_URLS` environment variable instead. Also, use `0.0.00` instead of `*`. Not sure if it's better, but at least it's consistent with the consumer api then. :)

ConsumerApi/appsettings.override.json
```diff
@@ -79,6 +79,11 @@
             }
         },
         "Relationships": {
+            "BlobStorage": {
+                "CloudProvider": "Azure",
+                "ConnectionInfo": "",
+                "ContainerName": ""
+            },
```
tnotheis, line 86: Remove this. Shouldn't be needed, because we removed blob storage for relationships

docker-compose/docker-compose.yml
```diff
@@ -11,16 +11,17 @@ services:
       - ASPNETCORE_URLS=http://0.0.0.0:80
 
       - Modules__Files__Infrastructure__BlobStorage__ConnectionInfo=${ENMESHED_BLOB_STORAGE_CONNECTION_STRING} # set this environment variable on your local system to an appropriate value (DefaultEndpointsProtocol=https;AccountName=<account-name>;AccountKey=<account-key>;EndpointSuffix=core.windows.net)
-      
+      - Modules__Relationships__Infrastructure__BlobStorage__ConnectionInfo=${ENMESHED_BLOB_STORAGE_CONNECTION_STRING} # set this environment variable on your local system to an appropriate value (DefaultEndpointsProtocol=https;AccountName=<account-name>;AccountKey=<account-key>;EndpointSuffix=core.windows.net)
```
tnotheis, line 14: See previous comment

docker-compose/docker-compose.yml
```diff
@@ -85,15 +84,50 @@ services:
...
     volumes:
       - seq-data:/data
     environment:
       - ACCEPT_EULA=Y
     restart: unless-stopped
 
+  adminApi:
+    image: nmshd-adminapi
+    container_name: adminApi
+    hostname: adminApi
```
tnotheis, line 100: Can you convert this to lowercase please? (same for the consumer api; don't know why it is camel case)

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/package.json
Dannyps: I did not. It is not clear how much this should impact performance, but `1000` the most common value out there in the wild. I suggest we keep an eye on this and increase it if need be.

AdminUi/src/AdminUi/Properties/.gitignore
Dannyps: The file is not used when launching the AdminUi from Docker. Keeping it out makes this clearer.

AdminUi/src/AdminUi/Dockerfile
Dannyps: This file was re-generated using Visual Studio. I'm unsure how the old one worked, but it does sound weird that it would compile without that folder.

AdminUi/src/AdminUi/Dockerfile
Dannyps: This is the Dockerfile used in the `docker-compose.test.yml` file for all integration/transport tests, so I would say we're good.

tnotheis: Right. :) I forgot about that

AdminUi/src/AdminUi/Dockerfile
tnotheis: It's a file (`Directory.Build.props`). But it's weird nonetheless.

AdminUi/src/AdminUi/AdminUi.csproj
tnotheis: Is there a reason why you ignored this?

tnotheis (changes requested):
see open comment

AdminUi/src/AdminUi/AdminUi.csproj
Dannyps: I must have pressed "Resolve conversation" inadvertently. I have added back the option to run the SPA from VS:

image

The AdminUI option load the SPA.
The AdminAPI option only loads the backend.

Naturally, database, rabbitmq and seq's hostnames must be updated accordingly when switching from one to another. (Which I find unsettling. Is there any reason why we can't override those settings with Env vars in the `launchSettings.json` file?)

tnotheis: > Is there any reason why we can't override those settings with Env vars in the launchSettings.json file?

Why do you think that it is not possible? The following should actually work:

```json
    "<profileName>": {
      "commandName": "Project",
      "environmentVariables": {
        "myVariableName": "theValueGoesHere"
      }
    }
```

Dannyps: I know it is possible. The question was if it is a good idea :)

tnotheis: Ah. Yeah, I think it totally makes sense.

tnotheis (approved):

tnotheis (approved):

PR #389, opened 2023-11-16
XSRF token error when switching environments through the same TCP port

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description
This is a bug that affects only the deployment process as far as we can see. The token from environment A gets used on calls to env B after the active env change, which leads to an (expected) error. What we did was detect such an error and reload the relevant tokens.


AdminUi/src/AdminUi/Controllers/Filters/RedirectAntiforgeryValidationFailedResultFilter.cs
```diff
@@ -0,0 +1,21 @@
...
+public class RedirectAntiforgeryValidationFailedResultFilter : IAlwaysRunResultFilter
+{
+    public void OnResultExecuting(ResultExecutingContext context)
+    {
+        if (context.Result is ObjectResult result
+            && result.StatusCode == 400
+            && result.Value is ProblemDetails problemDetails
+            )
+        {
+            problemDetails.Detail = "xsrf-token-may-be-invalid";
```
tnotheis, line 15: Aren't we hiding some potentially important information by overriding the details?

tnotheis (changes requested):

tnotheis:
@Dannyps What about this PR?

AdminUi/src/AdminUi/Controllers/Filters/RedirectAntiforgeryValidationFailedResultFilter.cs
Dannyps: In all my attempts, `problemDetails.Detail` has been null. I can add a check for that, however, if you consider that preferable.

tnotheis: Okay, I'm fine with leaving it as it is then.

tnotheis (approved):

PR #416, opened 2023-11-28
Actual Deletion of Identities

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

# Description

The process of Identity Deletion is a complex one which we have broken into smaller tasks. This may be the biggest of those, for a simple reason: it handles the actual deletion of an Identity. The flow is as follows:
1. The `Devices.Jobs.IdentityDeletion` job will execute the ~`UpdateDeletionProcessesCommand`~ `TriggerRipeDeletionProcessesCommand`, which:
    a. Finds all identities which have their DeletionGracePeriod in the past.
    b. marks those as `DeletionStarted`.
    c. returns the relevant IdentityAddresses.
2. foreach returned IdentityAddress:
    a. Sends a notification to it.
    a. Fires a `FindRelationshipsOfIdentityCommand`.
    a. Calls all available `IdentityDeleter.Delete` methods, in no particular order.

The following are pending:
* ~~For each notification, the creation/publishing of the `PeerIdentityDeleted` external event.~~
* Configuration of `IdentityDeletionConfiguration.DeletionStartsNotification.Message` and other configs not to be static, so that they can be loaded from the `appsettings.yml` file??
* ~Tests for the `FilesRepository`, specifically the `DeleteFilesOfIdentity` method.~


tnotheis:
@daniel-almeida-konkconsulting Can you review this please? According to Daniel Silva's comment in his task:

> I'll be on vacation for the next 3 days, but I'm confidant we have a lot here that can already be reviewed. Thus, here is an update of the task's current status:
> 
> - Implemented IdentityDeleter for each module. The Delete(IdentityAddress) method calls the relevant mediator command(s) to delete/change entities based on the relevant IdentityAddress.
> - Implemented tests for most IdentityDeleter implementations and commands it called. I'm unsure if I missed any; I might have.
> - As stated before, there are scripts to dump the database in a format that can be grep-ed for IdentityAddresses.
> - A job Devices.Jobs.IdentityDeletion has been created. It is the great orchestrator of the deletion, responsible for calling all the IdentityDeleters.
> - Several Cascade Deletion have been set-up. However, further tests should be executed to determine if any more are required.

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateDeletionProcesses/Handler.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Backbone.Modules.Devices.Application.Infrastructure.Persistence.Repository;
+using MediatR;
+using Microsoft.Extensions.Logging;
+
+namespace Backbone.Modules.Devices.Application.Identities.Commands.UpdateDeletionProcesses;
+public class Handler(IIdentitiesRepository identitiesRepository, ILogger<Handler> logger) : IRequestHandler<UpdateDeletionProcessesCommand, UpdateDeletionProcessesResponse>
+{
+    public async Task<UpdateDeletionProcessesResponse> Handle(UpdateDeletionProcessesCommand request, CancellationToken cancellationToken)
+    {
+        var response = new UpdateDeletionProcessesResponse
```
daniel-almeida-konkconsulting, line 10: Why this type of declaration and not something like `var response = new UpdateDeletionProcessesResponse();`

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateDeletionProcesses/Handler.cs
```diff
@@ -0,0 +1,32 @@
...
+        var identities = await identitiesRepository.FindAllActiveWithPastDeletionGracePeriod(cancellationToken, track: true);
+        foreach (var identity in identities)
+        {
+            try
+            {
+                identity.DeletionStarted();
+                await identitiesRepository.Update(identity, cancellationToken);
+                response.IdentityAddresses.Add(identity.Address);
+            }
+            catch (Exception ex)
```
daniel-almeida-konkconsulting, line 24: Should probably be more specific in what exceptions you're handling here.

Modules/Devices/src/Devices.Application/IntegrationEvents/Outgoing/IdentityStatusChangedIntegrationEvent.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Backbone.BuildingBlocks.Application.Abstractions.Infrastructure.EventBus.Events;
+using Backbone.DevelopmentKit.Identity.ValueObjects;
+using Backbone.Modules.Devices.Domain.Entities.Identities;
+
+namespace Backbone.Modules.Devices.Application.IntegrationEvents.Outgoing;
+public class IdentityStatusChangedIntegrationEvent : IntegrationEvent
+{
+    public IdentityAddress IdentityAddress { get; set; }
```
daniel-almeida-konkconsulting, line 8: Not really that important, but we typically place public properties after the constructor

Modules/Devices/src/Devices.Application/PushNotifications/Commands/DeleteRegistrationsByIdentityAddress/Handler.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Backbone.Modules.Devices.Application.Infrastructure.Persistence.Repository;
+using MediatR;
+
+namespace Backbone.Modules.Devices.Application.PushNotifications.Commands.DeleteRegistrationsByIdentityAddress;
+public class Handler(IPnsRegistrationsRepository pnsRegistrationRepository) : IRequestHandler<DeleteRegistrationsByIdentityAddressCommand>
+{
+    public async Task Handle(DeleteRegistrationsByIdentityAddressCommand request, CancellationToken cancellationToken)
```
daniel-almeida-konkconsulting, line 7: Should we have a validator for the command payload?

Modules/Devices/src/Devices.Domain/DomainErrors.cs
```diff
@@ -44,4 +44,16 @@ public static DomainError OnlyOneActiveDeletionProcessAllowed()
     {
         return new DomainError("error.platform.validation.device.onlyOneActiveDeletionProcessAllowed", "Only one active deletion process is allowed.");
     }
+
+    public static DomainError CannotChangeIdentityStatusForIdentityUndergoingDeletion()
```
daniel-almeida-konkconsulting, line 48: Couldn't find any usage for this, possibly a work in progress?

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -84,9 +93,18 @@ private void EnsureNoActiveProcessExists()
             throw new DomainException(DomainErrors.OnlyOneActiveDeletionProcessAllowed());
     }
 }
+public enum IdentityStatus
+{
+    Active,
+    ToBeDeleted,
```
daniel-almeida-konkconsulting, line 99: Is this `ToBeDeleted` status used?

daniel-almeida-konkconsulting: Should also specify each entry's value, like the enum below

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -84,9 +93,18 @@ private void EnsureNoActiveProcessExists()
...
+    Deleting
+}
 
 public enum DeletionProcessStatus
 {
     WaitingForApproval = 0,
-    Approved = 1
+    Approved = 1,
+    Deleting = 2,
+    Cancellled = 3,
```
daniel-almeida-konkconsulting, line 108: Typo in `Cancelled`

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcess.cs
```diff
@@ -59,7 +59,8 @@ private void Approve(DeviceId createdByDevice)
     }
 
     public IdentityDeletionProcessId Id { get; }
-    public DeletionProcessStatus Status { get; private set; }
+    public DeletionProcessStatus Status { get; internal set; }
```
daniel-almeida-konkconsulting, line 62: Why `internal set` here?

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20231121135428_IdentityStatus.cs
```diff
@@ -0,0 +1,64 @@
...
+                nullable: false,
+                defaultValue: 0);
+
+            migrationBuilder.AddForeignKey(
+                name: "FK_IdentityDeletionProcesses_Identities_IdentityAddress",
+                table: "IdentityDeletionProcesses",
+                column: "IdentityAddress",
+                principalTable: "Identities",
+                principalColumn: "Address",
+                onDelete: ReferentialAction.Cascade);
```
daniel-almeida-konkconsulting, line 38: Is this intended? Don't you configure it to set NULL on the entity type configuration?

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/IdentitiesRepository.cs
```diff
@@ -61,7 +63,7 @@ public async Task<int> CountByClientId(string clientId, CancellationToken cancel
 
     public async Task AddUser(ApplicationUser user, string password)
     {
-        var createUserResult = await _userManager.CreateAsync(user, password);
+        var createUserResult = await _serviceProvider.GetRequiredService<UserManager<ApplicationUser>>().CreateAsync(user, password);
```
daniel-almeida-konkconsulting, line 66: Why this change?

Modules/Devices/src/Devices.Jobs.IdentityDeletion/Extensions/EventBusServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
+﻿using Backbone.BuildingBlocks.Infrastructure.EventBus;
+using Backbone.BuildingBlocks.Infrastructure.EventBus.AzureServiceBus;
+using Backbone.BuildingBlocks.Infrastructure.EventBus.GoogleCloudPubSub;
+using Backbone.BuildingBlocks.Infrastructure.EventBus.RabbitMQ;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Backbone.Modules.Devices.Jobs.IdentityDeletion.Extensions;
+
+public static class EventBusServiceCollectionExtensions
```
daniel-almeida-konkconsulting, line 9: You can remove this class.

Modules/Files/src/Files.Application/Identities/Commands/DeleteFilesByIdentity/Handler.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Backbone.Modules.Files.Application.Infrastructure.Persistence.Repository;
+using MediatR;
+
+namespace Backbone.Modules.Files.Application.Identities.Commands.DeleteFilesByIdentity;
+
+public class Handler(IFilesRepository filesRepository) : IRequestHandler<DeleteFilesByIdentityCommand>
+{
+    public async Task Handle(DeleteFilesByIdentityCommand request, CancellationToken cancellationToken)
```
daniel-almeida-konkconsulting, line 8: validate the IdentityAddress?

Modules/Messages/src/Messages.Domain/Entities/Message.cs
```diff
@@ -43,4 +43,19 @@ public void LoadBody(byte[] bytes)
         }
         Body = bytes;
     }
+
+    public void ReplaceIdentityAddress(IdentityAddress oldIdentityAddress, IdentityAddress newIdentityAddress)
+    {
+        if (CreatedBy == oldIdentityAddress)
```
daniel-almeida-konkconsulting, line 49: Should we also check if `oldIdentityAddress != newIdentityAddress`?

Modules/Quotas/src/Quotas.Application/Infrastructure/Persistence/Repository/IIdentitiesRepository.cs
```diff
@@ -10,4 +11,6 @@ public interface IIdentitiesRepository
     Task Update(Identity identity, CancellationToken cancellationToken);
     Task<Identity> Find(string address, CancellationToken cancellationToken, bool track = false);
     Task<IEnumerable<Identity>> FindByAddresses(IReadOnlyCollection<string> identityAddresses, CancellationToken cancellationToken, bool track = false);
+    Task Delete(Identity identity, CancellationToken cancellationToken);
+    Task DeleteIdentities(Expression<Func<Identity, bool>> expression, CancellationToken cancellationToken);
```
daniel-almeida-konkconsulting, line 15: Isn't this just a `DeleteById`? If so, I don't think the use of delegates here serves any purpose other than make the code less readable.

Modules/Relationships/src/Relationships.Application/Identities/IdentityDeleter.cs
```diff
@@ -0,0 +1,17 @@
...
+using MediatR;
+
+namespace Backbone.Modules.Relationships.Application.Identities;
+public class IdentityDeleter(IMediator mediator) : IIdentityDeleter
+{
+    private readonly IMediator _mediator = mediator;
+
+    public async Task Delete(IdentityAddress identityAddress)
+    {
+        await _mediator.Send(new DeleteRelationshipsByIdentityCommand(identityAddress));
```
daniel-almeida-konkconsulting, line 14: Should we validate the outcome of this command before proceeding?

Modules/Relationships/src/Relationships.Application/Infrastructure/Persistence/Repository/IRelationshipTemplatesRepository.cs
```diff
@@ -11,4 +12,6 @@ public interface IRelationshipTemplatesRepository
     Task<RelationshipTemplate> Find(RelationshipTemplateId id, IdentityAddress identityAddress, CancellationToken cancellationToken, bool track = false, bool fillContent = true);
     Task Add(RelationshipTemplate template, CancellationToken cancellationToken);
     Task Update(RelationshipTemplate template);
+    Task Delete(IEnumerable<RelationshipTemplateId> templateIds, CancellationToken cancellationToken);
+    Task DeleteTemplates(Expression<Func<RelationshipTemplate, bool>> filter, CancellationToken cancellationToken);
```
daniel-almeida-konkconsulting, line 16: Same as above, albeit this isn't a delete by id, I think the losses on code readability make it hard to justify this approach.
We don't even have multiple use cases with different filters, and typically when we do, we use QueryableExtensions.

Modules/Synchronization/src/Synchronization.Application/SyncRuns/Commands/DeleteSyncRunsByIdentity/Handler.cs
```diff
@@ -0,0 +1,15 @@
...
+using Backbone.Modules.Synchronization.Domain.Entities.Sync;
+using MediatR;
+using Microsoft.EntityFrameworkCore;
+
+namespace Backbone.Modules.Synchronization.Application.SyncRuns.Commands.DeleteSyncRunsByIdentity;
+public class Handler(ISynchronizationDbContext dbContext) : IRequestHandler<DeleteSyncRunsByIdentityCommand>
+{
+    public async Task Handle(DeleteSyncRunsByIdentityCommand request, CancellationToken cancellationToken)
+    {
+        await dbContext.Set<Datawallet>().Where(d => d.Owner == request.IdentityAddress).ExecuteDeleteAsync(cancellationToken);
```
daniel-almeida-konkconsulting, line 12: Isn't this the responsability of the `DeleteDatawalletsByIdentityCommand`?

daniel-almeida-konkconsulting (changes requested):

daniel-almeida-konkconsulting:
Are we already shifting to primary constructors? I see you're using them on this PR but I thought we were holding off on this and then making the change for the entire project.

daniel-almeida-konkconsulting:
Note: there are some comments that I placed in specific classes that can be taken into account for most of the implementation (e.g., validation of command payloads, use of delegates).

tnotheis:
> Are we already shifting to primary constructors? I see you're using them on this PR but I thought we were holding off on this and then making the change for the entire project.

You're right, we should not use them yet until we decided on when/where to use them.

Modules/Quotas/src/Quotas.Application/Infrastructure/Persistence/Repository/IIdentitiesRepository.cs
tnotheis: It was actually my suggestion. The thing is: if we use a `DeleteById`, we would need a `Find` method that gives us the ids. And there we will have the same problem like we do here.

We discussed a `DeleteByAddress` method or similar, but the decision on what to actually delete should be part of the Domain, and not of the Infrastructure.

tnotheis: Imagine for example the Messages module. A Message should only be deleted if all participating Identities (sender and recipients) are deleted. If that's not the case, the Address of the deleted Identity should only be replaced by a "dummy".
This kind of logic definitely shouldn't be located in the Infrastructure

Modules/Messages/src/Messages.Domain/Entities/Message.cs
tnotheis: In fact, the method should be called `DeleteParticipant`, and it should just accept the address of the Identity that should be deleted. Internally, the method uses our placeholder address to replace all occurrences.

Modules/Synchronization/src/Synchronization.Application/SyncRuns/Commands/DeleteSyncRunsByIdentity/Handler.cs
Dannyps: Yes, this was duplicated. Ty

Modules/Files/src/Files.Application/Identities/Commands/DeleteFilesByIdentity/Handler.cs
Dannyps: I added validation.

Modules/Relationships/src/Relationships.Application/Identities/IdentityDeleter.cs
Dannyps: I don't see a reason to. Going down the call chain, we can see we're simply deleting a few entities via a repo. In the unlikely event that that fails, what should we do?

Modules/Messages/src/Messages.Domain/Entities/Message.cs
Dannyps: I created this method with two parameters because I was not comfortable having a Domain method access the `applicationOptions`, needed for getting the `AddressPrefix` configuration. Does this not make sense?

Modules/Devices/src/Devices.Application/PushNotifications/Commands/DeleteRegistrationsByIdentityAddress/Handler.cs
Dannyps: I have added a validator for all Commands extending `RequestWithIdentityAddress`, though I'm still unsure if it's being used. I need to test further.

tnotheis: It's not ideal to introduce the extra complexity of inheritance just to save a property. Adding the property is a one-time thing. And even the validator you're probably trying to avoid isn't that hard to implement.

I understand your urge to get rid of this duplication. I've made the same mistake (using inheritance wherever possible) in the past. But after reading a few books on the subject, I realised that it might not be the best idea.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/IdentitiesRepository.cs
tnotheis: This is a really bad anti pattern called "Service Locator" (see https://stackoverflow.com/questions/22795459/is-servicelocator-an-anti-pattern). We use dependency injection!

Modules/Messages/src/Messages.Domain/Entities/Message.cs
tnotheis: Hmm, yeah... Interesting. Okay, leave it as it is for now. I will have to rethink this whole address prefix thing soon anyway.

And I think you can ignore Daniel's comment ("Should we also check if oldIdentityAddress != newIdentityAddress?) as well. That's not really a problem, I think.

Modules/Relationships/src/Relationships.Application/Identities/IdentityDeleter.cs
tnotheis: Well, at some point we have to document the error in the Audit Log. But I think this should be done by the caller of this method.

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/IdentitiesRepository.cs
Dannyps: I looked for alternatives at the time of the commit and I have looked again. I don't see how we can create a dummy of `UserManager<ApplicationUser>` for use in this Job. Any suggestions?

Modules/Devices/src/Devices.Application/PushNotifications/Commands/DeleteRegistrationsByIdentityAddress/Handler.cs
Dannyps: In this case, the usage of inheritance allows us (I think) to eliminate the redundancy of Validators, i.e., we only need an implementation of a Validator for all Commands. Isn't that reason enough to keep it? Or this there another way to have such a Validator without inheritance?

Modules/Devices/src/Devices.Application/Identities/Commands/UpdateDeletionProcesses/Handler.cs
Dannyps: Good call, thanks.

Modules/Devices/src/Devices.Application/PushNotifications/Commands/DeleteRegistrationsByIdentityAddress/Handler.cs
tnotheis: This is not how the validators work. You can see how the validation is called in https://github.com/nmshd/backbone/blob/update-deletions-process/BuildingBlocks/src/BuildingBlocks.Application/MediatR/RequestValidationBehavior.cs

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20231121135428_IdentityStatus.cs
Dannyps: It's a great question. If an Identity is deleted on the database level for some reason, this makes all associated identityDeletionProcesses be deleted as well. As far as I know, the only desired remain to be left on the database is the AuditLogEntry(ies), so I don't see how this is an issue.

Dannyps:
> > Are we already shifting to primary constructors? I see you're using them on this PR but I thought we were holding off on this and then making the change for the entire project.
> 
> You're right, we should not use them yet until we decided on when/where to use them.

Readability is greatly improved using these constructors. Nevertheless, should I remove their usage, or is their adoption something to be done soon (i.e., before this PR is merged)?

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20231121135428_IdentityStatus.cs
tnotheis: I agree with @Dannyps

tnotheis:
> > > Are we already shifting to primary constructors? I see you're using them on this PR but I thought we were holding off on this and then making the change for the entire project.
> > 
> > 
> > You're right, we should not use them yet until we decided on when/where to use them.
> 
> Readability is greatly improved using these constructors. Nevertheless, should I remove their usage, or is their adoption something to be done soon (i.e., before this PR is merged)?

I'm not fully convinced of this feature yet. So I don't know if/in which cases I want to use it. So if there aren't too many occurrences, I would probably prefer if you used the "old" way. Otherwise you can keep them.

Modules/Devices/src/Devices.Infrastructure.Database.Postgres/Migrations/20231121135428_IdentityStatus.cs
Dannyps: I have tackled the inconsistency in the latest merge.

Modules/Devices/src/Devices.Application/PushNotifications/Commands/DeleteRegistrationsByIdentityAddress/Handler.cs
Dannyps: @tnotheis I have removed the interface and added a validator for each command. See:
 - 220e406a2d5ea0864c614c09499c6d54fdc0e561
 - ebab6bb04c1d8edc3b2cffc43344f215be1741d8

tnotheis:
@Dannyps Remember to convert this PR draft into a real PR once you're done, so that I know that I can review.
@daniel-almeida-konkconsulting Could you please do another review then?

Dannyps:
@tnotheis I'm not done yet. Will do asap.

I've been getting more and more outputs like this, though. I'd say the process itself is looking solid:
image

tnotheis:
> @tnotheis I'm not done yet. Will do asap.
> 
> I've been getting more and more outputs like this, though. I'd say the process itself is looking solid:

I don't understand. Did you attach the wrong screenshot?

Dannyps:
That's the result of searching the dump for that identity after having deleted it :)

tnotheis:
Ahh, and "like this" means "no search results". Got it. 😄

Modules/Files/src/Files.Infrastructure/Persistence/Database/Repository/FilesRepository.cs
Dannyps, line 1: Ensure tests are available

Modules/Messages/src/Messages.Application/Messages/Commands/DeleteMessagesByIdentity/DeleteMessagesByIdentityCommand.cs
Dannyps, line 1: review TODO comment

Modules/Tokens/src/Tokens.Application/Tokens/Commands/DeleteTokensByIdentity/Handler.cs
Dannyps, line 1: implement filter

Modules/Quotas/src/Quotas.Application/Identities/Commands/DeleteIdentity/Handler.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Backbone.Modules.Quotas.Application.Infrastructure.Persistence.Repository;
+using Backbone.Modules.Quotas.Domain.Aggregates.Identities;
+using MediatR;
+
+namespace Backbone.Modules.Quotas.Application.Identities.Commands.DeleteIdentity;
+public class Handler(IIdentitiesRepository identitiesRepository) : IRequestHandler<DeleteIdentityCommand>
+{
+    public async Task Handle(DeleteIdentityCommand request, CancellationToken cancellationToken)
+    {
+        // Deletion of Individual Quotas, Tier Quotas and MetricStatuses is ensured by Cascade Deletion of Foreign Key
```
Dannyps, line 10:
```suggestion
```

Modules/Challenges/src/Challenges.Application/Challenges/Commands/DeleteChallengesOfIdentity/DeleteChallengesOfIdentityCommand.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Backbone.DevelopmentKit.Identity.ValueObjects;
+using MediatR;
+
+namespace Backbone.Modules.Challenges.Application.Challenges.Commands.DeleteChallengesOfIdentity;
+
+public class DeleteChallengesOfIdentityCommand(IdentityAddress identityAddress) : IRequest
```
daniel-almeida-konkconsulting, line 6: Still using primary constructor here.

daniel-almeida-konkconsulting (changes requested):

Modules/Devices/src/Devices.Infrastructure/Persistence/Repository/PnsRegistrationsRepository.cs
```diff
@@ -57,4 +58,9 @@ public async Task Update(PnsRegistration registration, CancellationToken cancell
         _registrations.Update(registration);
         await _dbContext.SaveChangesAsync(cancellationToken);
     }
+
+    public async Task DeleteIPnsRegistrations(Expression<Func<PnsRegistration, bool>> filter, CancellationToken cancellationToken)
```
daniel-almeida-konkconsulting, line 62: `DeleteIPnsRegistrations`, IPns?

Modules/Devices/src/Devices.Jobs.IdentityDeletion/IdentityDeletionJobConfiguration.cs
```diff
@@ -0,0 +1,18 @@
...
+    [Required]
+    public InfrastructureConfiguration Infrastructure { get; set; } = new();
+}
+
+public class InfrastructureConfiguration
+{
+    [Required]
+    public EventBusConfiguration EventBus { get; set; } = new();
+    [Required]
+    public IServiceCollectionExtensions.DbOptions SqlDatabase { get; set; } = new();
```
daniel-almeida-konkconsulting, line 17: Please, let's finally move these DbOptions into one of our class libraries

daniel-almeida-konkconsulting (changes requested):

Modules/Relationships/src/Relationships.Application/Relationships/Commands/FindRelationshipsOfIdentity/DeleteRelationshipsByIdentityCommand.cs
daniel-almeida-konkconsulting, line 1: Just noticed, the file names for the command and response don't match the actual content.

daniel-almeida-konkconsulting: Btw, this is a query, not a command

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -76,6 +78,13 @@ public IdentityDeletionProcess StartDeletionProcessAsOwner(DeviceId asDevice)
         return deletionProcess;
     }
 
+    public void DeletionStarted()
+    {
+        Status = IdentityStatus.Deleting;
+        var deletionProcess = DeletionProcesses.Single(dp => dp.IsActive());
```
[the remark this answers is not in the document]
Dannyps: It's not something that should happen unless someone messes with the database. Moreover, this is not supposed to be run by the consumer, only the IdentityDeletion job. In light of this, there is nothing wrong in using a proper DomainException, but I don't see a pressing need to.

Modules/Devices/src/Devices.Jobs.IdentityDeletion/IdentityDeletionJobConfiguration.cs
Dannyps: Please let me know if this is what you had in mind 7694b5e2c032b87002606547f2f771506bc00ba3

daniel-almeida-konkconsulting: yes 🎈

tnotheis: Hmm, I don't like having such a change hidden in this PR. This only causes conflicts in places that don't necessarily have to be caused. Also, it makes it hard to find and undo the commit once it is merged.
If we actually want to make this refactoring (and I'm not fully convinced yet) we should do it in a separate PR.
Don't get me wrong: I highly appreciate you making small refactorings even if they are not task-related. But this one is too big (for the reasons mentioned above).

Dannyps: Agreed, naturally, and for the sake of it I did this in a single commit, which can be rolled back, or cherry-picked somewhere else. I'll roll it back for now.

Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/DeletionProcess/DeletionStartsNotification.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Reflection;
+
+namespace Backbone.Modules.Devices.Application.Infrastructure.PushNotifications.DeletionProcess;
+
+[NotificationText(Title = "Your Identity will be deleted")]
+public record DeletionStartsNotification
+{
+    public DeletionStartsNotification(string message) => GetType().GetCustomAttribute<NotificationTextAttribute>().Body = message;
+};
```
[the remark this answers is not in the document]
tnotheis: You should always keep it as simple as possible. The reason I decided to ignore the configuration for the first iteration is that I wanted to get something working asap. 
So in other words: I think it's okay to use the static configuration in this file.

tnotheis:
@daniel-almeida-konkconsulting How's your review goging? Don't forget to notify me once you're done so I know I can start my own review.

tnotheis (commented):
I now reviewed 21/178 files and decided to submit this before I continue to review the rest, as some of the comments contain general annotations that apply to most/all modules, and I want to avoid writing the same comment over and over again.
But so far it's looking pretty good.

Modules/Challenges/src/Challenges.Application/Challenges/Commands/DeleteChallengesOfIdentity/Validator.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Backbone.DevelopmentKit.Identity.ValueObjects;
+using FluentValidation;
+
+namespace Backbone.Modules.Challenges.Application.Challenges.Commands.DeleteChallengesOfIdentity;
+public class Validator : AbstractValidator<DeleteChallengesOfIdentityCommand>
+{
+    public Validator() => RuleFor(x => x.IdentityAddress).Must(x => IdentityAddress.IsValid(x));
```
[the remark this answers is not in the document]
Dannyps: If it is null, a NullReferenceException is propagated from the validator:
```
   at Backbone.DevelopmentKit.Identity.ValueObjects.IdentityAddress.op_Implicit(IdentityAddress identityAddress) in D:\Work\enmeshed\backbone\BuildingBlocks\src\DevelopmentKit.Identity\ValueObjects\IdentityAddress.cs:line 113
   at Backbone.Modules.Challenges.Application.Challenges.Commands.DeleteChallengesOfIdentity.Validator.<>c.<.ctor>b__0_1(IdentityAddress x) in D:\Work\enmeshed\backbone\Modules\Challenges\src\Challenges.Application\Challenges\Commands\DeleteChallengesOfIdentity\Validator.cs:line 7
   at FluentValidation.DefaultValidatorExtensions.<>c__DisplayClass27_0`2.<Must>b__0(T x, TProperty val)
   at FluentValidation.DefaultValidatorExtensions.<>c__DisplayClass28_0`2.<Must>b__0(T x, TProperty val, ValidationContext`1 _)
   at FluentValidation.DefaultValidatorExtensions.<>c__DisplayClass29_0`2.<Must>b__0(T instance, TProperty property, ValidationContext`1 propertyValidatorContext)
   at FluentValidation.Validators.PredicateValidator`2.IsValid(ValidationContext`1 context, TProperty value)
   at FluentValidation.Internal.RuleComponent`2.InvokePropertyValidator(ValidationContext`1 context, TProperty value)
   at FluentValidation.Internal.RuleComponent`2.<ValidateAsync>d__18.MoveNext()
   at System.Runtime.CompilerServices.ValueTaskAwaiter`1.GetResult()
   at FluentValidation.Internal.PropertyRule`2.<ValidateAsync>d__4.MoveNext()
   at FluentValidation.AbstractValidator`1.<ValidateInternalAsync>d__20.MoveNext()
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at FluentValidation.AbstractValidator`1.Validate(ValidationContext`1 context)
   at FluentValidation.AbstractValidator`1.FluentValidation.IValidator.Validate(IValidationContext context)
   at Backbone.BuildingBlocks.Application.MediatR.RequestValidationBehavior`2.<>c__DisplayClass2_0.<Handle>b__0(IValidator`1 v) in D:\Work\enmeshed\backbone\BuildingBlocks\src\BuildingBlocks.Application\MediatR\RequestValidationBehavior.cs:line 24
   at System.Linq.Enumerable.SelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.SelectManySingleSelectorIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.ToList()
   at Backbone.BuildingBlocks.Application.MediatR.RequestValidationBehavior`2.Handle(TRequest request, RequestHandlerDelegate`1 next, CancellationToken cancellationToken) in D:\Work\enmeshed\backbone\BuildingBlocks\src\BuildingBlocks.Application\MediatR\RequestValidationBehavior.cs:line 34
   at MediatR.Wrappers.RequestHandlerWrapperImpl`1.<>c__DisplayClass1_1.<Handle>b__2()
   at Backbone.BuildingBlocks.Application.MediatR.LoggingBehavior`2.<Handle>d__3.MoveNext() in D:\Work\enmeshed\backbone\BuildingBlocks\src\BuildingBlocks.Application\MediatR\LoggingBehavior.cs:line 25
   at Backbone.Modules.Challenges.Application.Identities.IdentityDeleter.<Delete>d__2.MoveNext() in D:\Work\enmeshed\backbone\Modules\Challenges\src\Challenges.Application\Identities\IdentityDeleter.cs:line 18
   at Backbone.Modules.Devices.Jobs.IdentityDeletion.Worker.<StartProcessing>d__7.MoveNext() in D:\Work\enmeshed\backbone\Modules\Devices\src\Devices.Jobs.IdentityDeletion\Worker.cs:line 57```

File: Modules/Challenges/src/Challenges.Infrastructure/Persistence/Database/Repository/ChallengesRepository.cs [hunk 25]
```diff
@@ -31,4 +32,9 @@ public async Task<int> DeleteExpiredChallenges(CancellationToken cancellationTok
     {
         return await _challenges.Where(Challenge.CanBeCleanedUp).ExecuteDeleteAsync(cancellationToken: cancellationToken);
     }
+
+    public async Task DeleteChallenges(Expression<Func<Challenge, bool>> filter, CancellationToken cancellationToken)
+    {
+        await _challenges.Where(filter).ExecuteDeleteAsync(cancellationToken: cancellationToken);
```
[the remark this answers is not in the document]
Dannyps: I thought this was a repeat comment from Daniel Almeida! Fixed, thanks.

Modules/Devices/src/Devices.Application/Devices/Commands/DeleteDevicesOfIdentity/DeleteDevicesOfIdentityCommand.cs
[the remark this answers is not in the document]
Dannyps: Perfect, that simplifies this task. In my early sketched I had things outlined as you describe, I'm not sure where this command came from.

BuildingBlocks/src/BuildingBlocks.Application/MediatR/LoggingBehavior.cs
```diff
@@ -5,7 +5,7 @@
 namespace Backbone.BuildingBlocks.Application.MediatR;
 
 public class LoggingBehavior<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse>
-    where TRequest : IRequest<TResponse>
+    where TRequest : notnull
```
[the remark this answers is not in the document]
Dannyps: As the the commit message of 81537ae9e5ec4ff05dffd86380c5210f5bf15e56
https://github.com/jbogard/MediatR/wiki/Migration-Guide-11.x-to-12.0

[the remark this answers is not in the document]
tnotheis: Hm, weird. I think that constaint made perfect sense.

Modules/Challenges/src/Challenges.Application/Challenges/Commands/DeleteChallengesOfIdentity/Validator.cs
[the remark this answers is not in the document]
Dannyps: I see your point. Should we remove the validators then?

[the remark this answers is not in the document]
tnotheis: No, I like the validators. :) We should probably implement them for all commands (but not in this PR)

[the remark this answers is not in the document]
tnotheis: But that's not good.

Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/DeletionProcess/DeletionStartsNotification.cs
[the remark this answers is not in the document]
tnotheis: As far as I can see, the issue this comment addresses isn't fixed yet.

Modules/Challenges/test/WebApplication1/Challenges.Application.Tests.csproj
tnotheis, line 1: The name of the test project folder is incorrect:

Modules/Challenges/test/**_WebApplication1_**/Challenges.Application.Tests.csproj

Modules/Challenges/test/WebApplication1/Tests/Identities/Commands/DeleteIdentityCommandTests/HandlerTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        var handler = CreateHandler(challengesRepository);
+
+        // Act
+        await handler.Handle(new DeleteChallengesOfIdentityCommand(identityAddress), CancellationToken.None);
+
+        // Assert
+        A.CallTo(() => challengesRepository.DeleteChallenges(A<Expression<Func<Challenge, bool>>>._, A<CancellationToken>._)).MustHaveHappenedOnceExactly();
+    }
+
+    private static Handler CreateHandler(IChallengesRepository challengesRepository = null)
```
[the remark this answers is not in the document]
tnotheis: And of course the check for null as well..

Modules/Challenges/test/WebApplication1/Tests/Identities/IdentityDeleterTests.cs
```diff
@@ -0,0 +1,25 @@
...
+        // Arrange
+        var mediator = A.Fake<IMediator>();
+        var identityAddress = TestDataGenerator.CreateRandomIdentityAddress();
+        var deleter = new IdentityDeleter(mediator);
+
+        // Act
+        await deleter.Delete(identityAddress);
+
+        // Assert
+        A.CallTo(() => mediator.Send(A<DeleteChallengesOfIdentityCommand>.That.Matches(it => it.IdentityAddress == identityAddress), A<CancellationToken>._)).MustHaveHappened();
```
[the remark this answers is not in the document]
tnotheis: `r` is just as unfitting for a type called "DeleteChallengesOfIdentityCommand". There's not even an `r` in the name.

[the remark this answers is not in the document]
Dannyps: It was `request` in my head, no idea where I got that from.

Modules/Challenges/src/Challenges.Application/Challenges/Commands/DeleteChallengesOfIdentity/Validator.cs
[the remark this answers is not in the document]
Dannyps: Agreed. I figure a check before calling the deleters should be enough:

```c#
if (identityAddress == null)
{
    _logger.LogError("The command {commandName} returned a null Identity Address.", nameof(TriggerRipeDeletionProcessesCommand));
    continue;
}
```

Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/DeletionProcess/DeletionStartsNotification.cs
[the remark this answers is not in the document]
Dannyps: I would propose we ship this issue to the next PR.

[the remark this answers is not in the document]
tnotheis: Okay.

Modules/Challenges/src/Challenges.Application/Challenges/Commands/DeleteChallengesOfIdentity/Validator.cs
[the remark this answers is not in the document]
tnotheis: But then you're extracting some part of the input validation into the Handler. Why would we do that?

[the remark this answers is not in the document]
tnotheis: Sorry, that comment was nonsense. ;) What I meant to say is that you then loosen the input validation, so in theory someone could call the Handler without making the null check first.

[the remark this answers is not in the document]
Dannyps: They could, and the validator would throw a `NullReferenceException`. I don't see how else it should behave.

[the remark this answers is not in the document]
Dannyps: Would you rather have an `ArgumentNullException` perhaps?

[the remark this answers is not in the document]
tnotheis: Hmm, no, but the Validator's Validate method should return that it is invalid. Just like in case the address has an invalid format.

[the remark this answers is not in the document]
tnotheis: See Backbone.Modules.Relationships.Application.Relationships.Queries.ListRelationships.ListRelationshipsValidator for an example

PR #441, opened 2023-12-12
Use chiseled images

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


HealthCheck/Program.cs
```diff
@@ -0,0 +1,15 @@
...
+var isValidUri = args.Length == 2 && Uri.IsWellFormedUriString(uriArgument, UriKind.Absolute);
+if (!isValidUri)
+{
+    throw new ArgumentException("A valid URI must be given as first argument");
+}
+
+var uri = new Uri(uriArgument);
+
+var response = await httpClient.GetAsync(uri);
+return response.IsSuccessStatusCode ? 0 : 1;
```
jkoenig134, line 15: is this like `process.exit(0 / 1)` in nodejs?

jkoenig134 (approved):

HealthCheck/Program.cs
tnotheis: Yes. I use so called top level statements here. Under the hood, this looks as follows:

```c#
public class Program 
{
  public static int Main(string[] args) {
    // Here goes all the code that is currently in the Program.cs file.
  }
}
```

And as in many other programming languages, the return code of the main function will be used as the exit code of the process.

jkoenig134: gotcha :)

PR #450, opened 2023-12-14
Consumer API (Synchronization): save encrypted payload in the database instead of in the blob storage

NikolaVetnic:
# Readiness checklist

-   [X] I labeled the PR.

# Description

Similarly to what has already been done with Messages, Tokens and Relationships, the `EncryptedPayload` property of the `DatawalletModification` class is now saved in the database instead of in the blob storage.

To support legacy code the `SynchronizationDbContextSeeder` class has been added which checks for the presence of the blob storage configuration, in which case the database entries are updated with blob storage content.

Database migrations are added to introduce a new column into the `Synchronization.DatawalletModifications` table.


BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
```diff
@@ -41,7 +41,7 @@ public static void AddBlobStorage(this IServiceCollection services, BlobStorageO
 
 public class BlobStorageOptions
 {
-    public string ConnectionInfo { get; set; } = string.Empty;
-    public string Container { get; set; } = string.Empty;
-    public string CloudProvider { get; set; } = string.Empty;
+    public string ConnectionInfo { get; set; }
+    public string Container { get; set; }
+    public string CloudProvider { get; set; }
```
tnotheis, line 46: Make them required please.

ConsumerApi/SynchronizationDbContextSeeder.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    private async Task FillEncryptedPayloadColumnFromBlobStorage(SynchronizationDbContext context)
+    {
+        // _blobRootFolder is null when blob storage configuration is not provided, meaning the content of database entries should not be loaded from blob storage
+        if (_blobRootFolder == null)
+            return;
+
+        var modificationsWithMissingContent = await context.DatawalletModifications
+            .Where(t => t.EncryptedPayload == null)
+            .ToListAsync();
```
tnotheis, line 38: Where is the pagination we were talking about?

ConsumerApi/SynchronizationDbContextSeeder.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    private async Task FillEncryptedContentForModificationsWithoutBlobReference(SynchronizationDbContext context, List<DatawalletModification> modificationsWithMissingContent)
+    {
+        var modificationsWithoutBlobReferences = modificationsWithMissingContent.Where(m => m.BlobReference.IsEmpty()).ToList();
+
+        foreach (var datawalletModification in modificationsWithoutBlobReferences)
+        {
+            try
+            {
+                var blobTokenContent = await _blobStorage!.FindAsync(_blobRootFolder!, datawalletModification.Id);
```
tnotheis, line 92: Copy and paste error
```suggestion
                var blobContent = await _blobStorage!.FindAsync(_blobRootFolder!, datawalletModification.Id);
```

ConsumerApi/SynchronizationDbContextSeeder.cs
```diff
@@ -0,0 +1,105 @@
...
+                datawalletModification.LoadEncryptedPayload(blobTokenContent);
+                context.DatawalletModifications.Update(datawalletModification);
+            }
+            catch (NotFoundException)
+            {
+                // due to missing validation, it was possible to create a relationship template without request content;
+                // therefore we cannot tell whether this exception is an error or not
```
tnotheis, line 99: Copy and paste error
```suggestion
                // The encrypted payload of a datawallet modification is not required.
                // Therefore we cannot tell whether this exception is an error or not
```

Modules/Synchronization/src/Synchronization.Application/Datawallets/Commands/PushDatawalletModifications/Handler.cs
```diff
@@ -21,8 +18,8 @@ public class Handler : IRequestHandler<PushDatawalletModificationsCommand, PushD
 {
     private readonly DeviceId _activeDevice;
     private readonly IdentityAddress _activeIdentity;
-    private readonly IBlobStorage _blobStorage;
-    private readonly BlobOptions _blobOptions;
+    //private readonly IBlobStorage _blobStorage;
+    //private readonly BlobOptions _blobOptions;
```
tnotheis, line 22: Remove

Modules/Synchronization/src/Synchronization.Application/Datawallets/Queries/GetModifications/Handler.cs
```diff
@@ -50,49 +42,15 @@ private async Task<List<DatawalletModificationDTO>> MapToDtos(IEnumerable<Datawa
...
-            {
-                // blob not found means that there is no payload for this modification
-            }
-        }
-        else
-        {
-            payloads.TryGetValue(modification.Index, out var payload);
-            dto.EncryptedPayload = payload;
-        }
+        dto.EncryptedPayload = modification.EncryptedPayload;
```
tnotheis, line 53: This line can be removed, since it's done automatically by AutoMapper based on property name convention.

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
NikolaVetnic: This was left like that on purpose because making them required breaks some of the tests. The problem is in the lines instancing `BlobStorageOptions` such as this one in the `AddBlobStorage` method in the same file:
```csharp
    var options = new BlobStorageOptions();
```

tnotheis: Okay. 👍

ConsumerApi/SynchronizationDbContextSeeder.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    private async Task FillEncryptedPayloadColumnFromBlobStorage(SynchronizationDbContext context)
+    {
+        // _blobRootFolder is null when blob storage configuration is not provided, meaning the content of database entries should not be loaded from blob storage
+        if (_blobRootFolder == null)
+            return;
+
+        var paginationResult = await context.DatawalletModifications
+            .Where(t => t.EncryptedPayload == null)
+            .OrderAndPaginate(m => m.Id, new PaginationFilter(1, 20), CancellationToken.None);
```
tnotheis, line 39: Now you're just migrating the first 20 items... 😑 

Also, you can increase the page size to at least 500.

tnotheis (changes requested):

ConsumerApi/SynchronizationDbContextSeeder.cs
NikolaVetnic: Ok, that's what I was afraid of. I'll get it immediately.

tnotheis (approved):

PR #490, opened 2024-01-10
Admin UI: Show Messages of Identity in the Identity details page

JLSRKonk:
# Readiness checklist

-   [x] I labeled the PR.

# Screenshots

ReceivedMessages
SentMessages1
SentMessages2
SentMessagesDialog


AdminUi/src/AdminUi/ClientApp/src/app/app-routing.module.ts
```diff
@@ -10,13 +10,21 @@ import { IdentityDetailsComponent } from "./components/quotas/identity/identity-
...
     { path: "dashboard", component: DashboardComponent, canActivate: [AuthGuard] },
     { path: "identities", component: IdentityListComponent, canActivate: [AuthGuard] },
-    { path: "identities/:address", component: IdentityDetailsComponent, canActivate: [AuthGuard] },
+    {
+        path: "identities/:address",
+        component: IdentityDetailsComponent,
+        canActivate: [AuthGuard],
+        data: {
+            reuseComponent: true
+        }
```
Dannyps, line 26: I'm not familiar with this. I looked it up and I understand the idea, but I'm having a hard time understanding why it's used here. Could you explain?

Dannyps (dismissed):
Looks good 👌

AdminUi/src/AdminUi/ClientApp/src/app/app-routing.module.ts
JLSRKonk: When we want to link components we generally use something like this:
`await this.router.navigate(["/identities/${identityAddress}"]);`
Which for the most part is fine, but when we want to navigate inside the same component, but with different parameters, using the default router.navigate logic, the current page status is not reset.

We could manually reset the page and run all the methods again, but it becomes tricky when we are also dealing with other imbedded components (Relationships, Messages, etc..), so I figured the best way to handle this is to just make this particular page reload when we navigate with different parameters.

Dannyps: Got it, thanks!

AdminUi/src/AdminUi/ClientApp/src/app/services/message-service/message.service.ts
```diff
@@ -0,0 +1,54 @@
...
+        const httpOptions = {
+            params: new HttpParams()
+                .set("Participant", address)
+                .set("Type", "Outgoing")
+                .set("PageNumber", pageNumber + 1)
+                .set("PageSize", pageSize)
+        };
+
+        return this.http.get<PagedHttpResponseEnvelope<MessageOverview>>(this.apiUrl, httpOptions);
+    }
```
daniel-almeida-konkconsulting, line 39: Can't this just be a single method given that only the `Type` changes?

daniel-almeida-konkconsulting (changes requested):

daniel-almeida-konkconsulting:
Where are the integration tests?

tnotheis:
I just noticed that there are 3 problems in the UI:
1. When there are more than 3 recipients, the first three recipients should be shown, and **additionally** there should be a button that allows the user to open a popup that shows all recipients. But currently only the button is shown.
2. I think that the button should just have a plus-sign instead of the text "Show Recipients".
3. The popup is pretty wide, so there is a lot of empty space. I think it should be way smaller.

tnotheis:
@HunorTotBagi 
1. Can you resolve the conflicts in this branch please?
2. The formatting pipeline failed. Can you fix it?

tnotheis (changes requested):
There are compiler errors and code that should be removed. I will review once this is fixed.

AdminApi/src/AdminApi/ClientApp/package.json
```diff
@@ -57,6 +57,7 @@
         "typescript": "~5.4.3"
     },
     "overrides": {
-        "autoprefixer": "10.4.5"
+        "autoprefixer": "10.4.5",
+        "vite": "5.1.7"
```
tnotheis, line 61: Why do we need `vite` now? I don't see any references

Modules/Messages/src/Messages.Domain/Messages.Domain.csproj
```diff
@@ -1,6 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-    <ItemGroup>
+	<PropertyGroup>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
```
tnotheis, line 5: Can be removed. We set this via Directory.Build.props

tnotheis (changes requested):

Modules/Messages/src/Messages.Domain/Messages.Domain.csproj
HunorTotBagi: Changes made.

tnotheis (approved):

daniel-almeida-konkconsulting (approved):

PR #497, opened 2024-01-18
Fix compiler warnings related to nullability in Synchronization module

NikolaVetnic:
# Readiness checklist

-   [X] I labeled the PR.

# Description

Various code issues and warnings related to `null` values found using the ReSharper tool in the `Synchronization` module were investigated and addressed.


Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/ExhaustionDate.cs
```diff
@@ -33,8 +33,8 @@ public override string ToString()
         return left.Value > right.Value;
     }
 
-    public int CompareTo(ExhaustionDate other)
+    public int CompareTo(ExhaustionDate? other)
```
tnotheis, line 36: This change breaks the implementation of the interface. According to the interface, the parameter of the `CompareTo` method must not be nullable

tnotheis (changes requested):
Why didn't you enable `Nullable` in all the projects?

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/ExhaustionDate.cs
NikolaVetnic: Ok, missed that one - noted.

tnotheis (approved):

PR #500, opened 2024-01-22
Admin UI: Breadcrumbs

stamenione:
# Readiness checklist

-   [x] I labeled the PR.

# Screenshots

image

image

image

## Summary by CodeRabbit

- **New Features**
	- Introduced breadcrumb navigation across the Admin UI for enhanced user navigation experience.
- **Enhancements**
	- Improved route protection with updated authentication checks.
- **Tests**
	- Added comprehensive test suites for the new `BreadcrumbComponent` and `BreadcrumbService`.
- **Chores**
	- Updated `.gitignore` to exclude `.history` directories, enhancing repository cleanliness.


tnotheis:
@NikolaVetnic Can you do a first review? Once you approved, I will have a look myself.
This hopefully spares me some time because you find at least some of the issues.

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.ts
tnotheis, line 1: I'm confused by the naming in this file. There are 3 words that all seem to have the same meaning:
* breadcrumbs
* breadcrumbHistory
* breadcrumbTrail

If there is actually no difference, you should unify the naming.

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    private generateBreadcrumbs(route: ActivatedRoute | null, url = "", breadcrumbs: Breadcrumb[] = []): void {
+        if (!route?.children) {
+            return;
+        }
+
+        const children: ActivatedRoute[] = route.children;
+
+        for (const child of children) {
+            const routeURL: string = child.snapshot.url.map((segment) => segment.path).join("/");
```
tnotheis, line 87: Redundant type definition 
```suggestion
            const routeURL = child.snapshot.url.map((segment) => segment.path).join("/");
```
Make the same change at the other places as well please (e.g. 2 lines below this one)

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private trimBreadcrumbHistory(): void {
+        const currentSize = this.breadcrumbHistory.length;
+        if (currentSize > this.maxHistorySize) {
+            this.breadcrumbHistory = this.breadcrumbHistory.slice(currentSize - this.maxHistorySize);
+        }
+    }
+
+    private generateBreadcrumbs(route: ActivatedRoute | null, url = "", breadcrumbs: Breadcrumb[] = []): void {
```
tnotheis, line 79: I'm not 100% sure, but it appears to me that it should be possible to return a breadcrumb array from this function, instead of passing an array to it which is then getting filled. Could you please try to refactor?
As the Clean Code book says: the less parameters a function has, the better.

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    it("should return the breadcrumb history array", async function () {
+        // Arrange
+        const expectedHistory = [
+            { label: "Home", url: "/home" },
+            { label: "Details", url: "/details" },
+            { label: "Products", url: "/products" }
+        ];
+
+        // Act
+        expectedHistory.forEach((trail) => {
```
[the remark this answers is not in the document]
stamenione: Changing from forEach to map function is duable in 2 out of 3 test cases, in test case "should not modify the internal breadcrumb history array when modifying the result" i was trying to create a deep copy of expectedHistory, so when i use map on something like that it is not duable because you can achieve a deep copy using map if your array contains only simple objects (no nested objects or arrays).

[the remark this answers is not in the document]
tnotheis: I think you misunderstood me. I didn't want to say that you should replace `forEach` with `map`. I just described why I think that `forEach` is different than all the other methods (e.g. `map`).

So you should replace `forEach` with a simple loop:

```ts
for (const x of y) {
  ...
}
```

tnotheis:
@coderabbitai review

tnotheis:
@coderabbitai review

tnotheis:
@coderabbitai review

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.spec.ts
```diff
@@ -64,9 +64,9 @@ describe("BreadcrumbService", function () {
         ];
 
         // Act
-        expectedHistory.forEach((trail) => {
-            breadcrumbService["breadcrumbHistory"].push(trail);
-        });
+        for (const e of expectedHistory) {
```
tnotheis, line 67: `e` is not a good variable name

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.spec.ts
```diff
@@ -99,9 +99,9 @@ describe("BreadcrumbService", function () {
...
```
tnotheis, line 102: See above

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.spec.ts
```diff
@@ -118,9 +118,9 @@ describe("BreadcrumbService", function () {
...
```
tnotheis, line 121: See above

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.ts
tnotheis: Now we still have `breadcrumbHistory` and `breadcrumbTrail`. What's the difference?

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.ts
```diff
@@ -34,10 +34,9 @@ export class BreadcrumbService {
     }
 
     private updateBreadcrumbHistory(): void {
-        const breadcrumbs: Breadcrumb[] = [];
-        this.generateBreadcrumbs(this.activatedRoute.root, "", breadcrumbs);
+        const breadcrumbHistory: Breadcrumb[] = this.generateBreadcrumbHistory(this.activatedRoute.root, "");
```
tnotheis, line 37:
```suggestion
        const breadcrumbHistory = this.generateBreadcrumbHistory(this.activatedRoute.root, "");
```

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/services/breadcrumb-service/breadcrumb.service.ts
stamenione: There is absolutely no difference.. I for some reason created a second copy of breadcrumb array, but i forgot the reason why i did it like that

tnotheis (approved):

PR #502, opened 2024-01-23
Fix compiler warnings related to nullability in Messages module

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Removed various code issues and warnings found in the `Messages` module related to null values revealed by ReSharper.

### DIFF
--- a/Modules/Messages/src/Messages.Application/Messages/Commands/SendMessage/SendMessageResponse.cs
+++ b/Modules/Messages/src/Messages.Application/Messages/Commands/SendMessage/SendMessageResponse.cs
@@ -7,5 +7,5 @@ namespace Backbone.Modules.Messages.Application.Messages.Commands.SendMessage;
 public class SendMessageResponse : IMapTo<Message>
 {
     public required MessageId Id { get; set; }
-    public DateTime CreatedAt { get; set; }
+    public required DateTime CreatedAt { get; set; }
 }
--- a/Modules/Messages/src/Messages.Application/Messages/DTOs/MessageDTO.cs
+++ b/Modules/Messages/src/Messages.Application/Messages/DTOs/MessageDTO.cs
@@ -9,7 +9,7 @@ public class MessageDTO : IMapTo<Message>
 {
     public required MessageId Id { get; set; }
 
-    public DateTime CreatedAt { get; set; }
+    public required DateTime CreatedAt { get; set; }
     public required IdentityAddress CreatedBy { get; set; }
     public required DeviceId CreatedByDevice { get; set; }
 


tnotheis:
@NikolaVetnic Can you please perform a first review?

Modules/Messages/src/Messages.Common/Range.cs
```diff
@@ -10,17 +10,17 @@ public Range(T from, T to)
...
-    public T From { get; set; }
-    public T To { get; set; }
+    public T? From { get; set; }
+    public T? To { get; set; }
 
     public bool HasFrom()
     {
-        return !From.Equals(default(T));
+        return From?.Equals(default(T)) ?? false;
     }
```
NikolaVetnic, line 19: As it is right now I **think** that when `From` is `null` the `From?` will force the flow to return `false` after `??`, and if it is not it will be compared with `default(T)` which may also be null, i.e. it will always be false.

Modules/Messages/src/Messages.Common/Range.cs
```diff
@@ -10,17 +10,17 @@ public Range(T from, T to)
...
 
     public bool HasTo()
     {
-        return !To.Equals(default(T));
+        return To?.Equals(default(T)) ?? false;
```
NikolaVetnic, line 23: Same as above.

Modules/Messages/src/Messages.ConsumerApi/Messages.ConsumerApi.csproj
```diff
@@ -17,4 +13,8 @@
         <PackageReference Include="ReHackt.Extensions.Options.Validation" Version="7.0.1" />
     </ItemGroup>
 
+    <PropertyGroup>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
```
NikolaVetnic, line 19: I think it should be at the top of the file.

Modules/Messages/src/Messages.Domain/Ids/FileId.cs
```diff
@@ -22,9 +22,9 @@ public static FileId Parse(string stringValue)
         return new FileId(stringValue);
     }
 
-    public static bool IsValid(string stringValue)
+    public static bool IsValid(string? stringValue)
```
NikolaVetnic, line 25: First you make the value nullable, and then below you suppress the nullable warning - it's contradicting I think.

Modules/Messages/src/Messages.Domain/Messages.Domain.csproj
```diff
@@ -1,12 +1,12 @@
...
-
 	<ItemGroup>
 		<ProjectReference Include="..\..\..\..\BuildingBlocks\src\DevelopmentKit.Identity\DevelopmentKit.Identity.csproj" />
 		<ProjectReference Include="..\..\..\..\BuildingBlocks\src\Tooling\Tooling.csproj" />
 	</ItemGroup>
 
+    <PropertyGroup>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
```
NikolaVetnic, line 11: To the top.

Modules/Messages/src/Messages.Infrastructure.Database.Postgres/Messages.Infrastructure.Database.Postgres.csproj
```diff
@@ -1,11 +1,11 @@
...
+    <ItemGroup>
+        <ProjectReference Include="..\Messages.Infrastructure\Messages.Infrastructure.csproj" />
+    </ItemGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\Messages.Infrastructure\Messages.Infrastructure.csproj" />
-  </ItemGroup>
+    <PropertyGroup>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
```
NikolaVetnic, line 9: Reverse order, this to the top.

Modules/Messages/src/Messages.Infrastructure.Database.SqlServer/Messages.Infrastructure.Database.SqlServer.csproj
```diff
@@ -1,11 +1,11 @@
...
+    <ItemGroup>
+        <ProjectReference Include="..\Messages.Infrastructure\Messages.Infrastructure.csproj" />
+    </ItemGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="..\Messages.Infrastructure\Messages.Infrastructure.csproj" />
-  </ItemGroup>
+    <PropertyGroup>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
```
NikolaVetnic, line 9: To the top.

Modules/Messages/src/Messages.Infrastructure/Messages.Infrastructure.csproj
```diff
@@ -12,4 +12,9 @@
 		<ProjectReference Include="..\Messages.Application\Messages.Application.csproj" />
 		<ProjectReference Include="..\Messages.Domain\Messages.Domain.csproj" />
 	</ItemGroup>
+
+    <PropertyGroup>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
```
NikolaVetnic, line 19: To the top.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/IServiceCollectionExtensions.cs
```diff
@@ -53,8 +53,8 @@ public static void AddDatabase(this IServiceCollection services, DbOptions optio
 
 public class DbOptions
 {
-    public string Provider { get; set; }
-    public string DbConnectionString { get; set; }
+    public string? Provider { get; set; }
+    public string? DbConnectionString { get; set; }
```
NikolaVetnic, line 57: I think these should be initialized with `null!` since the `DbOptions` are filled like so:
```csharp
    public static void AddDatabase(this IServiceCollection services, Action<DbOptions> setupOptions)
    {
        var options = new DbOptions();
        setupOptions.Invoke(options);

        services.AddDatabase(options);
    }
```

I.e. the object is instanced as an empty object and then passed onto `Invoke` call which fills the data in.

NikolaVetnic (changes requested):

Modules/Messages/src/Messages.Application/Messages/Commands/SendMessage/SendMessageCommand.cs
```diff
@@ -9,17 +9,17 @@ public class SendMessageCommand : IRequest<SendMessageResponse>
...
+    public byte[]? Body { get; set; }
     public ICollection<SendMessageCommandAttachment> Attachments { get; set; } = new List<SendMessageCommandAttachment>();
 }
 
 public class SendMessageCommandRecipientInformation
 {
-    public IdentityAddress Address { get; set; }
-    public byte[] EncryptedKey { get; set; }
+    public IdentityAddress? Address { get; set; }
+    public byte[]? EncryptedKey { get; set; }
```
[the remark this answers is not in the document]
tnotheis: > I am not even sure it would make sense for an object of this class to miss either `Address` or `EncryptedKey`...

It doesn't. :)

Modules/Messages/src/Messages.Common/Range.cs
```diff
@@ -10,17 +10,17 @@ public Range(T from, T to)
...
 
-    public T From { get; set; }
-    public T To { get; set; }
+    public T From { get; set; } = default!;
+    public T To { get; set; } = default!;
 
     public bool HasFrom()
     {
-        return !From.Equals(default(T));
+        return From != null && !From.Equals(default(T));
```
NikolaVetnic, line 18: I think both sides of the `AND` operation are basically the same thing:
1. For most types `null` is indeed the default value, in which case both expressions are saying the same thing.
2. If `null` is not the default value (say, in case of `int` where default is `0`) then the left expression will always evaluate to `true` and is thus useless.

I think the solution here is simply `!From!.Equals(default(T));` where:
1. First `!` means a negation, and
2. Second suppresses the `null` warning which is ok since the goal of the expression is to check whether it is indeed null.

Modules/Messages/src/Messages.Common/Range.cs
```diff
@@ -10,17 +10,17 @@ public Range(T from, T to)
...
     }
 
     public bool HasTo()
     {
-        return !To.Equals(default(T));
+        return To != null && !To.Equals(default(T));
```
NikolaVetnic, line 23: Same as above.

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/MessagesDbContext.cs
```diff
@@ -14,9 +15,9 @@ public class MessagesDbContext : AbstractDbContextBase
 
     public MessagesDbContext(DbContextOptions<MessagesDbContext> options, IServiceProvider serviceProvider) : base(options, serviceProvider) { }
 
-    public virtual DbSet<Message> Messages { get; set; }
-    public virtual DbSet<RecipientInformation> RecipientInformation { get; set; }
-    public virtual DbSet<Relationship> Relationships { get; set; }
+    public virtual DbSet<Message> Messages { get; set; } = null!;
+    public virtual DbSet<RecipientInformation> RecipientInformation { get; set; } = null!;
+    public virtual DbSet<Relationship> Relationships { get; set; } = null!;
```
NikolaVetnic, line 20: This class has an empty constructor (line 12) and is tied to the EF Core in a way so I think message by @tnotheis in the **Product - Backbone** team on Teams regarding the `= null!` initializers.

NikolaVetnic (changes requested):

Modules/Messages/src/Messages.Infrastructure/Persistence/Database/MessagesDbContext.cs
NikolaVetnic: Ok, I may have been wrong here. 

Timo said: *"I was ONLY referring to the DbSet properties of the DbContexts, and NOT about ANY class."*, so it may be ok to leave it here as it is.

@tnotheis am I mistaken with this?

tnotheis: The current implementation is exactly how it's supposed to be. 
The properties are not set in any of the constructors. Instead, they are set via reflection on initialization of the DbContext.

HunorTotBagi: Thanks Timo for the answer. Marking this then as resolved.

Modules/Messages/src/Messages.Common/Range.cs
tnotheis: But if `From` is actually `null`, then `From.Equals(...)` throws a `NullReferenceException`.

It should probably be `Equals(request, default(ListMetricsQuery))`

tnotheis: Sorry, my suggestion was bullshit. :)

I meant Equals(From, default(T))

NikolaVetnic (approved):

tnotheis (dismissed):

tnotheis (approved):

PR #504, opened 2024-01-24
Quotas: fix null warnings

NikolaDmitrasinovic:
# Readiness checklist

-   [x] I labeled the PR.

#Description
Removed code warnings found in the Quotas module related to null values revealed by ReSharper.


Modules/Quotas/src/Quotas.Application/DTOs/MetricDTO.cs
```diff
@@ -14,8 +14,8 @@ public MetricDTO(Metric metric)
         DisplayName = metric.DisplayName;
     }
 
-    public string Key { get; set; }
-    public string DisplayName { get; set; }
+    public string Key { get; set; } = null!;
```
NikolaVetnic, line 17: This is fine as it is, however...

This class is not persisted so I think the empty constructor is not used by the EF Core, and otherwise it has zero references in the project. I tried deleting it - it compiles, but I haven't run the tests. I think it makes sense to try deleting it and then running the tests in the pipeline, and if that works then these `null!` initializers can be dropped as well.

Modules/Quotas/src/Quotas.Application/DTOs/MetricDTO.cs
```diff
@@ -14,8 +14,8 @@ public MetricDTO(Metric metric)
...
+    public string DisplayName { get; set; } = null!;
```
NikolaVetnic, line 18: Same as above.

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/ListQuotasForIdentityResponse.cs
```diff
@@ -11,14 +11,14 @@ public class ListQuotasForIdentityResponse : CollectionResponseBase<QuotaGroupDT
 public class SingleQuotaDTO
 {
     public QuotaSource Source { get; set; }
-    public string MetricKey { get; set; }
+    public required string MetricKey { get; set; }
     public int Max { get; set; }
     public uint Usage { get; set; }
-    public string Period { get; set; }
+    public required string Period { get; set; }
 }
```
NikolaVetnic, line 18: Judging by the instancing of this class through initializers, it seems that all five properties (lines 13-17) can be set as `required`. Even though the current situation doesn't produce any warnings, I think this would be a good thing to do.

@tnotheis what do you think?

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierOfIdentityChanged/TierOfIdentityChangedIntegrationEventHandler.cs
```diff
@@ -21,7 +21,7 @@ public async Task Handle(TierOfIdentityChangedIntegrationEvent @event)
         var identity = await _identitiesRepository.Find(@event.IdentityAddress, CancellationToken.None, track: true);
         var newTier = await _tiersRepository.Find(@event.NewTier, CancellationToken.None, track: true);
 
-        await identity.ChangeTier(newTier, _metricCalculatorFactory, CancellationToken.None);
+        await identity!.ChangeTier(newTier, _metricCalculatorFactory, CancellationToken.None);
```
NikolaVetnic, line 24: I don't think this should be suppressed, see how `null` return value is handled in `Backbone.Modules.Quotas.Application.Identities.Commands.DeleteQuotaForIdentity.Handler` - I think we need some kind of reaction to the `null` returned.

Modules/Quotas/src/Quotas.Domain/Aggregates/FileMetadata/FileMetadata.cs
```diff
@@ -1,8 +1,8 @@
 ﻿namespace Backbone.Modules.Quotas.Domain.Aggregates.FileMetadata;
 public class FileMetadata : ICreatedAt
 {
-    public string Id { get; set; }
+    public string? Id { get; set; }
```
NikolaVetnic, line 4: I checked the `Files.FileMetadata` table, you can see it in pgAdmin, and there you can see the columns corresponding to all of these four properties, therefore I don't think two of those should be nullable.

Modules/Quotas/src/Quotas.Domain/Aggregates/FileMetadata/FileMetadata.cs
```diff
@@ -1,8 +1,8 @@
...
     public DateTime CreatedAt { get; set; }
-    public string CreatedBy { get; set; }
+    public string? CreatedBy { get; set; }
```
NikolaVetnic, line 6: Same as above.

Modules/Quotas/src/Quotas.Domain/Aggregates/Messages/Message.cs
```diff
@@ -1,7 +1,7 @@
 ﻿namespace Backbone.Modules.Quotas.Domain.Aggregates.Messages;
 public class Message : ICreatedAt
 {
-    public string Id { get; set; }
-    public string CreatedBy { get; set; }
+    public string? Id { get; set; }
```
NikolaVetnic, line 4: This is a primary key, I don't think it should be nullable.

Modules/Quotas/src/Quotas.Domain/Aggregates/Messages/Message.cs
```diff
@@ -1,7 +1,7 @@
...
+    public string? CreatedBy { get; set; }
```
NikolaVetnic, line 5: I am not completely sure but I think this should not be nullable either.

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/Relationship.cs
```diff
@@ -1,9 +1,9 @@
 ﻿namespace Backbone.Modules.Quotas.Domain.Aggregates.Relationships;
 public class Relationship : ICreatedAt
 {
-    public string Id { get; set; }
-    public string From { get; set; }
-    public string To { get; set; }
+    public string? Id { get; set; }
```
NikolaVetnic, line 4: This is a primary key, I don't think it should be nullable.

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/Relationship.cs
```diff
@@ -1,9 +1,9 @@
...
+    public string? From { get; set; }
```
NikolaVetnic, line 5: Not sure but I think this should not be nullable either.

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/Relationship.cs
```diff
@@ -1,9 +1,9 @@
...
+    public string? To { get; set; }
```
NikolaVetnic, line 6: Not sure but I think this should not be nullable either.

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/RelationshipTemplate.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Backbone.Modules.Quotas.Domain.Aggregates.Relationships;
 public class RelationshipTemplate : ICreatedAt
 {
-    public string CreatedBy { get; set; }
+    public string? CreatedBy { get; set; }
```
NikolaVetnic, line 4: Not sure but I think this should not be nullable either.

Modules/Quotas/src/Quotas.Domain/Aggregates/Tokens/Token.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Backbone.Modules.Quotas.Domain.Aggregates.Tokens;
 public class Token : ICreatedAt
 {
-    public string CreatedBy { get; set; }
+    public string? CreatedBy { get; set; }
```
NikolaVetnic, line 4: Not sure but I think this should not be nullable either - please check.

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Database/IServiceCollectionExtensions.cs
```diff
@@ -65,8 +65,8 @@ public static void AddDatabase(this IServiceCollection services, DbOptions optio
 
     public class DbOptions
     {
-        public string Provider { get; set; }
-        public string DbConnectionString { get; set; }
+        public string? Provider { get; set; }
```
NikolaVetnic, line 68: I think these should be initialized with `= null!` since `DbOptions` is instanced as empty object but is then immediately filled in with required data like so:
```csharp
    public static void AddDatabase(this IServiceCollection services, Action<DbOptions> setupOptions)
    {
        var options = new DbOptions();
        setupOptions.Invoke(options);

        services.AddDatabase(options);
    }
```

@tnotheis what do you think?

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Database/IServiceCollectionExtensions.cs
```diff
@@ -65,8 +65,8 @@ public static void AddDatabase(this IServiceCollection services, DbOptions optio
...
+        public string? DbConnectionString { get; set; }
```
NikolaVetnic, line 69: Same as above.

Modules/Quotas/src/Quotas.Application/DTOs/MetricDTO.cs
tnotheis: I agree.

Modules/Quotas/src/Quotas.Application/Identities/Queries/ListQuotasForIdentity/ListQuotasForIdentityResponse.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Application/IntegrationEvents/Incoming/TierOfIdentityChanged/TierOfIdentityChangedIntegrationEventHandler.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/FileMetadata/FileMetadata.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Messages/Message.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Messages/Message.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/Relationship.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/Relationship.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/Relationship.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Relationships/RelationshipTemplate.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Tokens/Token.cs
tnotheis: 👍

Modules/Quotas/src/Quotas.Infrastructure/Persistence/Database/IServiceCollectionExtensions.cs
tnotheis: Yeah, I agree

Modules/Quotas/src/Quotas.Application/DTOs/MetricDTO.cs
NikolaVetnic: Unfortunately, we were wrong :) CNS transport tests fail, this is the reason as stated in the logs: `"Message":"Cannot dynamically create an instance of type 'Backbone.Modules.Quotas.Application.DTOs.MetricDTO'. Reason: No parameterless constructor defined."`

So it seems the that constructor has to stay...

NikolaVetnic (approved):

tnotheis (approved):

PR #512, opened 2024-01-31
Fix multiple sync runs error

NikolaVetnic:
# Readiness checklist

-   [X] I labeled the PR.

# Description

Due to the concurrency involved in the sync runs, the test results were not consistent between the runs. However, by setting the breakpoints at `Backbone.Modules.Synchronization.Application.SyncRuns.Commands.StartSyncRun.Handler` (lines 129 and 137) the relevant exception was caught and examined. The `catch` block was previously implemented like so:
```csharp
        catch (DbUpdateException ex)
        {
            if (ex.HasReason(DbUpdateExceptionReason.DuplicateIndex))
                throw new OperationFailedException(ApplicationErrors.SyncRuns.CannotStartSyncRunWhenAnotherSyncRunIsRunning());

            throw;
        }
```

The problem with the tests was that some of the `DbUpdateException` instances were not being caught by the `if` check. Code examination of `DbUpdateException` yields the following condition:
```csharp
            case DbUpdateExceptionReason.DuplicateIndex:
                return ex.Message.Contains("Index") || ex.InnerException != null && ex.InnerException.Message.Contains("UNIQUE");
```

Examination of the exception caught previously revealed the following values:
```json
{
   "ex":{
      "Message":"An error occurred while saving the entity changes. See the inner exception for details.",
      "InnerException":{
         "Message":"23505: duplicate key value violates unique constraint 'IX_SyncRuns_CreatedBy_Index' \nDETAIL: Detail redacted as it may contain sensitive data. Specify 'Include Error Detail' in the connection string to include this information.",
         "SqlState":23505
      }
   }
}
```

In other words, no criteria defined for the `DuplicateIndex` variety of the `DbUpdateExceptionReason` was met. On the other hand, the exception matched the `UniqueKeyViolation` variety well:
```csharp
            case DbUpdateExceptionReason.UniqueKeyViolation:
                return ex.GetBaseException() is SqlException { Number: 2627 } // SqlServer
                    or Npgsql.PostgresException { SqlState: "23505" };
```

For that reason the check in the `catch` block was updated as follows:
```csharp
        catch (DbUpdateException ex)
        {
            if (ex.HasReason(DbUpdateExceptionReason.DuplicateIndex) || ex.HasReason(DbUpdateExceptionReason.UniqueKeyViolation))
                throw new OperationFailedException(ApplicationErrors.SyncRuns.CannotStartSyncRunWhenAnotherSyncRunIsRunning());

            throw;
        }
```

The tests now pass as expected.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for synchronization processes by specifically addressing unique key violation scenarios.


tnotheis:
@coderabbitai review

PR #514, opened 2024-02-01
Admin UI: display error in quota dialog

stamenione:
# Readiness checklist

-   [x] I labeled the PR.

# Description

image

image

## Summary by CodeRabbit

- **New Features**
	- Introduced error handling in quota assignment with user-friendly messages.
	- Added a client secret warning indicator for better security awareness.
- **Bug Fixes**
	- Corrected quota creation logic to prevent quotas with non-positive max values.
	- Enhanced quota assignment dialog to improve data handling and user interaction.
- **Style**
	- Implemented new styling for error and warning messages in the quota management interface.
- **Refactor**
	- Streamlined quota assignment processes and improved data refresh mechanism for immediate feedback.


Modules/Quotas/src/Quotas.Domain/DomainErrors.cs
```diff
@@ -11,7 +11,7 @@ public static DomainError UnsupportedMetricKey()
 
     public static DomainError MaxValueCannotBeLowerOrEqualToZero()
```
tnotheis, line 12:
```suggestion
    public static DomainError MaxValueCannotBeLowerThanZero()
```

AdminUi/src/AdminUi/ClientApp/src/app/services/quotas-service/quotas.service.ts
```diff
@@ -10,10 +10,17 @@ import { environment } from "src/environments/environment";
 export class QuotasService {
     private readonly apiUrl: string;
 
+    private readonly refreshDataSubject = new BehaviorSubject<boolean>(false);
+    public refreshData$ = this.refreshDataSubject.asObservable();
```
tnotheis, line 14: A service should always be stateless. Otherwise there may be weird side effects.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
```diff
@@ -73,7 +109,7 @@ export class AssignQuotasDialogComponent {
             period: this.period!
         };
 
-        this.dialogRef.close(quota);
+        this.createIdentityQuota(quota);
```
tnotheis, line 112: Now the assign-quotas-dialog always creates quota for an identity. But what if we want to use it to create quota for a tier?

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
```diff
@@ -62,6 +66,38 @@ export class AssignQuotasDialogComponent {
...
+                    duration: 4000,
+                    verticalPosition: "top",
+                    horizontalPosition: "center"
+                });
+            },
+            complete: () => {
+                this.loading = false;
+                if (this.isAllOk) {
+                    this.dialogRef.close();
+                }
```
tnotheis, line 92: Why not just move this to `next`? Then you can get rid of this weird `isAllOk` variable.

tnotheis (changes requested):
The error message in your screenshot shows some random JS error. I don't know where exactly it comes from, but it's definitely not the one coming from the backbone (there is no `undefined` in C#)

tnotheis:
@coderabbitai review

tnotheis (changes requested):
What's the state of this PR again? Is the error message problem fixed by now? If so, update the screenshot in the description please

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.html
```diff
@@ -31,6 +31,9 @@ <h2 mat-dialog-title>{{ header }}</h2>
         <mat-progress-spinner color="primary" mode="indeterminate"> </mat-progress-spinner>
     </div>
 </mat-dialog-content>
+<div *ngIf="errorMessage !== '' && loading" class="quota-error-container">
```
tnotheis, line 34: Why does it only show if data is currently loading?

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
```diff
@@ -62,6 +64,56 @@ export class AssignQuotasDialogComponent {
...
     }
 
+    public createQuota(quotaData: AssignQuotaData): void {
+        this.loading = true;
+
+        if (this.data.address) {
+            this.createIdentityQuota(this.data.address, quotaData);
+        } else if (this.data.tierId) {
+            this.createTierQuota(this.data.tierId, quotaData);
+        }
```
tnotheis, line 74: This is an ugly hack. `data` is (as far as I can see) of type `any`. There has to be a different way.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
```diff
@@ -62,6 +64,56 @@ export class AssignQuotasDialogComponent {
...
+                this.snackBar.open("Successfully assigned quota.", "Dismiss");
+                this.dialogRef.close(true);
+            },
+            complete: () => (this.loading = false),
+            error: (err: any) => {
+                this.loading = false;
+                this.errorMessage = err.error?.error?.message ?? err.message;
+            }
+        });
+    }
```
tnotheis, line 115: This dialog should be reusable. Currently it is not, because you have to change it every time you want to use it somewhere. What was wrong with the solution as it was before?

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.html
stamenione: My mistake, i overlooked that one.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
stamenione: I see.. I made it not reusable.. I'll revert my changes and try a different solution. :)

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
tnotheis, line 1: The changes in this file seem way too complicated. I can't believe that all of this is required just to show a simple error message.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-details/identity-details.component.ts
```diff
@@ -217,10 +221,7 @@ export class IdentityDetailsComponent {
             error: (err: any) => {
                 this.loading = false;
                 const errorMessage = err.error?.error?.message ?? err.message;
-                this.snackBar.open(errorMessage, "Dismiss", {
-                    verticalPosition: "top",
-                    horizontalPosition: "center"
-                });
+                this.openAssignQuotaDialog(errorMessage, quotaData);
```
tnotheis, line 224: I don't get it. Why do you reopen the dialog instead of just not closing it at the first place?

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/tier/tier-edit/tier-edit.component.ts
```diff
@@ -142,10 +146,7 @@ export class TierEditComponent {
             error: (err: any) => {
                 this.loading = false;
                 const errorMessage = err.error?.error?.message ?? err.message;
-                this.snackBar.open(errorMessage, "Dismiss", {
-                    verticalPosition: "top",
-                    horizontalPosition: "center"
-                });
+                this.openAssignQuotaDialog(errorMessage, quota);
```
tnotheis, line 149: See above

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
```diff
@@ -44,8 +44,8 @@ public Identity(string address, TierId tierId)
 
     public IndividualQuota CreateIndividualQuota(MetricKey metricKey, int max, QuotaPeriod period)
     {
-        if (max <= 0)
-            throw new DomainException(DomainErrors.MaxValueCannotBeLowerOrEqualToZero());
+        if (max < 0)
+            throw new DomainException(DomainErrors.MaxValueCannotBeLowerThanZero());
```
tnotheis, line 48: Didn't we make this change in another PR already?

Modules/Quotas/src/Quotas.Domain/Aggregates/Tiers/Tier.cs
```diff
@@ -20,8 +20,8 @@ public Tier(TierId id, string name)
 
     public Result<TierQuotaDefinition, DomainError> CreateQuota(MetricKey metricKey, int max, QuotaPeriod period)
     {
-        if (max <= 0)
-            return Result.Failure<TierQuotaDefinition, DomainError>(DomainErrors.MaxValueCannotBeLowerOrEqualToZero());
+        if (max < 0)
+            return Result.Failure<TierQuotaDefinition, DomainError>(DomainErrors.MaxValueCannotBeLowerThanZero());
```
tnotheis, line 24: See above

Modules/Quotas/src/Quotas.Domain/DomainErrors.cs
```diff
@@ -9,9 +9,9 @@ public static DomainError UnsupportedMetricKey()
         return new DomainError("error.platform.quotas.unsupportedMetricKey", $"The given metric key is not supported. The supported metric keys are: [{string.Join(", ", MetricKey.GetSupportedMetricKeyValues())}].");
     }
 
-    public static DomainError MaxValueCannotBeLowerOrEqualToZero()
+    public static DomainError MaxValueCannotBeLowerThanZero()
     {
-        return new DomainError("error.platform.quotas.invalidValueForMaxLimitInQuota", "A quota max value cannot be lower or equal to zero.");
+        return new DomainError("error.platform.quotas.invalidValueForMaxLimitInQuota", "A quota max value cannot be lower than zero.");
```
tnotheis, line 14: See above

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-details/identity-details.component.ts
stamenione: I couldn't figure out a better solution than this.. now i implemented another one with service as a transmitter of data between two components.

Modules/Quotas/src/Quotas.Domain/Aggregates/Identities/Identity.cs
NikolaDmitrasinovic: I'm sure I only made this change here, it was last week though.

tnotheis: Hmm, okay. Maybe I am remembering this from last week then.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-details/identity-details.component.ts
tnotheis: What if you pass a callback function via the `data` of the dialog? I don't like that it lacks type safety, but it's still way better than what's currently implemented.

stamenione: I need something cleared.. do u want to use callback function when the data is sent from dialog to the parent or vise versa or in both cases?

tnotheis: If there is an error message, it can be displayed by the dialog. If there is no error message, the dialog can be closed.

stamenione: I added a callback function.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/tier/tier-edit/tier-edit.component.ts
```diff
@@ -29,6 +29,10 @@ export class TierEditComponent {
     public tier: Tier;
     public loading: boolean;
 
+    public quotaSubscription: Subscription;
```
tnotheis, line 32: This isn't used...

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
```diff
@@ -87,7 +77,12 @@ export class AssignQuotasDialogComponent {
             period: this.period!
         };
 
-        this.dialogRef.close(quota);
+        const callback = this.data.callback;
+        callback(quota);
+    }
+
+    public updateErrorMessage(errorMessage: string): void {
```
tnotheis, line 84: Should rather be called "ShowErrorMessage".

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
```diff
@@ -87,7 +77,12 @@ export class AssignQuotasDialogComponent {
             period: this.period!
         };
 
-        this.dialogRef.close(quota);
+        const callback = this.data.callback;
```
tnotheis, line 80: Isn't the callback available in the constructor already? Because then I would set it to a strongly typed field there already.

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
```diff
@@ -38,42 +39,31 @@ export class AssignQuotasDialogComponent {
...
-                this.max = this.data.assignQuotaData.max;
-                this.period = this.data.assignQuotaData.period;
-            }
-        });
+        this.getMetrics();
         this.getPeriods();
     }
 
-    public getMetrics(): Observable<any> {
+    public getMetrics(): void {
```
tnotheis, line 51: Why is it called `get...` if it doesn't return anything?

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-details/identity-details.component.ts
```diff
@@ -44,6 +44,8 @@ export class IdentityDetailsComponent {
     public updatedTier?: TierOverview;
     public tier?: TierOverview;
 
+    private dialogRef: MatDialogRef<AssignQuotasDialogComponent> | undefined;
```
tnotheis, line 47: Why do you define all other properties with `name?: Type`, and here you suddenly start with `name: Type | undefined`?

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/tier/tier-edit/tier-edit.component.ts
```diff
@@ -29,6 +29,10 @@ export class TierEditComponent {
...
+
+    private dialogRef: MatDialogRef<AssignQuotasDialogComponent> | undefined;
```
tnotheis, line 34: See above

tnotheis (changes requested):

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
stamenione: To be honest I am not sure.. the name of that method is not one of my doing.. I will rename it as "showMetrics"

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/identity/identity-details/identity-details.component.ts
stamenione: I was preoccupied to make dialogRef work in the code and simply forgot to change it to name?: .

AdminUi/src/AdminUi/ClientApp/src/app/components/quotas/assign-quotas-dialog/assign-quotas-dialog.component.ts
stamenione: Yes it is but.. with that i cannot trigger the parent method in the dialog component. One of the ways is doing like this.. now when I think about it I found an even cleaner solution via EventEmmitter.

tnotheis: I don't think that EventEmitter is a good idea. Too much failure potential. If you are on the Identity Details of an Identity, then switch to the Tier Details and don't unsubscribe before that and then assign quota to the Tier, this quota might also end up in the identity.

tnotheis: > with that i cannot trigger the parent method in the dialog component

I don't understand why this shouldn't be possible

stamenione: What I wanted to say there when I trigger the parent component in this dialog component then i would have to trigger all other parent components that are calling this dialog in my case now is two of those.

tnotheis: Hmm, I still don't understand. Why are there multiple parents? One dialog is always only created by one component.

stamenione: Yes, but the dialog is the same and when you want to access a parent method from a component in a dialog it makes not reusable.. if i call a method from a one component to send the data to that components method i cannot send that data to the other component i would simply need to call that other components method and do the same.

tnotheis (dismissed):

tnotheis (approved):

PR #531, opened 2024-02-21
Consumer API: Get single deletion process

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


Modules/Devices/src/Devices.Application/Identities/Queries/GetDeletionProcess/Handler.cs
```diff
@@ -0,0 +1,29 @@
...
+    public Handler(IIdentitiesRepository identitiesRepository, IUserContext userContext)
+    {
+        _identitiesRepository = identitiesRepository;
+        _userContext = userContext;
+    }
+
+    public async Task<IdentityDeletionProcessDTO> Handle(GetDeletionProcessQuery request, CancellationToken cancellationToken)
+    {
+        var identity = await _identitiesRepository.FindByAddress(_userContext.GetAddress(), cancellationToken) ?? throw new NotFoundException(nameof(Identity));
+        var deletionProcess = identity.DeletionProcesses.FirstOrDefault(p => p.Id == request.Id) ?? throw new NotFoundException(nameof(IdentityDeletionProcess));
```
NikolaVetnic, line 24: This approach certainly works, but maybe it's a bit of an overkill to fetch the entire identity and all its deletion processes just to get the one you need... On the other hand, I can imagine that there will rarely be an identity with more than one process, and probably never with a significant number of them.

I will not request changes here so that @tnotheis may have the final word on this.

Modules/Devices/src/Devices.ConsumerApi/Controllers/IdentitiesController.cs
```diff
@@ -76,6 +77,15 @@ public async Task<IActionResult> ApproveDeletionProcess([FromRoute] string id, C
...
     }
+
+    [HttpGet("Self/DeletionProcesses/{id}")]
+    [ProducesResponseType(StatusCodes.Status200OK)]
+    [ProducesError(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> GetDeletionProcess([FromRoute] string id, CancellationToken cancellationToken)
+    {
+        var response = await _mediator.Send(new GetDeletionProcessQuery { Id = id }, cancellationToken);
+        return Ok(response);
+    }
```
NikolaVetnic, line 88: Maybe you would also need integration tests for this but I suppose we should wait for decision of @tnotheis on this one.

NikolaVetnic (approved):

Modules/Devices/src/Devices.Application/Identities/Queries/GetDeletionProcess/Handler.cs
tnotheis: I agree that it is not ideal. But on the other hand it's as you said: there won't be that much data, so I don't think it is worth the time we spend on implementing something better. 
Also, we would have to pollute the Identity class as well as the IdentityRepository to achieve this.
So in short: we will leave it as it is.

tnotheis (approved):

PR #534, opened 2024-02-23
Consumer API: Cancel identity deletion process 

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.

# Description

The following changes were made in this pull request:

1. Methods `EnsureIdentityOwnsDevice` and `CancelDeletionProcessDuringGracePeriod` were added to `Identity` 
2. Domain errors `DeviceDoesNotBelongsToIdentity` and `DeletionProcessGracePeriodHasEnded`
3. `Cancelled` option in `DeletionProcessStatus` enum
4. Properties `CancelledAt` and `CancelledByDevice` has been added to `IdentityDeletionProcess` 
5. Database migrations were applied to accommodate the new columns
6. New command `CancelDeletionProcessDuringGracePeriodCommand`, response class `CancelDeletionProcessDuringGracePeriodResponse` and handler
7. New endpoint `PUT` on the route `Identities/Self/DeletionProcesses/{id}/Cancel` in the `IdentitiesController`
8. Tests were added for the handler and related domain logic

## Summary by CodeRabbit

- **New Features**
	- Introduced a feature to cancel the deletion process of device identities during a grace period.
- **Bug Fixes**
	- Enhanced error handling related to device ownership and deletion process grace periods.
- **Documentation**
	- Updated API documentation to include the new endpoint for canceling deletion processes.
- **Refactor**
	- Improved logic for identity deletion processes and ownership verification.
- **Tests**
	- Added comprehensive tests for the new cancellation feature and related functionalities.


Modules/Devices/src/Devices.Application/Identities/Commands/CancelDeletionProcessDuringGracePeriod/CancelDeletionProcessDuringGracePeriodCommand.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Backbone.Modules.Devices.Application.Identities.Commands.CancelDeletionProcessDuringGracePeriod;
+
+public class CancelDeletionProcessDuringGracePeriodCommand : IRequest<CancelDeletionProcessDuringGracePeriodResponse>
+{
+    public CancelDeletionProcessDuringGracePeriodCommand(string id)
+    {
+        CancelProcessId = id;
+    }
+
+    public string CancelProcessId { get; set; }
```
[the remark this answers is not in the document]
tnotheis: @HunorTotBagi Why did you call it `CancelProcessId` and not just `ProcessId`?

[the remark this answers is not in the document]
tnotheis: Regarding the comment of coderabbit: in theory it is of course better to make the setter private. But as all the other commands and queries have public setters as well, you can leave it public.

Modules/Devices/src/Devices.Application/Identities/Commands/CancelDeletionProcessDuringGracePeriod/Handler.cs
```diff
@@ -0,0 +1,33 @@
...
+        _userContext = userContext;
+    }
+
+    public async Task Handle(CancelDeletionProcessDuringGracePeriodCommand request, CancellationToken cancellationToken)
+    {
+        var identity = await _identitiesRepository.FindByAddress(_userContext.GetAddress(), cancellationToken, true) ?? throw new NotFoundException(nameof(Identity));
+        identity.CancelDeletionProcessDuringGracePeriod(identity.Devices[0].Id);
+
+        await _identitiesRepository.Update(identity, cancellationToken);
+    }
```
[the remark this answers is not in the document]
tnotheis: No, that's incorrect. We have the `IUserContext` to get the device the user is currently logged in with.

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -84,6 +88,12 @@ public IdentityDeletionProcess StartDeletionProcessAsOwner(DeviceId asDevice)
         return deletionProcess;
     }
 
+    private void EnsureIdentityOwnsDevice(DeviceId currentDeviceId)
+    {
+        if (Devices.All(device => device.Id != currentDeviceId))
```
NikolaVetnic, line 93: Wow, this is quite some mental gymnastics... It works though. Maybe it should be simplified, I'd let @tnotheis decide.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcessDuringGracePeriod/HandlerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+    public async Task Happy_path()
+    {
+        // Arrange
+        var activeIdentity = TestDataGenerator.CreateIdentityWithApprovedDeletionProcess(SystemTime.UtcNow.Date);
+        var activeDevice = activeIdentity.Devices[0];
+
+        var mockIdentitiesRepository = A.Fake<IIdentitiesRepository>();
+        var fakeUserContext = A.Fake<IUserContext>();
+
+        A.CallTo(() => mockIdentitiesRepository.FindByAddress(A<IdentityAddress>._, A<CancellationToken>._, A<bool>._))
```
NikolaVetnic, line 28: I would be more specific here, i.e. instead of `A<IdentityAddress>._` I would use `activeIdentity.Address`. Also, I think we agreed earlier that `CancellationToken.None` should be used instead of `A.CancellationToken>._`.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcessDuringGracePeriod/HandlerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            .Returns(activeIdentity);
+        A.CallTo(() => fakeUserContext.GetAddressOrNull()).Returns(activeIdentity.Address);
+        A.CallTo(() => fakeUserContext.GetDeviceId()).Returns(activeDevice.Id);
+
+        var handler = new Handler(mockIdentitiesRepository, fakeUserContext);
+
+        // Act
+        var command = new CancelDeletionProcessDuringGracePeriodCommand(activeDevice.Id);
```
NikolaVetnic, line 36: Only line 37 should be in `Act`, the command is created in the `Arrange` section.

NikolaVetnic (changes requested):

Modules/Devices/src/Devices.Application/Identities/Commands/CancelDeletionProcessDuringGracePeriod/CancelDeletionProcessDuringGracePeriodCommand.cs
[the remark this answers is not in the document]
HunorTotBagi: @tnotheis Nikola Vetnić suggested a name change for this property to be `DeletionProcessId` that I already applied before I saw this comment. If I change it to `ProcessId` as you suggested, there will be a discrepancy with `ApproveDeletionProcessCommand` where this property is already called `DeletionProcessId`. I think it would be nice that this property is called the same in both commands. With what naming should we proceed?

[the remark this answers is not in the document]
tnotheis: I like Nikola's approach better.

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
HunorTotBagi: At this point I resolved all other comments. Waiting for Timo's decision on this.

tnotheis: How about

```suggestion
        if (!Devices.Exists(device => device.Id == currentDeviceId))
```

Should be easier to understand. What do you think @NikolaVetnic?

NikolaVetnic: Absolutely.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -1,18 +1,18 @@
...
+using Backbone.Modules.Devices.Application.Identities.Commands.CancelDeletionProcess;
 using Backbone.Modules.Devices.Application.Infrastructure.Persistence.Repository;
 using Backbone.Modules.Devices.Domain.Entities.Identities;
 using Backbone.Tooling;
 using FakeItEasy;
 using FluentAssertions;
 using Xunit;
 using static Backbone.UnitTestTools.Data.TestDataGenerator;
-using Handler = Backbone.Modules.Devices.Application.Identities.Commands.CancelDeletionProcessDuringGracePeriod.Handler;
+using Handler = Backbone.Modules.Devices.Application.Identities.Commands.CancelDeletionProcess.Handler;
```
NikolaVetnic, line 12: Here's how I wrote the import:
```csharp
using Backbone.Modules.Devices.Application.Identities.Commands.RejectDeletionProcess;
```

I think this is the usual way we do it on the project.

NikolaVetnic (changes requested):

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
HunorTotBagi: My bad, `using Backbone.Modules.Devices.Application.Identities.Commands.CancelDeletionProcess;` is already on line 3. So `using Handler = Backbone.Modules.Devices.Application.Identities.Commands.CancelDeletionProcess.Handler;` is totally removed.

Modules/Devices/src/Devices.Application/Identities/Commands/CancelDeletionProcess/CancelDeletionProcessResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+using Backbone.Modules.Devices.Domain.Entities.Identities;
+
+namespace Backbone.Modules.Devices.Application.Identities.Commands.CancelDeletionProcess;
+public class CancelDeletionProcessResponse
+{
+    public CancelDeletionProcessResponse(IdentityDeletionProcess deletionProcess)
+    {
+        Id = deletionProcess.Id;
+        Status = deletionProcess.Status;
+        CancelledAt = deletionProcess.CancelledAt ?? throw new Exception($"The '{nameof(IdentityDeletionProcess.CancelledAt)}' property of the given deletion process must not be null.");
```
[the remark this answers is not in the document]
tnotheis: It's debatable. The `CancelledAt` property theoretically is nullable. But since the CancelDeletionProcessResponseCommand always sets it, it's never null when creating the response. 
I think I would go with a null suppression in this case.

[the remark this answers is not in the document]
NikolaVetnic: What should we do with `ApproveDeletionProcessResponse` then, it's the same there:
```csharp
        ApprovedAt = deletionProcess.ApprovedAt ?? throw new Exception($"The '{nameof(IdentityDeletionProcess.ApprovedAt)}' property of the given deletion process must not be null.");
        ApprovedByDevice = deletionProcess.ApprovedByDevice ?? throw new Exception($"The '{nameof(IdentityDeletionProcess.ApprovedByDevice)}' property of the given deletion process must not be null.");
```

I'm fine with either way, I just think we should be uniform about it.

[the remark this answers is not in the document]
HunorTotBagi: Also in `StartDeletionProcessAsOwnerResponse`
```csharp
   ApprovedAt = deletionProcess.ApprovedAt ?? throw new Exception($"The '{nameof(IdentityDeletionProcess.ApprovedAt)}' property of the given deletion process must not be null.");
   ApprovedByDevice = deletionProcess.ApprovedByDevice ?? throw new Exception($"The '{nameof(IdentityDeletionProcess.ApprovedByDevice)}' property of the given deletion process must not be null.");
```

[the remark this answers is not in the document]
tnotheis: Hmm. Okaaay, then let's keep throwing the exception. 😄

Modules/Devices/src/Devices.Domain/DomainErrors.cs
```diff
@@ -49,4 +49,14 @@ public static DomainError NoDeletionProcessWithRequiredStatusExists()
     {
         return new DomainError("error.platform.validation.device.noDeletionProcessWithRequiredStatusExists", "The deletion process does not have the correct status to perform this action.");
     }
+
+    public static DomainError DeviceDoesNotBelongsToIdentity()
```
tnotheis, line 53: Typo

Modules/Devices/src/Devices.Domain/DomainErrors.cs
```diff
@@ -49,4 +49,14 @@ public static DomainError NoDeletionProcessWithRequiredStatusExists()
...
+    {
+        return new DomainError("error.platform.validation.device.deviceDoesNotBelongsToIdentity", "This device does not belong to this identity.");
```
tnotheis, line 55: Same typo

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -84,6 +85,12 @@ public IdentityDeletionProcess StartDeletionProcessAsOwner(DeviceId asDevice)
         return deletionProcess;
     }
 
+    private void EnsureIdentityOwnsDevice(DeviceId currentDeviceId)
+    {
+        if (!Devices.Exists(device => device.Id == currentDeviceId))
+            throw new DomainException(DomainErrors.DeviceDoesNotBelongsToIdentity());
```
tnotheis, line 91: Why don't you just use `GenericDomainErrors.NotFound(nameof(Device))`?

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -165,12 +172,27 @@ public void DeletionGracePeriodReminder3Sent()
     {
         return DeletionProcesses.FirstOrDefault(x => x.Status == deletionProcessStatus);
     }
+
+    public IdentityDeletionProcess CancelDeletionProcess(DeviceId canceledByDeviceId)
+    {
+        var deletionProcess = GetDeletionProcessInStatus(DeletionProcessStatus.Approved) ?? throw new DomainException(DomainErrors.NoDeletionProcessWithRequiredStatusExists());
+
+        if (deletionProcess.GracePeriodEndsAt < SystemTime.UtcNow)
```
tnotheis, line 180: I would only check for the correct status. Once the grace period has expired, the deletion process will transition to the status `Deleting`.
In theory there is a short period of time were this transition didn't occur yet, even though the grace period is over. But we're okay with the deletion process being cancelled during this period.

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -165,12 +172,27 @@ public void DeletionGracePeriodReminder3Sent()
...
+            throw new DomainException(DomainErrors.DeletionProcessGracePeriodHasEnded((DateTime)deletionProcess.GracePeriodEndsAt));
+
+        deletionProcess.Cancel(Address, canceledByDeviceId);
+        TierId = TierIdBeforeDeletion;
```
tnotheis, line 184: You should also clear the `TierIdBeforeDeletion`

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcess.cs
```diff
@@ -53,6 +53,13 @@ private void Approve(DeviceId createdByDevice)
         GracePeriodEndsAt = SystemTime.UtcNow.AddDays(IdentityDeletionConfiguration.LengthOfGracePeriod);
     }
 
+    private void Cancel(DeviceId cancelledByDevice)
```
tnotheis, line 56: Private methods should always be placed below the public methods they belong to.

Also, methods in general should never be placed above the properties.

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcess.cs
```diff
@@ -126,4 +136,13 @@ public void Approve(IdentityAddress address, DeviceId approvedByDevice)
         Approve(approvedByDevice);
         _auditLog.Add(IdentityDeletionProcessAuditLogEntry.ProcessApproved(Id, address, approvedByDevice));
     }
+
+    public void Cancel(IdentityAddress address, DeviceId cancelledByDevice)
+    {
+        if (Status != DeletionProcessStatus.Approved)
+            throw new DomainException(DomainErrors.NoDeletionProcessWithRequiredStatusExists());
+
+        Cancel(cancelledByDevice);
```
tnotheis, line 145: Out of curiosity: why did you extract this into a separate method?

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,64 @@
...
+public class HandlerTests
+{
+    [Fact]
+    public async Task Happy_path()
+    {
+        // Arrange
+        var activeIdentity = TestDataGenerator.CreateIdentityWithApprovedDeletionProcess(SystemTime.UtcNow.Date);
+        var activeDevice = activeIdentity.Devices[0];
+
+        var mockIdentitiesRepository = A.Fake<IIdentitiesRepository>();
```
tnotheis, line 22: This is not used as a mock and therefore should not be called a mock.

BUT: your test is missing an assertion that makes sure the Identity and the DeletionProcess were updated in the database

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    [Fact]
+    public void Cannot_start_when_given_identity_does_not_exist()
+    {
+        // Arrange
+        var address = CreateRandomIdentityAddress();
+
+        var fakeUserContext = A.Fake<IUserContext>();
+        var fakeIdentitiesRepository = A.Fake<IIdentitiesRepository>();
```
tnotheis, line 48: This is a dummy, not a fake. And this also means that you should extract a `CreateHandler` method with an optional parameter for the repository.

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using Backbone.Modules.Devices.Domain.Entities.Identities;
+using Backbone.Tooling;
+using FluentAssertions;
+using Xunit;
+
+namespace Backbone.Modules.Devices.Domain.Tests.Identities;
+public class CancelDeletionProcessTests
+{
+    [Fact]
+    public void Should_cancel_deletion_process_during_grace_period()
```
tnotheis, line 11: We don't use `Should` in test names. Tests are supposed to be documentation. And if a documentation for a product says that it **should** do something, it sounds like "I wish it would, but sometimes it doesn't".

Modules/Devices/test/Devices.Domain.Tests/Identities/StartDeletionProcessAsOwnerTests.cs
```diff
@@ -42,17 +43,35 @@ public void Start_deletion_process()
         auditLogEntry.NewStatus.Should().Be(DeletionProcessStatus.Approved);
     }
 
+    [Fact]
+    public void Start_deletion_process_as_owner_from_another_device()
```
tnotheis, line 47: What does "another" device mean?

Modules/Devices/test/Devices.Domain.Tests/Identities/StartDeletionProcessAsOwnerTests.cs
```diff
@@ -42,17 +43,35 @@ public void Start_deletion_process()
...
+    {
+        // Arrange
+        var activeIdentity1 = CreateIdentity();
+        var activeIdentity2 = CreateIdentity();
+        var activeDevice1 = new Device(activeIdentity1);
+        var activeDevice2 = new Device(activeIdentity2);
```
tnotheis, line 53: How can there be multiple active Identities and multiple active Devices?
=> You should really improve the naming

Modules/Devices/test/Devices.Domain.Tests/Identities/StartDeletionProcessAsOwnerTests.cs
```diff
@@ -42,17 +43,35 @@ public void Start_deletion_process()
         auditLogEntry.NewStatus.Should().Be(DeletionProcessStatus.Approved);
     }
 
+    [Fact]
+    public void Start_deletion_process_as_owner_from_another_device()
+    {
+        // Arrange
+        var activeIdentity1 = CreateIdentity();
+        var activeIdentity2 = CreateIdentity();
```
tnotheis, line 51: Why not create the Device without the Identity?

Modules/Devices/test/Devices.Domain.Tests/Identities/StartDeletionProcessAsOwnerTests.cs
tnotheis, line 1: Hats off for remembering to write tests for the updated behavior of the `Start DeletionProcessAsOwner` method.

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        // Arrange
+        var currentDate = SystemTime.UtcNow.Date;
+        SystemTime.Set(currentDate);
```
tnotheis, line 15: I would use a hardcoded date. E.g.

```c#
DateTime.Parse("2020-01-01");
```

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        var activeIdentity = TestDataGenerator.CreateIdentity();
+        var activeDevice = new Device(activeIdentity);
+        activeIdentity.Devices.Add(activeDevice);
+
+        activeIdentity.StartDeletionProcessAsOwner(activeDevice.Id);
+        SystemTime.Set(currentDate.AddDays(5));
```
tnotheis, line 22: Same as above. Use a hardcoded date that is one day after the previous one.

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        // Arrange
+        var currentDate = SystemTime.UtcNow.Date;
+        SystemTime.Set(currentDate);
+
+        var activeIdentity = TestDataGenerator.CreateIdentity();
+        var activeDevice = new Device(activeIdentity);
+        activeIdentity.Devices.Add(activeDevice);
+
+        activeIdentity.StartDeletionProcessAsOwner(activeDevice.Id);
```
tnotheis, line 21: This method returns the started deletion process. Why don't you save this in a variable so that you can use it in the assertions?

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        SystemTime.Set(currentDate.AddDays(5));
+
+        // Act
+        activeIdentity.CancelDeletionProcess(activeDevice.Id);
+
+        // Assert
+        var result = activeIdentity.DeletionProcesses.Where(dp => dp.Status == DeletionProcessStatus.Cancelled).ToList();
```
tnotheis, line 28: Here you could use the deletion process you saved as described in my previous comment

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
tnotheis, line 1: Please update the second and third test according to my comments on the first one.

tnotheis (changes requested):

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcess.cs
HunorTotBagi: To be honest I took a look at the `Approve` method and followed the same pattern.

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcess.cs
HunorTotBagi: @NikolaVetnic already fixed up this file. In merge conflict I will adjust my changes in right places.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -21,47 +18,58 @@ public async Task Happy_path()
...
 
-        var fakeUserContext = A.Fake<IUserContext>();
-        var fakeIdentitiesRepository = A.Fake<IIdentitiesRepository>();
-
-        A.CallTo(() => fakeIdentitiesRepository.FindByAddress(
-                address,
-                CancellationToken.None,
-                A<bool>._))
-            .Returns<Identity>(null);
+        A.CallTo(() => dummyIdentitiesRepository.FindByAddress(address, CancellationToken.None, A<bool>._)).Returns<Identity>(null);
```
NikolaVetnic, line 55: I'd say this is a fake.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -21,47 +18,58 @@ public async Task Happy_path()
...
+
+        response.Status.Should().Be(DeletionProcessStatus.Cancelled);
     }
 
     [Fact]
     public void Cannot_start_when_given_identity_does_not_exist()
     {
         // Arrange
         var address = CreateRandomIdentityAddress();
+        var dummyUserContext = A.Fake<IUserContext>();
```
NikolaVetnic, line 52: If you are not setting up anything special with this user context then it should be created like this in the private `CreateHandler` method.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -21,47 +18,58 @@ public async Task Happy_path()
...
-
-        A.CallTo(() => fakeIdentitiesRepository.FindByAddress(
-                address,
-                CancellationToken.None,
-                A<bool>._))
-            .Returns<Identity>(null);
+        A.CallTo(() => dummyIdentitiesRepository.FindByAddress(address, CancellationToken.None, A<bool>._)).Returns<Identity>(null);
 
-        var handler = new Handler(fakeIdentitiesRepository, fakeUserContext);
+        var handler = CreateHandler(dummyUserContext);
```
NikolaVetnic, line 57: If you are not passing this `dummyIdentitiesRepository` to the `CreateHandler` I think it is not even used then.

NikolaVetnic (changes requested):

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+using static Backbone.UnitTestTools.Data.TestDataGenerator;
+
+namespace Backbone.Modules.Devices.Application.Tests.Tests.Identities.Commands.CancelDeletionProcess;
+public class HandlerTests
+{
+    [Fact]
+    public async Task Happy_path()
+    {
+        // Arrange
+        var activeIdentity = TestDataGenerator.CreateIdentityWithApprovedDeletionProcess(SystemTime.UtcNow.Date);
```
tnotheis, line 19: You should make the parameter of the `CreateIdentityWithApprovedDeletionProcess` method optional, and set the default value to `SystemTime.UtcNow`.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+    [Fact]
+    public void Cannot_start_when_given_identity_does_not_exist()
+    {
+        // Arrange
+        var address = CreateRandomIdentityAddress();
+        var fakeIdentitiesRepository = A.Fake<IIdentitiesRepository>();
+        var fakeUserContext = A.Fake<IUserContext>();
+
+        A.CallTo(() => fakeIdentitiesRepository.FindByAddress(address, CancellationToken.None, A<bool>._)).Returns<Identity>(null);
```
tnotheis, line 55: Man, ddin't you see the huge blue line that Visual Studio shows you below line of code?

```suggestion
        A.CallTo(() => fakeIdentitiesRepository.FindByAddress(address, CancellationToken.None, A<bool>._)).Returns<Identity?>(null);
```

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        var handler = CreateHandler(fakeUserContext);
```
tnotheis, line 57: This arrange block is a mess.
1. You initialize and set up the `fakeIdentitiesRepository`, but you never use it.
2. You initialize the `fakeUserContext`, but you never set it up (you can just move the initilazion into the `CreateHandler` method).

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        // Assert
+        acting.Should().ThrowAsync<NotFoundException>();
+    }
+
+    private static Handler CreateHandler(IUserContext userContext, IIdentitiesRepository? identityRepository = null)
+    {
+        return identityRepository == null
+            ? new Handler(A.Fake<IIdentitiesRepository>(), userContext)
+            : new Handler(identityRepository, userContext);
```
tnotheis, line 70: Why so complicated? All the other `CreateHandler` methods look different:

```suggestion
        identitiesRepository ??= A.Fake<IIdentitiesRepository>()
        return new Handler(identityRepository, userContext);
```

But keep in mind that according to the comment above, the `userContext` parameter should become optional, and the `identitiesRepository` should become mandatory.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        var handler = CreateHandler(fakeUserContext);
+
+        // Act
+        var acting = async () => await handler.Handle(new CancelDeletionProcessCommand(address), CancellationToken.None);
+
+        // Assert
+        acting.Should().ThrowAsync<NotFoundException>();
+    }
+
+    private static Handler CreateHandler(IUserContext userContext, IIdentitiesRepository? identityRepository = null)
```
tnotheis, line 66: Naming
```suggestion
    private static Handler CreateHandler(IUserContext userContext, IIdentitiesRepository? identitiesRepository = null)
```

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Backbone.Modules.Devices.Domain.Tests.Identities;
+public class CancelDeletionProcessTests
+{
+    [Fact]
+    public void Cancel_deletion_process()
+    {
+        // Arrange
+        var currentDate = DateTime.Parse("2020-01-01");
+        SystemTime.Set(currentDate);
```
tnotheis, line 15: You don't even assert on the `SystemTime` you set here. So why do you even set it?

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        identity.Devices.Add(device);
+
+        var deletionProcess = identity.StartDeletionProcessAsOwner(device.Id);
+        SystemTime.Set(DateTime.Parse("2020-01-02"));
+
+        // Act
+        identity.CancelDeletionProcess(deletionProcess.Id, device.Id);
+
+        // Assert
+        deletionProcess.Status.Should().Be(DeletionProcessStatus.Cancelled);
```
tnotheis, line 28: There are soooo many assertions missing. I bet you didn't develop this in a test driven way. 😠

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        // Arrange
+        var currentDate = DateTime.Parse("2020-01-01");
+        SystemTime.Set(currentDate);
+
+        var identity = TestDataGenerator.CreateIdentity();
+        var device = new Device(identity);
+        identity.Devices.Add(device);
+
+        var deletionProcess = identity.StartDeletionProcessAsOwner(device.Id);
```
tnotheis, line 21: You could create a `CreateIdentityWithApprovedDeletionProcess` method in the `TestDataGenerator` that does all this.

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -165,12 +172,31 @@ public void DeletionGracePeriodReminder3Sent()
...
+
+    public IdentityDeletionProcess CancelDeletionProcess(IdentityDeletionProcessId deletionProcessId, DeviceId canceledByDeviceId)
+    {
+        EnsureIdentityOwnsDevice(canceledByDeviceId);
+
+        var deletionProcess = DeletionProcesses.FirstOrDefault(x => x.Id == deletionProcessId) ??
+                              throw new DomainException(GenericDomainErrors.NotFound(nameof(IdentityDeletionProcess)));
+
+        if (Status == IdentityStatus.Active)
+            throw new DomainException(DomainErrors.DeletionProcessGracePeriodHasEnded((DateTime)deletionProcess.GracePeriodEndsAt));
```
tnotheis, line 184: I think we can even remove this check completely and instead rely on the `deletionProcess.Cancel method` to throw.
Remember to add a corresponding test for this.

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcess.cs
tnotheis: The `Approve´ method is a special case, because it is called multiple times. `Cancel` instead is only called once.

Modules/Devices/test/Devices.Domain.Tests/Identities/StartDeletionProcessAsOwnerTests.cs
```diff
@@ -42,17 +43,36 @@ public void Start_deletion_process()
...
+        var device = new Device(identity);
+
+        identity.Devices.Add(device);
+        identity.StartDeletionProcessAsOwner(device.Id);
+
+        // Act
+        var acting = () => identity.CancelDeletionProcess(identity.DeletionProcesses[0].Id, DeviceId.Parse("DVC"));
+
+        // Assert
+        acting.Should().Throw<DomainException>().Which.Code.Should().Be("error.platform.recordNotFound");
```
tnotheis, line 61: You should additionally assert that the message contains the name of the entity that wasn't found.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    [Fact]
+    public async Task Happy_path()
+    {
+        // Arrange
+        var activeIdentity = TestDataGenerator.CreateIdentityWithApprovedDeletionProcess(SystemTime.UtcNow.Date);
+        var activeDevice = activeIdentity.Devices[0];
+        var deletionProcess = activeIdentity.GetDeletionProcessInStatus(DeletionProcessStatus.Approved)!;
+
+        var fakeIdentitiesRepository = A.Fake<IIdentitiesRepository>();
```
tnotheis, line 23: This is used as a mock 😑

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        var command = new CancelDeletionProcessCommand(deletionProcess.Id);
+
+        // Act
+        var response = await handler.Handle(command, CancellationToken.None);
+
+        // Assert
+        A.CallTo(() => fakeIdentitiesRepository.Update(A<Identity>.That.Matches(i =>
+                i.Address == activeIdentity.Address
+                && i.Status == IdentityStatus.Active
+                && i.DeletionProcesses.Any(d => d.Id == deletionProcess.Id)), A<CancellationToken>._))
```
tnotheis, line 41: This assertion is weird. You just make sure that there is a deletion process with the given id. Thais doesn't have much effect. You should at least assert that the deletion process is in the correct status.

tnotheis (changes requested):

Modules/Devices/test/Devices.Domain.Tests/Identities/StartDeletionProcessAsOwnerTests.cs
```diff
@@ -42,17 +43,36 @@ public void Start_deletion_process()
...
+        // Arrange
+        SystemTime.Set(DateTime.Parse("2020-01-01"));
+        var identity = TestDataGenerator.CreateIdentity();
+        var device = new Device(identity);
+
+        identity.Devices.Add(device);
+        identity.StartDeletionProcessAsOwner(device.Id);
+
+        // Act
+        var acting = () => identity.CancelDeletionProcess(identity.DeletionProcesses[0].Id, DeviceId.Parse("DVC"));
```
tnotheis, line 58: Wtf. Why is there a `CancelDeletionProcess` test in this file??

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -165,12 +172,31 @@ public void DeletionGracePeriodReminder3Sent()
...
     }
+
+    public IdentityDeletionProcess CancelDeletionProcess(IdentityDeletionProcessId deletionProcessId, DeviceId canceledByDeviceId)
+    {
+        EnsureIdentityOwnsDevice(canceledByDeviceId);
+
+        var deletionProcess = DeletionProcesses.FirstOrDefault(x => x.Id == deletionProcessId) ??
+                              throw new DomainException(GenericDomainErrors.NotFound(nameof(IdentityDeletionProcess)));
+
+        if (Status == IdentityStatus.Active)
```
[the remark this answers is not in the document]
NikolaVetnic: @tnotheis I apologize.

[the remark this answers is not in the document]
HunorTotBagi: I got confused with your last comment Timo about status `Deleting`. I checked in the code that it doesn't exist yet, so i opted to solve this way. But now after thinking it makes sense to use `!= ToBeDeleted`.

Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelDeletionProcess/HandlerTests.cs
HunorTotBagi: Oh, damn, now I see how the `CreateHandler` methods look like in other tests. I should have taken a look into them. My bad. Also, I fixed which parameter is optional and which is mandatory.

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcess.cs
HunorTotBagi: I understand. I have removed the private `Cancel` method and put the logic inside the public `Cancel` method.

NikolaVetnic (changes requested):
Requesting changes so Hunor can re-request a review.

Modules/Devices/test/Devices.Application.Tests/TestDataGenerator.cs
```diff
@@ -35,16 +35,14 @@ public static Identity CreateIdentityWithOneDevice()
         return identity;
     }
 
-    public static Identity CreateIdentityWithApprovedDeletionProcess(DateTime approvalDate)
+    public static Identity CreateIdentityWithApprovedDeletionProcess(DateTime? approvalDate = null)
     {
-        var currentDateTime = SystemTime.UtcNow;
+        var dateTime = approvalDate ?? SystemTime.UtcNow;
```
tnotheis, line 40: Just use 

```
approvalDate ??= SystemTime.UtcNow;
```
just like in all other `Create*` methods 😞

Modules/Devices/src/Devices.Domain/DomainErrors.cs
```diff
@@ -50,13 +50,8 @@ public static DomainError NoDeletionProcessWithRequiredStatusExists()
         return new DomainError("error.platform.validation.device.noDeletionProcessWithRequiredStatusExists", "The deletion process does not have the correct status to perform this action.");
     }
 
-    public static DomainError DeviceDoesNotBelongToIdentity()
+    public static DomainError NoIdentityWithRequiredStatusExists()
```
tnotheis, line 53: Is this even used?

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
tnotheis: Mhhh... What about assertions on
* identity.TierId
* deletionProcess.CancelledAt
* deletionProcess.CancelledBy

?

It's not hard to find out which assertions you have to make. Just look at the production code. Or you just perform TDD, as you were supposed to. Then it is impossible to forget anything.

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelDeletionProcessTests.cs
tnotheis: The comment says "create a `CreateIdentityWithApprovedDeletionProcess` method in the **_TestDataGenerator_**

Modules/Devices/test/Devices.Domain.Tests/Identities/StartDeletionProcessAsOwnerTests.cs
tnotheis: Now you act twice. Look into other occurrences of this "pattern" or ask @NikolaVetnic

tnotheis (changes requested):
.

Modules/Devices/test/Devices.Application.Tests/TestDataGenerator.cs
HunorTotBagi: The solution that you are suggesting was my initial approach, but then in line 43 `SystemTime.Set(approvalDate)` says that it cannot convert from 'System.Date.Time?' to 'System.DateTime' becase `approvalDate` is nullable. So putting it in a new variable that's not nullable `var dateTime = approvalDate ?? SystemTime.UtcNow` seemed like a good solution.

tnotheis:
```c# 
SystemTime.Set(approvalDate.Value);
```

tnotheis (approved):

PR #546, opened 2024-03-04
IdentityDeletionJob: Automatic cancellation of an identity deletion process if no response within the approval period

NikolaDmitrasinovic:
# Readiness checklist

-   [x] I labeled the PR.

# Description
Implemented the automatic cancellation of an identity deletion process started by the support, but not approved by the owner of the identity within the approval time period.
 
- Implemented `CancelIdentityDeletionProcessWorker` with tests - this class makes the cancel identity deletion process `Hander` calls, receives IDs of the canceled processes and adds them to the log message.
- Implemented `Handler` with tests - this handler iterates through the identity repository, makes a list of identities with deletion processes that are past due approval, calls `CancelStaleDeletionProcess()` on those identities, publishes related integration event and returns a list of canceled processes IDs.
- Added  `CancelStaleDeletionProcess()` to `Identity` class with related tests - this new method uses `IdentityDeletionId` to to find the identity deletion process and calls `Cancel()` on that process. 
- Added `Cancel()` to `IdentityDeletionProcess` calss - this new method sets deletion process status to `DeletionProcessStatus.Canceled`, record the time of a change and creates audit log about the change.
- Added `ProcessCanceledAutomatically()` to `IdentityDeletionProcessAuditLogEntry` - this new method creates an audit log entry about canceled identity deletion process.
- Some related smaller changes are also included.


Modules/Devices/test/Devices.Application.Tests/Tests/Identities/Commands/CancelStaleIdentityDeletionProcesses/HandlerTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        // Act
+        var response = await handler.Handle(new CancelStaleIdentityDeletionProcessesCommand(), CancellationToken.None);
+
+        // Assert
+        response.Should().HaveCount(1);
+        response.First().Should().Be(identityWithStaleDeletionProcess.DeletionProcesses[0].Id);
+    }
+
+    [Fact]
+    public async Task Publish_IntegrationEvent_for_canceled_deletion_process()
```
tnotheis, line 51: I think I would create two stale processes and assert that the Publish method is called once for each identity.

Modules/Synchronization/src/Synchronization.Application/IntegrationEvents/Incoming/IdentityDeletionProcessStatusChanged/IdentityDeletionProcessStatusChangedIntegrationEvent.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public IdentityDeletionProcessStatusChangedIntegrationEvent(string identityAddress, string deletionProcessId)
+        : base($"{identityAddress}/IdentityDeletionProcessStatusChanged/{deletionProcessId}")
+    {
+        Address = identityAddress;
+        DeletionProcessId = deletionProcessId;
+    }
+
+    public string Address { get; }
+    public object DeletionProcessId { get; }
```
tnotheis, line 14: Why object?

Modules/Synchronization/test/Synchronization.Application.Tests/Tests/IntegrationEvents/IdentityDeletionProcessStatusChangedIntegrationEventHandlerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+using Backbone.Modules.Synchronization.Domain.Entities.Sync;
+using FakeItEasy;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace Backbone.Modules.Synchronization.Application.Tests.Tests.IntegrationEvents;
+public class IdentityDeletionProcessStatusChangedIntegrationEventHandlerTests
+{
+    [Fact]
+    public async Task Creates_an_external_event()
```
tnotheis, line 15: The test name promises to test that an external event is created. But what it actually does it is tests if an integration event is published.

You should actually do both (=> write 2 tests)

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelStaleDeletionProcessTests.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Backbone.Modules.Devices.Domain.Entities.Identities;
+using Backbone.Tooling;
+using Backbone.UnitTestTools.Extensions;
+using FluentAssertions;
+using Xunit;
+
+namespace Backbone.Modules.Devices.Domain.Tests.Identities;
+public class CancelStaleDeletionProcessTests
```
tnotheis, line 8: There are tests missing for the following cases:
* The deletion process id does not exist
* The deletion process is not in status "WaitingForApproval"

tnotheis (changes requested):

Modules/Synchronization/src/Synchronization.Application/IntegrationEvents/Incoming/IdentityDeletionProcessStatusChanged/IdentityDeletionProcessStatusChangedIntegrationEvent.cs
NikolaDmitrasinovic: Doing things in a TDD way and using VS to create methods, props... automatically, I should have changed all of these during the actual implementation. No excuse how I missed this one...

tnotheis: That's what I expected. And to be honest: this has been happening to me as well. So no worries. ;) I'm happy to see that you're actually doing TDD.

Modules/Devices/test/Devices.Domain.Tests/Identities/CancelStaleDeletionProcessTests.cs
NikolaDmitrasinovic: Reminder: we agreed that there is no need for id test as id is not used in the method

On another note, if you remember, the tested method used to return Result and also throw the exception. I think I found a really nice way to solve that. Test is also updated accordingly.

tnotheis: In your solution you just return an error message. But I think I remember saying that you should return a DomainError so that we have an error code.

NikolaDmitrasinovic: Yeah, I misunderstood you when we talked about returning Result and throwing an exception in a same method. It's fixed.

tnotheis (approved):

tnotheis (approved):

PR #547, opened 2024-03-05
Admin UI: Initialize flutter setup

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.


tnotheis:
I close it until it is fully implemented, to keep the PR overview clean.

tnotheis:
Yeah, I thought about that. But I would at least like to have the pipe setup completed, and I won't manage to do this today and maybe not even tomorrow

tnotheis:
I didn't say anything about WHICH pipeline. ;) 
But at least the build and test pipelines should exist.

jkoenig134:
ah, right.. good catch. should I add these?

tnotheis:
Maybe you can give me the commands. Then I will add the pipelines once I find time.

jkoenig134:
```
name: Test

on:
  push:

jobs:
  run-checks:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: subosito/flutter-action@v2.8.0
      - run: flutter pub get
      - run: flutter analyze
      - run: dart format . --set-exit-if-changed -l 150
```

AdminUi/.gitignore
```diff
@@ -0,0 +1,9 @@
+pubspec_overrides.yaml
+.dart_tool
+build
+.flutter-plugins
+.flutter-plugins-dependencies
+.DS_Store
+
+# config.json for connecting to the admin api
+config.json
```
tnotheis, line 9: @jkoenig134 Shouldn't we commit this with the connection to the local server?

tnotheis (approved):
@jkoenig134 I skimmed the new files and I think I'm okay with it. But you should give your final Go before we merge. (and maybe we should make a real PR out of this draft 😉)

AdminUi/.gitignore
tnotheis: With "this" I mean the `config.json` file

tnotheis: Sounds overcomplicated. Why not just committing the config.json file?
But if that's the way to go in flutter, I'm okay with it.

jkoenig134: As mac users will have a different config I always have to make sure to not commit it.

tnotheis:
We will add the Dockerfile and the corresponding build and publish pipelines in a separate PR.

tnotheis (approved):

PR #565, opened 2024-03-18
Technical ID for RecipientInformation

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

It seems it is not trivial to change the order of the columns in Postgres (https://wiki.postgresql.org/wiki/Alter_column_position). Hence, the newly added Id column is the last one.

It also seems that the EF Core method `AddColumn` does not support specifying a position for creation.


Dannyps:
> Did you test what's happening when the migration is applied on a database that has existing RecipientInformation?

That's something I'm currently doing.

> there now is the possibility of collisions, if multiple messages are sent in parallel.

I don't see how such could happen, considering the PK is generated by the database and both SQL Server and Postgres are ACID-compliant.

> Afaik there are different strategies of how to generate the auto incrementing primary keys. 

Regardless of the point above, I'll look into this matter.

tnotheis:
> I don't see how such could happen, considering the PK is generated by the database and both SQL Server and Postgres are ACID-compliant.

You are totally right. I wrote this comment seconds before I had to leave. So ignore that. ;)

tnotheis:
> The PK is successfully moved to the Id column and values are automatically generated for all pre-existing rows

Is it just me or is the fact that the Id column is the last one bugging you as well? 😆 (no need to change it, even though I feel the urge to 😉)

tnotheis (approved):

Dannyps:
Regarding the different strategies of how to generate the auto incrementing primary keys, the only things I found were changing the initial value and the step.

PR #570, opened 2024-03-21
Backbone-specific addresses

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.

See 9d532e3baf4f0db2e85b4d00a2fb99d5f18cd41a for important info before merging.


tnotheis:
@Dannyps The definition of "refactoring" is that it doesn't change behavior.
This changes in this PR is the complete opposite. Nothing will work anymore after we merged this.

Dannyps:
Quite fair. I wasn't looking at the bigger picture!

tnotheis:
@Dannyps There is a conflict in this PR. You should resolve it rather soon than later, as it will only become worse. ;)

AdminApi/src/AdminApi/appsettings.json
```diff
@@ -23,6 +23,7 @@
         "Devices": {
             "Application": {
                 "AddressPrefix": "id1",
```
tnotheis, line 25: We don't need the AddressPrefix anymore.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -47,29 +48,39 @@ public static bool IsValid(string? stringValue)
...
+            var realmBytes = Encoding.UTF8.GetBytes(realm);
+            var correctChecksum = CalculateChecksum(realmBytes, hashedPublicKey);
 
-        var checksumIsValid = givenChecksum.SequenceEqual(correctChecksum);
+            var checksumIsValid = givenChecksum.SequenceEqual(correctChecksum);
 
-        return lengthIsValid && checksumIsValid;
+            return lengthIsValid && checksumIsValid;
+        }
+        catch (IndexOutOfRangeException _)
```
tnotheis, line 69: 1. Remove the `_` to get rid of the compiler warning
2. Instead of provoking the exception, we should do length checks beforehand. In general you should try to avoid exceptions whenever possible

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -47,29 +48,39 @@ public static bool IsValid(string? stringValue)
...
+        const string pattern = @"^did\:web\:(.+?)\:dids\:(.{3})(.+)$";
+        var matches = Regex.Matches(stringValue, pattern, RegexOptions.IgnoreCase);
 
-        var concatenation = Base58.Bitcoin.Decode(stringValue.AsSpan(3)).ToArray();
-        var hashedPublicKey = concatenation[..20];
-        var givenChecksum = concatenation[20..];
+        try
+        {
+            var realm = matches.First().Groups[2].Value;
+            var concatenation = Base58.Bitcoin.Decode(matches.First().Groups[3].Value).ToArray();
```
tnotheis, line 57: The two `First`s will throw unhandled exceptions

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -47,29 +48,39 @@ public static bool IsValid(string? stringValue)
...
-        var realm = stringValue[..3];
+        const string pattern = @"^did\:web\:(.+?)\:dids\:(.{3})(.+)$";
+        var matches = Regex.Matches(stringValue, pattern, RegexOptions.IgnoreCase);
 
-        var concatenation = Base58.Bitcoin.Decode(stringValue.AsSpan(3)).ToArray();
-        var hashedPublicKey = concatenation[..20];
-        var givenChecksum = concatenation[20..];
+        try
+        {
+            var realm = matches.First().Groups[2].Value;
```
tnotheis, line 56: There will be no more `realm`. So you can remove it from the validation and generation

BuildingBlocks/src/UnitTestTools/Data/TestDataGenerator.cs
```diff
@@ -14,7 +14,7 @@ public static string GenerateString(int resultLength, char[]? chars = null)
 
     public static IdentityAddress CreateRandomIdentityAddress()
     {
-        return IdentityAddress.Create(CreateRandomBytes(), "id1");
+        return IdentityAddress.Create(CreateRandomBytes(), "id1", "url");
```
tnotheis, line 17: I would make more realistic examples:

```suggestion
        return IdentityAddress.Create(CreateRandomBytes(), "id1", "prod.enmeshed.eu");
```

BuildingBlocks/test/DevelopmentKit.Identity.Tests/IdentityAddressTests.cs
tnotheis, line 1: The test data in this file will need to be updated after you remove the realm.

Modules/Devices/src/Devices.Application/ApplicationOptions.cs
```diff
@@ -11,6 +11,11 @@ public class ApplicationOptions
 
     [Required]
     public PaginationOptions Pagination { get; set; } = new();
+
+    [Required]
+    [MinLength(3)]
+    [MaxLength(45)]
```
tnotheis, line 17: Why 45? With this we will never reach the max characters of 100 - or do we?

Modules/Devices/src/Devices.Application/Users/Commands/SeedTestUsers/Handler.cs
```diff
@@ -26,7 +26,7 @@ public async Task Handle(SeedTestUsersCommand request, CancellationToken cancell
         var basicTier = await _tiersRepository.FindBasicTier(cancellationToken);
 
         var user = new ApplicationUser(new Device(new Identity("test",
-            IdentityAddress.Create([1, 1, 1, 1, 1], "id1"),
+            IdentityAddress.Create([1, 1, 1, 1, 1], "id1", "url"),
```
tnotheis, line 29: You should use the configured InstanceUrl

Modules/Devices/test/Devices.Domain.Tests/Identities/ApproveDeletionProcessTests.cs
```diff
@@ -96,7 +96,7 @@ private static void AssertAuditLogEntryWasCreated(IdentityDeletionProcess deleti
 
     private static Identity CreateIdentity()
     {
-        var address = IdentityAddress.Create([], "id1");
+        var address = IdentityAddress.Create([], "id1", "url");
```
tnotheis, line 99: Use `prod.enmeshed.eu`. Same for all other places where you use `"url"`

Modules/Quotas/src/Quotas.Infrastructure.Database.SqlServer/Migrations/20240321144435_IdentityAddress100.cs
tnotheis, line 1: Why is there a second sql server migration?

tnotheis (changes requested):

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
Dannyps: We use this function: https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.matches?view=net-8.0#system-text-regularexpressions-regex-matches(system-string-system-string-system-text-regularexpressions-regexoptions-system-timespan)

There does not seem to be a way to avoid the exceptions when the input does not match the expected format. The type I put in the catch block in wrong, granted, but I don't see a simple way to avoid it.

Modules/Devices/src/Devices.Application/ApplicationOptions.cs
Dannyps: We do! We can go with 48 now that we removed the realm, leaving a margin of two chars (or we can go with 50 with no margin).

tnotheis:
@Dannyps Remember to update the target branch.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
tnotheis: There are 3 exceptions documented:
1. `ArgumentException`: This one is thrown when the given regular expression is invalid. But if the regular expression actually IS invalid, then we will notice this pretty quickly, as this logic is used in pretty much every test.
2. `ArgumentNullException`: This one is thrown when either `input` or `pattern` is `null`. We can simply check `input` for `null` before we check the regex. And for `pattern`, the same arguments as in (1) apply.
3. `ArgumentOutOfRangeException`: Here the same arguments as in (1) apply.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -45,37 +47,41 @@ public static bool IsValid(string? stringValue)
...
 
-        var realm = stringValue[..3];
-
-        var concatenation = Base58.Bitcoin.Decode(stringValue.AsSpan(3)).ToArray();
-        var hashedPublicKey = concatenation[..20];
-        var givenChecksum = concatenation[20..];
+        const string pattern = @"^did\:e\:(.+?)\:dids\:(.+)(.{2})$";
+        try
+        {
+            var matches = Regex.Matches(stringValue, pattern, RegexOptions.IgnoreCase).First().Groups;
```
tnotheis, line 53: Hm, if there is no match, then `First` will throw an exception. So you should make sure that there is a match before getting it.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -45,37 +47,41 @@ public static bool IsValid(string? stringValue)
...
 
-        var realmBytes = Encoding.UTF8.GetBytes(realm);
-        var correctChecksum = CalculateChecksum(realmBytes, hashedPublicKey);
+            var givenChecksum = matches[3].Value;
```
tnotheis, line 55: If there is no checksum, `matches[3]` will throw an exception...

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -45,37 +47,41 @@ public static bool IsValid(string? stringValue)
...
+            var calculatedChecksum = CalculateChecksum(stringValue[..^2]);
```
tnotheis, line 56: If `stringValue` is less than 2 characters long, `stringValue[..^2]` will throw an `ArgumentOutOfRangeException`.

In general, this `VIsValid` function barely does any validation.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -45,37 +47,41 @@ public static bool IsValid(string? stringValue)
 
         var lengthIsValid = stringValue.Length <= MAX_LENGTH;
 
-        var realm = stringValue[..3];
-
-        var concatenation = Base58.Bitcoin.Decode(stringValue.AsSpan(3)).ToArray();
-        var hashedPublicKey = concatenation[..20];
-        var givenChecksum = concatenation[20..];
+        const string pattern = @"^did\:e\:(.+?)\:dids\:(.+)(.{2})$";
```
tnotheis, line 50:
```regex
^did\:e\:(.+?)\:dids\:(.+)(.{2})$
```

The `(.+?)` part is too open. We should 
* limit the number of valid characters to characters that are valid in a URL
* require a URL format

Modules/Challenges/src/Challenges.Infrastructure.Database.Postgres/Migrations/20240321084302_IdentityAddress80.cs
```diff
@@ -0,0 +1,50 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
+                name: "CreatedBy",
+                schema: "Challenges",
+                table: "Challenges",
+                type: "character(80)",
+                unicode: false,
+                fixedLength: true,
+                maxLength: 100,
```
tnotheis, line 20: Weird. The type is charachter(80), but the maxLength is 100.

Modules/Challenges/src/Challenges.Infrastructure.Database.SqlServer/Migrations/20240321143836_IdentityAddress80.cs
```diff
@@ -0,0 +1,50 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AlterColumn<string>(
+                name: "CreatedBy",
+                schema: "Challenges",
+                table: "Challenges",
+                type: "char(80)",
+                unicode: false,
+                fixedLength: true,
+                maxLength: 100,
```
tnotheis, line 20: See above. Check in every migration please.

tnotheis (changes requested):

tnotheis (changes requested):
Some migrations are called "Identity80". Can you just rename them to "IdentityAddress80" (remember to change the occurrences in code as well)

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
timnot90: > we're already limiting the length of the whole address. Wouldn't such a check be redundant?

I should have said the "set" of valid characters, instead of the "number" of valid characters.

tnotheis: > we're already limiting the length of the whole address. Wouldn't such a check be redundant?

I should have said the "set" of valid characters, instead of the "number" of valid characters.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -47,22 +47,19 @@ public static bool IsValid(string? stringValue)
 
         var lengthIsValid = stringValue.Length <= MAX_LENGTH;
 
-        const string pattern = @"^did\:e\:(.+?)\:dids\:(.+)(.{2})$";
-        try
-        {
-            var matches = Regex.Matches(stringValue, pattern, RegexOptions.IgnoreCase).First().Groups;
+        const string pattern = @"^did\:e\:((?:[a-zA-Z]+\.)+[a-zA-Z]+)\:dids\:(.+)(.{2})$";
```
tnotheis, line 50: Actually, we only want lowercase characters.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
tnotheis: This is still not fixed...

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -45,37 +47,38 @@ public static bool IsValid(string? stringValue)
...
 
-        var concatenation = Base58.Bitcoin.Decode(stringValue.AsSpan(3)).ToArray();
-        var hashedPublicKey = concatenation[..20];
-        var givenChecksum = concatenation[20..];
+        if (matches is null) return false;
 
-        var realmBytes = Encoding.UTF8.GetBytes(realm);
-        var correctChecksum = CalculateChecksum(realmBytes, hashedPublicKey);
+        var givenChecksum = matches[3].Value;
+        var calculatedChecksum = CalculateChecksum(stringValue[..^2]);
```
tnotheis, line 57: You should extract the 2 into a `CHECKSUM_LENGTH` constant and use it everywhere you use currently use the checksum length.

tnotheis, line 57: If you had a capturing group for the whole backbone specific part, you could use this group instead of having to get a substring.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
tnotheis: Well, currently it does, because you never check if the regex validated successfully.
But you're right, as soon as you do this correctly, it will not be necessary to check for the array length.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -45,37 +47,38 @@ public static bool IsValid(string? stringValue)
 
         var lengthIsValid = stringValue.Length <= MAX_LENGTH;
 
-        var realm = stringValue[..3];
+        const string pattern = @"^did\:e\:((?:[a-zA-Z]+\.)+[a-zA-Z]+)\:dids\:(.+)(.{2})$";
```
tnotheis, line 50: This regex will be used quite often. So it probably makes sense to convert it into a compiled one. ReSharper should suggest you to convert it for you, and it will give you something like this:

```cs
    [GeneratedRegex(@"^did\:e\:((?:[a-zA-Z]+\.)+[a-zA-Z]+)\:dids\:(.+)(.{2})$", RegexOptions.IgnoreCase, "de-DE")]
    private static partial Regex AddressValidationRegex();
```

which you can then use as follows:

```cs
var matches = AddressValidationRegex().Matches(stringValue).First().Groups;
```

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -45,37 +47,38 @@ public static bool IsValid(string? stringValue)
...
+
+        var matches = Regex.Matches(stringValue, pattern, RegexOptions.IgnoreCase).First().Groups;
```
tnotheis, line 52: `matches` is not a good name for something that contains the `Groups` of the first match.

tnotheis (changes requested):

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
Dannyps: This turns `public record IdentityAddress` into a `public partial record IdentityAddress`. Is that acceptable?

tnotheis: Of course. Why not?

Dannyps: For no specific reason, I'm just not sure what ramifications that may have.

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -116,5 +124,8 @@ public override object ConvertFrom(ITypeDescriptorContext? context, CultureInfo?
         return ParseUnsafe(stringValue);
     }
 
+    [GeneratedRegex(@"^(?<addressWithoutChecksum>did\:e\:(?<instanceUrl>(?:[a-z0-9]+\.)+[a-z]{2,})\:dids\:(?<identitySpecificPart>[0-9abcdef]{20}))(?<checksum>[0-9abcdef]{2})$")]
```
tnotheis, line 127: Use the `CHECKSUM_LENGTH`

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
```diff
@@ -43,39 +44,46 @@ public static bool IsValid(string? stringValue)
...
+        var matchGroups = matches.First().Groups;
 
-        var checksumIsValid = givenChecksum.SequenceEqual(correctChecksum);
+        if (!matchGroups.TryGetValue("checksum", out var givenChecksum))
+            return false;
+
+        if (!matchGroups.TryGetValue("addressWithoutChecksum", out var addressWithoutChecksum))
+            return false;
+
+        var calculatedChecksum = CalculateChecksum(addressWithoutChecksum.Value);
```
tnotheis, line 62:
```suggestion
        var expectedChecksum = CalculateChecksum(addressWithoutChecksum.Value);
```

tnotheis (changes requested):

BuildingBlocks/src/DevelopmentKit.Identity/ValueObjects/IdentityAddress.cs
Dannyps: I had already tried that, but failed. I'm unsure that will be feasible:
image

tnotheis (approved):

tnotheis (approved):

PR #595, opened 2024-04-08
Set default schemas for all DbContexts

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

As in the future we will probably run all migrations with a single user, we cannot rely on the default schema of the users anymore (up to now we always ran the migrations of the challenges module with the challenges user, the migrations of the devices module with the devices user, and so on).
With this PR I set the default schema for all db contexts to the corresponding schemas (ChallengesDbContext => "Challenges" schema, ...). 
The generated migrations' "Up" methods don't really do anything. But EF Core doesn't know that the default schemas for the users were the same as we now configured explicitly.


Modules/Devices/src/Devices.Infrastructure.Database.SqlServer/Migrations/20240408115400_SetDefaultSchema.cs
```diff
@@ -0,0 +1,191 @@
...
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.EnsureSchema(
+                name: "Devices");
+
+            migrationBuilder.RenameTable(
+                name: "Tiers",
+                newName: "Tiers",
+                newSchema: "Devices");
```
Dannyps, line 19: What's happening here? Do these things make sense?
It seems EF Core understands that the schema has changed, but it has not, actually.

tnotheis: See PR description.

Dannyps: I did. 

> The generated migrations' "Up" methods don't really do anything

Why are they needed then?

tnotheis: Well, imagine in the future there is a new database, and there the users don't have a default schema configured. When we now execute all migrations, the first ones will create the tables in the `dbo`/`postgres` schema, because that's the default default schema - if you know what I mean.

Though as you can see I now deleted the migration in the Admin API project. So there will be the same problem. But I didn't want to waste time figuring out what's necessary to move the views to a different schema (in postgres that is straightforward, but in SQL Server it's not). And since we will soon combine all the migrations anyway, I decided to ignore this potential problem. Currently there is no short-term plan to have a new database anyway.

Dannyps: So you're saying that the code I highlighted takes care of moving the table to the correct schema? Will that work well with references to other tables?

Why don't we simply change the migrations where tables are created and specify the correct schema there? I realise changing old migrations is not good practice, but I don't see the issue with it in this case.

tnotheis: Tbh, I wanted to avoid the work necessary to do this. 
But I just tried to run the migrations in a new database, and they fail. So I will do exactly what you suggested.

tnotheis: Done. You can have another look. I also wouldn't mind if you run the code on your machine to see if it works.

Dannyps:
There are several `b.ToTable("<table>", (string)null);` occurrences in the migrations. I wonder if those are ok or should be changed.

Dannyps (approved):

jkoenig134 (approved):

PR #596, opened 2024-04-08
Consumer API SDK: BaseAddress has Backslashes

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/BuildingBlocks/src/BuildingBlocks.SDK/UriUtils.cs
+++ b/BuildingBlocks/src/BuildingBlocks.SDK/UriUtils.cs
@@ -2,6 +2,5 @@
 
 public class UriUtils
 {
-    public static Uri CreateUriFromParts(params string[] parts)
-        => new Uri(parts.Aggregate((a, b) => $"{a.TrimEnd('/')}/{b.TrimStart('/')}"));
+    public static Uri CreateUriFromParts(params string[] parts) => new(parts.Aggregate((a, b) => $"{a.TrimEnd('/')}/{b.TrimStart('/')}"));
 }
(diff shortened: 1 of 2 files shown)


BuildingBlocks/src/BuildingBlocks.SDK/UriUtils.cs
```diff
@@ -2,6 +2,5 @@
 
 public class UriUtils
 {
-    public static Uri CreateUriFromParts(params string[] parts)
-        => new Uri(parts.Aggregate((a, b) => $"{a.TrimEnd('/')}/{b.TrimStart('/')}"));
+    public static Uri CreateUriFromParts(params string[] parts) => new(parts.Aggregate((a, b) => $"{a.TrimEnd('/')}/{b.TrimStart('/')}"));
```
tnotheis, line 5: Should be extracted into a method. Currently it's very hard to read and understand

tnotheis (changes requested):
A few comments about the title of the PR:
1. This doesn't apply to the Consumer API. It's just the SDK.
As it currently is it would show up in the bugs in the Consumer API changelog. And that's just not true.
2. Bugs should always be formulated like you would formulate a bug report. This means: describe what the user sees.

BuildingBlocks/src/BuildingBlocks.SDK/UriUtils.cs
MH321Productions: Would it be better like this?
```c#
var path = parts.Aggregate((a, b) => string.Format("{0}/{1}", a.TrimEnd('/'), b.TrimStart('/')));
```

Or with the interpolated string from earlier (bc ReSharper recommends me to use it)

tnotheis: Hmm, I'm okay with both.

PR #608, opened 2024-04-18
Consumer API: Reject a Relationship reactivation request

NikolaDmitrasinovic:
# Readiness checklist

-   [x] I labeled the PR.

# Description
Identity can reject request to reactivate previously terminated relationship.
Changes made:
* Implemented `RejectReactivation()` - new method in `Relationship` class creates new audit log entry and leaves relationship in `Terminatd` status if all check are passed.
Check: Ensures that request exists and it is addressed to self.
* Added new DomainEvent - `RelationshipReactivationCompleted` 
* Added new endpoint: PUT /Relationships/{id}/Reactivate/Reject


tnotheis (changes requested):
Before I review this, can you please look through the comments in the Acceptance-PR and fix those that apply to the Rejection here as well?

Modules/Relationships/src/Relationships.Application/Relationships/Commands/RejectRelationshipReactivation/Handler.cs
```diff
@@ -0,0 +1,41 @@
...
+        _activeIdentity = userContext.GetAddress();
+        _activeDevice = userContext.GetDeviceId();
+    }
+
+    public async Task<RejectRelationshipReactivationResponse> Handle(RejectRelationshipReactivationCommand request, CancellationToken cancellationToken)
+    {
+        var relationshipId = RelationshipId.Parse(request.RelationshipId);
+
+        var relationship = await _relationshipsRepository.FindRelationship(relationshipId, _activeIdentity, cancellationToken, track: true) ??
+                           throw new NotFoundException(nameof(Relationship));
```
tnotheis, line 31: `FindRelationships` cannot return null, can it?

Modules/Relationships/src/Relationships.Domain/DomainErrors.cs
```diff
@@ -52,4 +52,10 @@ public static DomainError CannotRequestReactivationWhenThereIsAnOpenReactivation
         return new DomainError("error.platform.validation.relationshipRequest.cannotRequestReactivationWhenThereIsAnOpenReactivationRequest",
             $"You cannot request reactivation when there is an open reactivation request.");
     }
+
+    public static DomainError NoRejectableReactivationRequestExists()
+    {
+        return new DomainError("error.platform.validation.relationshipRequest.noRejectableReactivationRequestExists",
+            "There is no pending reactivation request or you are not allowed to revoke it. A reactivation request can only be rejected by the identity that did not requested it.");
```
tnotheis, line 59:
```suggestion
            "There is no pending reactivation request or you are not allowed to revoke it. A reactivation request can only be rejected by the identity that did not request it.");
```

Same for 
* NoRevocableReactivationRequestExists
* NoAcceptableReactivationRequestExists (other PR)

Modules/Relationships/src/Relationships.Domain/DomainEvents/Outgoing/RelationshipReactivationCompletedDomainEvent.cs
```diff
@@ -6,9 +6,10 @@ namespace Backbone.Modules.Relationships.Domain.DomainEvents.Outgoing;
 public class RelationshipReactivationCompletedDomainEvent : DomainEvent
 {
     public RelationshipReactivationCompletedDomainEvent(Relationship relationship, IdentityAddress peer)
+        : base($"{relationship.Id}/ReactivationCompleted/{relationship.AuditLog.Last().CreatedAt}")
```
tnotheis, line 9: OrderBy

Modules/Synchronization/src/Synchronization.Domain/DomainEvents/Incoming/RelationshipReactivationCompleted/RelationshipReactivationCompletedDomainEvent.cs
```diff
@@ -3,7 +3,7 @@
 namespace Backbone.Modules.Synchronization.Domain.DomainEvents.Incoming.RelationshipReactivationCompleted;
 public class RelationshipReactivationCompletedDomainEvent : DomainEvent
 {
-    public RelationshipReactivationCompletedDomainEvent(string relationshipId, string peer) : base($"{relationshipId}/Reactivation/Completed")
+    public RelationshipReactivationCompletedDomainEvent(string relationshipId, string peer) : base($"{relationshipId}/ReactivationCompleted")
```
tnotheis, line 6: No need for the id in case of incoming events

```suggestion
    public RelationshipReactivationCompletedDomainEvent(string relationshipId, string peer)
```

tnotheis (changes requested):

Modules/Relationships/src/Relationships.Application/Relationships/Commands/RejectRelationshipReactivation/Handler.cs
NikolaDmitrasinovic: now when I look at it... `FindRelationship` calls on `FirstWithId` and that method will throw if null, so no, it can't.

Modules/Relationships/src/Relationships.Domain/DomainErrors.cs
NikolaDmitrasinovic: I was debating myself on that one and went with `requested` in the end...

Modules/Relationships/src/Relationships.Domain/DomainEvents/Outgoing/RelationshipReactivationCompletedDomainEvent.cs
NikolaDmitrasinovic: good thing I mocked Nikola V' for not ordering his logs...

Modules/Relationships/src/Relationships.Domain/DomainErrors.cs
tnotheis: But "did not requested" doesn't make sense, does it? 😲

tnotheis: You should fix the error message for NoRevocableReactivationRequestExists as well (as mentioned in my initial comment)

NikolaDmitrasinovic: Somehow my brain read Rejectable instead of Revocable....

tnotheis (approved):

PR #610, opened 2024-04-19
Usage of Consumer API SDK for running challenges integration tests

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.


BuildingBlocks/src/BuildingBlocks.SDK/Endpoints/Common/EndpointClient.cs
```diff
@@ -82,9 +82,19 @@ private async Task<ApiResponse<T>> Execute<T>(HttpRequestMessage request)
             responseContent = new MemoryStream(Encoding.UTF8.GetBytes(EMPTY_RESULT));
         }
 
-        var deserializedResponseContent = JsonSerializer.Deserialize<ApiResponse<T>>(responseContent, _jsonSerializerOptions)!;
+        var deserializedResponseContent = statusCode is HttpStatusCode.OK or HttpStatusCode.Created ? JsonSerializer.Deserialize<ApiResponse<T>>(responseContent, _jsonSerializerOptions)! : new ApiResponse<T>();
```
tnotheis, line 85: Hmm, this creates a valid ApiResponse even though the status code is an unsuccessful one, doesn't it?

ConsumerApi.Sdk/Client.cs
```diff
@@ -18,7 +18,7 @@ namespace Backbone.ConsumerApi.Sdk;
 
 public class Client
 {
-    public Client(Configuration configuration) : this(new HttpClient { BaseAddress = UriUtils.CreateUriFromParts(configuration.BaseUrl, "api", configuration.ApiVersion) }, configuration)
+    public Client(Configuration configuration) : this(new HttpClient { BaseAddress = UriUtils.CreateUriFromParts(configuration.BaseUrl, "api", configuration.ApiVersion, "/") }, configuration)
```
tnotheis, line 21: Instead of this "hack", I would rather suggest to update the `CreateUriFromParts` method to make sure the Uri always contains a trailing backslash.

ConsumerApi.Sdk/ConsumerApi.Sdk.csproj
```diff
@@ -5,8 +5,13 @@
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
```
tnotheis, line 9: I would prefer System.Text.Json over Newtonsoft.Json. Is there a reason why you chose Newtonsoft?

ConsumerApi.Sdk/Configuration.cs
```diff
@@ -16,4 +16,15 @@ public class AuthenticationConfiguration
...
+    public object CloneWith(AuthenticationConfiguration newAuthentication)
+    {
+        return new Configuration()
+        {
+            Authentication = newAuthentication,
+            JsonSerializerOptions = JsonSerializerOptions,
+            BaseUrl = null!,
+            ApiVersion = null!
+        };
+    }
```
tnotheis, line 29: I don't understand this method. It says "clone" but out of the 4 properties, only 1 is "cloned", while 2 are set to null (even though the properties are not even allowed to be null), and one is overridden.

ConsumerApi.Sdk/Endpoints/Devices/Types/SignedChallenge.cs
```diff
@@ -1,7 +1,21 @@
...
+using Backbone.Crypto;
+using Newtonsoft.Json;
+
+namespace Backbone.ConsumerApi.Sdk.Endpoints.Devices.Types;
 
 public class SignedChallenge
 {
-    public required string Challenge { get; set; }
-    public required byte[] Signature { get; set; }
+    [SetsRequiredMembers]
```
tnotheis, line 10: What does this attribute do?

ConsumerApi.Sdk/Services/AccountController.cs
```diff
@@ -0,0 +1,80 @@
+﻿using Backbone.ConsumerApi.Sdk.Endpoints.Common.Crypto;
+using Backbone.ConsumerApi.Sdk.Endpoints.Devices.Types;
+using Backbone.ConsumerApi.Sdk.Endpoints.Identities.Types.Requests;
+using Backbone.Crypto;
+using Backbone.Crypto.Implementations;
+using Newtonsoft.Json;
+
+namespace Backbone.ConsumerApi.Sdk.Services;
+public class AccountController(Client client)
```
tnotheis, line 9: We don't use these kinds of constructors (yet)

ConsumerApi.Sdk/Services/AccountController.cs
tnotheis, line 1: Why is there an "AccountController" all of a sudden?

ConsumerApi.Sdk/Services/PasswordHelper.cs
tnotheis, line 1: I don't think that we have to create random passwords, do we? Can't we just use one password for all users?

ConsumerApi.Tests.Integration/API/ChallengesApi.cs
```diff
@@ -10,9 +10,4 @@ internal async Task<HttpResponse<Challenge>> CreateChallenge(RequestConfiguratio
     {
         return await Post<Challenge>("/Challenges", requestConfiguration);
     }
-
-    internal async Task<HttpResponse<Challenge>> GetChallengeById(RequestConfiguration requestConfiguration, string id)
-    {
-        return await Get<Challenge>($"/Challenges/{id}", requestConfiguration);
-    }
```
tnotheis, line 17: Why did you remove this?

ConsumerApi.Tests.Integration/StepDefinitions/BaseStepDefinitions.cs
```diff
@@ -29,7 +29,7 @@ public BaseStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, ISigna
                 Username = "USRa",
                 Password = "a"
             },
-            ContentType = "application/json"
+            ContentType = "application/json",
```
tnotheis, line 32:
```suggestion
            ContentType = "application/json"
```

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
tnotheis, line 1: This seems to be solved way to complicated. But I'm not 100% certain. I will have a look by myself.

tnotheis (changes requested):

ConsumerApi.Sdk/ConsumerApi.Sdk.csproj
Dannyps: We're using `JsonConvert.DeserializeObject` in other places, and that's available through `Newtonsoft.Json`. I'm not sure if `System.Text.Json` provides an equivalent API.

ConsumerApi.Tests.Integration/API/ChallengesApi.cs
Dannyps: It is no longer used. I figured we'd want to only keep SDK methods in the future.

ConsumerApi.Sdk/Services/PasswordHelper.cs
Dannyps: We can. However, this is an SDK in first place, I see no harm in it providing a password helper. The fact that we use it for tests is a matter of convenience, the way I see it.

Nevertheless, this class can be removed.

ConsumerApi.Sdk/ConsumerApi.Sdk.csproj
tnotheis: Okay, then let's keep it for now.

ConsumerApi.Sdk/Services/PasswordHelper.cs
tnotheis: I don' think that I would expect the SDK to provide a password helper. You can put the class into our Tooling project, if you like.

ConsumerApi.Tests.Integration/API/ChallengesApi.cs
tnotheis: Okay, you're right. 👍

daniel-almeida-konkconsulting (approved):

BuildingBlocks/src/BuildingBlocks.SDK/Endpoints/Common/EndpointClient.cs
```diff
@@ -108,7 +108,19 @@ private async Task<ApiResponse<T>> Execute<T>(HttpRequestMessage request)
...
+        {
+            deserializedResponseContent = JsonSerializer.Deserialize<ApiResponse<T>>(responseContent, _jsonSerializerOptions);
+        }
+        catch (JsonException)
+        {
+            deserializedResponseContent = new ApiResponse<T>
+            {
+                Result = default
+            };
+        }
```
tnotheis, line 122: Why do you think this is a good solution?
It just covers the issue, and the user will get a seemingly valid response even though there was an error.

ConsumerApi.Sdk/Endpoints/Challenges/ChallengesEndpoint.cs
```diff
@@ -6,8 +6,23 @@ namespace Backbone.ConsumerApi.Sdk.Endpoints.Challenges;
...
+
+    public async Task<ApiResponse<Challenge>> GetChallenge(string id)
+    {
+        return await _client.Get<Challenge>($"api/{API_VERSION}/Challenges/{id}");
+    }
+
+    public async Task<ApiResponse<Challenge>> GetChallengeUnauthenticated(string id)
+    {
+        return await _client.GetUnauthenticated<Challenge>($"api/{API_VERSION}/Challenges/{id}");
+    }
```
tnotheis, line 27: It is impossible to get a challenge without authentication

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -1,49 +1,70 @@
...
+    private readonly ClientCredentials _clientCredentials;
 
-    public ChallengesApiStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, ISignatureHelper signatureHelper, ChallengesApi challengesApi, IdentitiesApi identitiesApi, DevicesApi devicesApi) :
-        base(httpConfiguration, signatureHelper, challengesApi, identitiesApi, devicesApi)
+    public ChallengesApiStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, HttpClientFactory factory)
     {
         _challengeId = string.Empty;
+        _httpClient = factory.CreateClient();
+        _clientCredentials = new ClientCredentials(httpConfiguration.Value.ClientCredentials.ClientId, httpConfiguration.Value.ClientCredentials.ClientSecret);
+        _sdk = Client.CreateUnauthenticated(_httpClient, _clientCredentials);
```
tnotheis, line 29: Is this value ever used? Not really, right? It's overriden in either `AuthenticatedUser` or `GivenTheUserIsUnauthenticated` (depending on which of those the test calls)

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -1,49 +1,70 @@
...
+        _isAuthenticated = false;
```
tnotheis, line 30: Can be removed

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -1,49 +1,70 @@
...
     }
 
+    [Given("the user is authenticated")]
+    public async Task AuthenticatedUser()
```
tnotheis, line 34:
```suggestion
    public async Task GivenTheUserIsAuthenticated()
```

ConsumerApi.Tests.Integration/api.appsettings.local.override.json
tnotheis, line 1: Why don't you use the dedicated users instead of the admin ones?

tnotheis (changes requested):

ConsumerApi.Sdk/Endpoints/Challenges/ChallengesEndpoint.cs
Dannyps: https://github.com/nmshd/backbone/blob/NMSHDB-83-create-identities-via-sdk/ConsumerApi.Tests.Integration/Features/Challenges/%7Bid%7D/GET.feature#L13-L17

In that case, should we ditch this scenario?

tnotheis: Yeah, I would only test scenarios that are possible with the sdk. At least for now.

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
Dannyps: The only line that could be removed here is the `_sdk` attribution. Even that one is necessary if we wish to allow tests which do not start with `the user is (un)authenticated`. Should this line be removed?

Also, note that it can't really be removed. We cant set `_sdk = null!;` or we can make `_sdk` nullable.

ConsumerApi.Tests.Integration/api.appsettings.local.override.json
Dannyps: This is in line with the admin API tests, as well as the Consumer API configuration. Should we revert this?

tnotheis: I don't see a reason why we shouldn't

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
tnotheis: > if we wish to allow tests which do not start with the user is (un)authenticated.

This will never be the case, I think.

> Should this line be removed?

Yes.

> we can make _sdk nullable.

Yes, do it.

tnotheis (approved):

PR #613, opened 2024-04-22
Consumer API: Rename AcceptanceContent property of Relationship class

NikolaVetnic:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Rename the `AcceptanceContent` property of `Relationship` class to `CreationResponseContent`. Enable sending content when rejecting and revoking a Relationship.


Modules/Relationships/src/Relationships.Application/Relationships/DTOs/RelationshipMetadataDTO.cs
```diff
@@ -25,6 +26,7 @@ public RelationshipMetadataDTO(Relationship relationship)
 
     public DateTime CreatedAt { get; set; }
 
+    public byte[]? CreationResponseContent { get; set; }
```
tnotheis, line 29: Why did you add this? The content hasn't been there before intentionally. This is the Relationship**Metadata**DTO, which means it doesn't have content. It is used for responses where we don't care about the content, namely the responses of all actions.

tnotheis (changes requested):

Modules/Relationships/src/Relationships.Application/Relationships/DTOs/RelationshipMetadataDTO.cs
NikolaVetnic: My bad, I misunderstood the requirements then.

Changes made.

tnotheis (approved):

PR #615, opened 2024-04-23
Consumer API: Use StatusChanged instead of RelationshipCreated DomainEvent when establishing a new Relationship

NikolaDmitrasinovic:
# Readiness checklist

-   [x] I labeled the PR.

# Description
Use `StatusChanged` insted of `RelationshipCreated` upon new `Relationship` creation. `RelationshipCreatedDomainEvent` is completely removed from the project.


tnotheis (changes requested):
You should remove the external event type

tnotheis (approved):

PR #623, opened 2024-04-25
Database migrator

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


helm/templates/databasemigrator/configmap.yaml
tnotheis, line 1: This is a kind of duplication of the "normal" configmap. The problem is that, since we're using a helm hook for the database migrator job, and the hook is running pre deployment, the "normal" configmap is not there yet. That's why I created this copy, which is also deployed as a pre deploy hook, with a lower weihgt, meaning it is deployed before the db migrator job. So it's kind of a hack, but luckily, it's not really duplication, as all the important stuff is part of the values.

.github/workflows/test.yml
```diff
@@ -62,6 +62,7 @@ jobs:
     name: Run Integration Tests (on SQL Server)
     runs-on: ubuntu-latest
     steps:
+      - run: docker compose version
```
jkoenig134, line 65: can be removed

AdminApi/src/AdminApi/Program.cs
```diff
@@ -79,7 +80,8 @@ static WebApplication CreateApp(string[] args)
     var app = builder.Build();
     Configure(app);
 
-    app.MigrateDbContext<AdminApiDbContext>();
+    if (app.Configuration.GetValue<bool>("RunMigrations"))
+        app.MigrateDbContext<AdminApiDbContext>();
```
jkoenig134, line 84: shouldn't all migrations be in the migrator?

ConsumerApi/Program.cs
```diff
@@ -98,16 +98,19 @@ static WebApplication CreateApp(string[] args)
...
-        .MigrateDbContext<ChallengesDbContext>()
-        .MigrateDbContext<DevicesDbContext>()
-        .MigrateDbContext<FilesDbContext>()
-        .MigrateDbContext<RelationshipsDbContext>()
-        .MigrateDbContext<QuotasDbContext>()
-        .MigrateDbContext<MessagesDbContext>()
-        .MigrateDbContext<SynchronizationDbContext>()
-        .MigrateDbContext<TokensDbContext>()
-        .MigrateDbContext<QuotasDbContext>();
+    if (app.Configuration.GetValue<bool>("RunMigrations"))
```
jkoenig134, line 101: same question as in admin api

AdminApi/src/AdminApi/Program.cs
tnotheis: This is only done if the "RunMigrations" config flag is set. It's for local development. This way you don't need to run the database migrator.
It's only a short term solution. Long term we will add the database migrator. I just didn't find a quick solution, so I postponed it.

jkoenig134 (dismissed):

AdminApi/src/AdminApi/Program.cs
jkoenig134: gotcha. can you make sure that this is only possible in development?

tnotheis: Hmm, I could. But I don't think it's worth it. Because the users won't have permission to create/update/drop tables in production environments anyway.

jkoenig134 (approved):

PR #625, opened 2024-04-25
Consumer API: Get registered identity details

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Modules/Devices/src/Devices.Application/DTOs/RegisteredIdentityDetailsDTO.cs
+++ b/Modules/Devices/src/Devices.Application/DTOs/RegisteredIdentityDetailsDTO.cs
@@ -1,0 +1,23 @@
+﻿using Backbone.Modules.Devices.Domain.Entities.Identities;
+using DateTime = System.DateTime;
+
+namespace Backbone.Modules.Devices.Application.DTOs;
+public class RegisteredIdentityDetailsDTO
+{
+    public RegisteredIdentityDetailsDTO(Identity identity)
+    {
+        Address = identity.Address.ToString();
+        CreatedAt = identity.CreatedAt;
+        IdentityVersion = identity.IdentityVersion;
+        TierId = identity.TierId;
+        Status = identity.Status;
+        DeletionGracePeriodEndsAt = identity.DeletionGracePeriodEndsAt;
+    }
+
+    public string Address { get; set; }
+    public DateTime CreatedAt { get; set; }
+    public byte IdentityVersion { get; set; }
+    public string TierId { get; set; }
+    public IdentityStatus Status { get; set; }
+    public DateTime? DeletionGracePeriodEndsAt { get; set; }
+}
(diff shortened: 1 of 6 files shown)


Modules/Devices/src/Devices.Application/DTOs/RegisteredIdentityDetailsDTO.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Backbone.Modules.Devices.Domain.Entities.Identities;
+using DateTime = System.DateTime;
+
+namespace Backbone.Modules.Devices.Application.DTOs;
+public class RegisteredIdentityDetailsDTO
```
tnotheis, line 5:
```suggestion
public class OwnIdentityDTO
```

Modules/Devices/src/Devices.Application/Identities/Queries/GetRegisteredIdentityDetails/GetRegisteredIdentityDetailsQuery.cs
```diff
@@ -0,0 +1,6 @@
+﻿using MediatR;
+
+namespace Backbone.Modules.Devices.Application.Identities.Queries.GetRegisteredIdentityDetails;
+public class GetRegisteredIdentityDetailsQuery : IRequest<GetRegisteredIdentityDetailsResponse>
```
tnotheis, line 4:
```suggestion
public class GetOwnIdentityQuery : IRequest<GetOwnIdentityResponse>
```

Do this renaming everywhere it applies, please

Modules/Devices/src/Devices.Application/Identities/Queries/GetRegisteredIdentityDetails/Handler.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public Handler(IIdentitiesRepository identitiesRepository, IUserContext userContext)
+    {
+        _identitiesRepository = identitiesRepository;
+        _userContext = userContext;
+    }
+
+    public async Task<GetRegisteredIdentityDetailsResponse> Handle(GetRegisteredIdentityDetailsQuery request, CancellationToken cancellationToken)
+    {
+        var identity = await _identitiesRepository.FindByAddress(_userContext.GetAddress(), cancellationToken) ?? throw new NotFoundException(nameof(Identity));
```
tnotheis, line 21: You shouldn't throw a NotFoundException here. If there is nothing coming back from the repository, this means that either the JWT is invalid, or that the data of the identity is lost. Both are not the fault of the user and should therefore result in a 500 instead of a 4xx.
Throw a normal exception with a meaningful error message.

tnotheis (changes requested):

Modules/Devices/src/Devices.Application/DTOs/RegisteredIdentityDetailsDTO.cs
HunorTotBagi: Changes made.

Modules/Devices/src/Devices.Application/Identities/Queries/GetRegisteredIdentityDetails/GetRegisteredIdentityDetailsQuery.cs
HunorTotBagi: Changes made.

Modules/Devices/src/Devices.Application/Identities/Queries/GetRegisteredIdentityDetails/Handler.cs
HunorTotBagi: Changes made.

tnotheis (approved):

PR #633, opened 2024-05-03
Improve Admin API and Consumer API SDKs

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


jkoenig134 (dismissed):

AdminApi.Sdk/Authentication/XsrfAndApiKeyAuthenticator.cs
daniel-almeida-konkconsulting, line 1: The namespace doesn't match the file location.

AdminApi.Sdk/Authentication/XsrfAndApiKeyAuthenticator.cs
```diff
@@ -40,7 +40,7 @@ private async Task<string> GetCookie()
     [MemberNotNull(nameof(_xsrfToken), nameof(_xsrfCookie))]
     private async Task RefreshToken()
     {
-        HttpRequestMessage request = new(HttpMethod.Get, "xsrf");
+        HttpRequestMessage request = new(HttpMethod.Get, "api/v1/xsrf");
```
daniel-almeida-konkconsulting, line 43: Would be nice to avoid having the Api Version hardcoded in more than 1 place.
We already have the `const` in the `AdminApiEndpoint`, any chance we can either use it or move it elsewhere?

AdminApi.Sdk/Client.cs
```diff
@@ -7,16 +8,20 @@
...
 {
-    public Client(Configuration config)
+    private Client(HttpClient httpClient, string apiKey)
     {
-        var httpClient = new HttpClient { BaseAddress = new Uri(config.BaseUrl) };
-        var authenticator = new XsrfAndApiKeyAuthenticator(config.ApiKey, httpClient);
-        var endpointClient = new EndpointClient(httpClient, authenticator, config.JsonSerializerOptions);
+        var authenticator = new XsrfAndApiKeyAuthenticator(apiKey, httpClient);
+
+        var jsonSerializerOptions = new JsonSerializerOptions() { PropertyNameCaseInsensitive = true };
```
daniel-almeida-konkconsulting, line 21: `var jsonSerializerOptions = new JsonSerializerOptions { PropertyNameCaseInsensitive = true };`

daniel-almeida-konkconsulting (changes requested):

AdminApi.Sdk/Authentication/XsrfAndApiKeyAuthenticator.cs
tnotheis: I want to get rid of the ApiVersion in the Admin API anyway, because we are and probably will always be the only users of the API. So no need for versioning here.

jkoenig134 (approved):

PR #642, opened 2024-05-08
Sidecar for performance tests

Dannyps:
# Readiness checklist

-   [x] I labeled the PR.


performance-tests/crypto-sidecar/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+app.use(async function (ctx, next) {
+    if (ctx.method !== "GET" || ctx.path !== "/keypair") return await next();
+
+    ctx.body = JSON.stringify(await crypto.CryptoSignatures.generateKeypair());
+    ctx.status = 200;
+});
+
+app.use(async function (ctx, next) {
+    if (ctx.method !== "GET" || ctx.path !== "/password") return await next();
```
jkoenig134, line 30: I don't know koa, but is there no `app.get("/password", callback)`?

Dannyps: I only found out about [koa-router](https://www.npmjs.com/package/koa-router) when finishing this up. It's something we can add if deemed necessary.

jkoenig134: If koa doesn't support this out of the box I would really vote for removing koa and adding express.

tnotheis: I agree with @jkoenig134. The current code looks really ugly.

performance-tests/crypto-sidecar/index.js
```diff
@@ -0,0 +1,53 @@
...
+    await next();
+    const ms = Date.now() - start;
+    console.log(`${ctx.method} ${ctx.url} - ${ms} ms`);
+});
+
+app.use(async function (ctx, next) {
+    if (ctx.method !== "GET" || ctx.path !== "/keypair") return await next();
+
+    ctx.body = JSON.stringify(await crypto.CryptoSignatures.generateKeypair());
+    ctx.status = 200;
```
MH321Productions, line 26: I'm no expert in JS/TS, but I would suggest adding a `ctx.type = "application/json";`, so that the response has an explicit type

performance-tests/crypto-sidecar/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+    const challenge = crypto.CoreBuffer.fromUtf8(body.challenge);
+    const privateKey = body.keyPair.prv;
+
+    const cspk = crypto.CryptoSignaturePrivateKey.fromJSON({ privateKey: privateKey.prv, algorithm: privateKey.alg });
+
+    const signedChallenge = await crypto.CryptoSignatures.sign(challenge, cspk);
+
+    ctx.body = signedChallenge.toJSON();
+    ctx.status = 200;
```
MH321Productions, line 48: Here as well

MH321Productions (changes requested):

performance-tests/crypto-sidecar/index.js
Dannyps: Great idea, ty!

tnotheis (changes requested):

performance-tests/src/scenario-02/s02.test.ts
```diff
@@ -0,0 +1,250 @@
...
+            rate: 8,
+            timeUnit: "1s",
+            duration: "1m",
+            preAllocatedVUs: 2,
+            maxVUs: 5
+        }
+    }
+};
+
+const session = new Httpx({
```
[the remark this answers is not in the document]
Dannyps: I suppose we could call it `client`.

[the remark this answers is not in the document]
tnotheis: Yeah, for example.

performance-tests/package.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "performance-test-challenges",
```
tnotheis, line 2: Why "challenges"?

performance-tests/src/libs/backbone-client/identity.ts
```diff
@@ -0,0 +1,65 @@
...
+            headers: { "Content-Type": "application/json" }
+        }) as Response;
+        return { httpResponse, generatedPassword };
+    } catch (e) {
+        console.error(e);
+        throw e;
+    }
+}
+
+export function exchangeToken(client: Httpx, createdIdentityResponse: CreateIdentityResponse, password: string) : TokenResponse {
```
tnotheis, line 39: Why are you passing the whole `CreateIdentityResponse` when you only need the username?

performance-tests/src/libs/crypto-helper.ts
```diff
@@ -0,0 +1,51 @@
+import { Httpx } from "https://jslib.k6.io/httpx/0.1.0/index.js";
+
+export class CryptoHelper {
+    public static readonly session = new Httpx({
```
tnotheis, line 4: "session" is a bad name

performance-tests/src/models/token.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+export interface TokenResponse {
```
tnotheis, line 2: Should be called `JwtResponse`. `Token` is misleading because there are Tokens in enmeshed.

performance-tests/src/scenario-02/s02.test.ts
```diff
@@ -0,0 +1,130 @@
...
+            duration: "1m",
+            preAllocatedVUs: 10
+        }
+    }
+};
+
+const client = new Httpx({
+    baseURL: "http://localhost:8081/api/v1/",
+    timeout: 20000 // 20s timeout
+}) as any; // TODO: remove cast to any
```
tnotheis, line 25: I told you fix the typings and remove the TODO comment afterwards. None of both has happened

performance-tests/src/scenario-02/s02.test.ts
```diff
@@ -0,0 +1,130 @@
...
+            /* eslint-disable @typescript-eslint/naming-convention */
+            "Content-Type": "application/json",
+            "X-Supported-Datawallet-Version": dataWalletVersion,
+            Authorization: `Bearer ${identity.token.access_token}`
+            /* eslint-enable @typescript-eslint/naming-convention */
+        }
+    }) as Response;
+
+    check(startSyncRunResponse, {
+        /* eslint-disable @typescript-eslint/naming-convention */
```
tnotheis, line 50: Disabling the rule everywhere is tedious and leads to really ugly code. You should find another way for this. E.g. change the naming convention configuration to allow spaces.

performance-tests/src/scenario-02/s02.test.ts
```diff
@@ -0,0 +1,130 @@
...
+            /* eslint-enable @typescript-eslint/naming-convention */
+        });
+
+        const createdIdentityResponseValue = httpResponse.json("result") as unknown as CreateIdentityResponse | undefined;
+
+        check(createdIdentityResponseValue, {
+            /* eslint-disable @typescript-eslint/naming-convention */
+            "response has Address": (r) => r?.address !== undefined,
+            "response has device": (r) => r?.device !== undefined,
+            "device has Id": (r) => r?.device.id !== undefined
```
tnotheis, line 75: Please, stay consistent. In all the other checks the first char is uppercase. Here it's lowercase

performance-tests/src/models/challenge.ts
```diff
@@ -0,0 +1,6 @@
+export interface ChallengeResponse {
```
tnotheis, line 1:
```suggestion
export interface CreateChallengeResponse {
```

performance-tests/src/libs/crypto-helper.ts
```diff
@@ -0,0 +1,51 @@
...
+    prv: {
+        prv: string;
+        alg: number;
+    };
+}
+
+export interface ChallengeRequestRepresentation {
+    expiresAt: string;
+    id: string;
+    type: string;
```
tnotheis, line 50:
```suggestion
    id: string;
    expiresAt: string;
    type: string;
```

performance-tests/src/libs/crypto-helper.ts
```diff
@@ -0,0 +1,51 @@
...
+        alg: number;
+    };
+
+    prv: {
+        prv: string;
+        alg: number;
+    };
+}
+
+export interface ChallengeRequestRepresentation {
```
tnotheis, line 47: I don't like the name "ChallengeRequestRepresenation". Which request are you referring to? And what does it represent?

performance-tests/src/libs/crypto-helper.ts
tnotheis, line 50: Do we really need the `type` property?

tnotheis (changes requested):

performance-tests/crypto-sidecar/index.js
```diff
@@ -0,0 +1,51 @@
...
+    const start = Date.now();
+    await next();
+    const ms = Date.now() - start;
+    console.log(`${ctx.method} ${ctx.url} - ${ms} ms`);
+});
+
+router.get("/keypair", async (ctx) => {
+    ctx.body = (await crypto.CryptoSignatures.generateKeypair()).toJSON();
+    ctx.status = 200;
+    ctx.type = "application/json";
```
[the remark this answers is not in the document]
jkoenig134: switching from koa to express would remove **all** confusion here, just saying 😆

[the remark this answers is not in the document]
tnotheis: Since koa seems to be a "better"/newer version of express, I'd try to go with it. Even though I don't like it either. 😉

performance-tests/crypto-sidecar/index.js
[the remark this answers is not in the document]
Dannyps: ~~The way I see it, the sidecar will be done™️ once we close this PR. I see the value of TypeScript in an evolving code base. Here, it just seems like needless work.~~

Should we switch to TypeScript?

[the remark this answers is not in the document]
tnotheis: I don't see any disadvantage in using typescript. Except for the minor increase in complexity.

performance-tests/crypto-sidecar/index.js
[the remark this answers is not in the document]
tnotheis: @Dannyps What about this comment?

performance-tests/src/libs/backbone-client/identity.ts
```diff
@@ -36,30 +36,31 @@ export function createIdentity(client: Httpx, clientId: string, clientSecret: st
...
         /* eslint-enable @typescript-eslint/naming-convention */
     };
-    return client.post("http://localhost:8081/connect/token", payload, {
-        headers: {
-            // eslint-disable-next-line @typescript-eslint/naming-convention
-            "Content-Type": "application/x-www-form-urlencoded"
-        }
-    }).json() as unknown as TokenResponse;
+    return client
+        .post("http://localhost:8081/connect/token", payload, {
```
tnotheis, line 50: I don't like the hardcoded domain here. This makes it impossible to configure it. Instead, we should do it similarly to how we do it in the SDK:
* The base url should be just the domain (without a path)
* There should be a constant containing the api version
* The api version should be appended on every request (if applicable)

performance-tests/crypto-sidecar/index.js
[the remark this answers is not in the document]
Dannyps: Koa and Express are basically equivalent for our use-case and we can use whichever is preferred.

performance-tests/src/libs/backbone-client/identity.ts
```diff
@@ -1,10 +1,10 @@
...
-import { CreateIdentityRequest, CreateIdentityResponse } from "../../models/identity";
-import { TokenResponse } from "../../models/token";
-import { ChallengeRequestRepresentation, CryptoHelper } from "../crypto-helper";
-import { Httpx } from "https://jslib.k6.io/httpx/0.1.0/index.js";
+import { CreateChallengeResponse } from "../../models/challenge";
+import { CreateIdentityRequest } from "../../models/identity";
+import { JwtResponse } from "../../models/jwt-response";
+import { ChallengeRequestPayload, CryptoHelper } from "../crypto-helper";
 
 export function createIdentity(client: Httpx, clientId: string, clientSecret: string): { httpResponse: Response; generatedPassword: string } {
```
tnotheis, line 9: If we need the password outside of this method, we should probably rather pass it as a parameter into this method, instead of returning it to the caller.

performance-tests/crypto-sidecar/index.js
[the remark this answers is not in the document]
tnotheis: I was referring to the original comment. Not to what Julian said.

performance-tests/custom-typings/k6-httpx/index.d.ts
```diff
@@ -0,0 +1,16 @@
+import { Response } from "k6/http";
+declare module "https://jslib.k6.io/httpx/0.1.0/index.js" {
+
+    export class Httpx {
+        constructor(options: Options);
+        get(...args: any[]): Response;
+        post(...args: any[]): Response;
+        put(...args: any[]): Response;
```
tnotheis, line 8: > ...args: any[]

These aren't type definitions... All you're doing is turning off typescript.

tnotheis (changes requested):

performance-tests/custom-typings/k6-httpx/index.d.ts
Dannyps: I'm not being able to use this custom typing. Moreover, I'm not sure how to go about this. Is there a way to import types from [`k6/http`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/k6/http.d.ts)?

tnotheis: > I'm not being able to use this custom typing.

So you thought that ignoring it is the way to go...? 😑 

I checked it, and it seems like you broke it when you started to define `Response` as the response type. I don't know exactly why this is happening, but I suppose it's because you're importing the whole class (=> the .js file) in a type mappings (.d.ts) file.

> Is there a way to import types from [`k6/http`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/k6/http.d.ts)?

I don't know. I'm afraid you're gonna have to google it.

performance-tests/src/scenario-02/s02.test.ts
```diff
@@ -54,7 +56,8 @@ export function setup(): IdentityWithToken[] {
     const testIdentities: IdentityWithToken[] = [];
 
     for (let i = 0; i < scenario.preAllocatedVUs; i++) {
-        const { httpResponse, generatedPassword } = createIdentity(client, "test", "test");
+        const password = "test-password";
+        const httpResponse = createIdentity(client, "test", "test", password);
```
tnotheis, line 60: I think we should extract these into global constants. You can put them into the same file as the api version

tnotheis (changes requested):

performance-tests/src/scenario-02/s02.test.ts
tnotheis: Ah, and by "these" I mean password, clientId and clientSecret.

tnotheis:
Hmm, I just noticed: I don't think we should include the first performance test scenario in a PR with the name "Sidecar for performance tests"

Dannyps:
The scenario is added as an example, I guess. But I agree that the title is misleading. Perhaps `Sidecar for performance tests with example scenario`?

performance-tests/src/scenario-02/s02.test.ts
```diff
@@ -0,0 +1,117 @@
...
+            executor: "constant-arrival-rate",
+            rate: 10,
+            timeUnit: "1s",
+            duration: "1m",
+            preAllocatedVUs: 10
+        }
+    }
+};
+
+const apiVersion = "v1";
```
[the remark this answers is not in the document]
Dannyps: Please clarify if the whole test must always use the same API version for all modules/controllers.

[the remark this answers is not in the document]
tnotheis: For now, yes. You can see in the SDKs that we also use the same API version everywhere.

tnotheis:
> The scenario is added as an example, I guess. But I agree that the title is misleading. Perhaps `Sidecar for performance tests with example scenario`?

I don't think we need an example scenario for now. The first scenario will soon be implemented anyway.

performance-tests/custom-typings/k6-httpx/index.d.ts
Dannyps: I had to go with manually copying the methods, but it seems to be working now. Also, it seems these custom typings cannot have any kind of imports.

tnotheis (approved):

PR #644, opened 2024-05-08
Admin UI: Tier details

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.


tnotheis:
I just made the following changes regarding the expansion tiles:
1. I made them collapsed by default. This should improve overview. And I think that most of the time you should only need either the identities or the quotas, but not both.
2. I increased the gap between the quotas tile and the identity tile
3. I removed the border around the expanded tiles. @jkoenig134 please have a look at [how i did it](https://github.com/nmshd/backbone/pull/644/commits/56478845bd741b052464f2d3009527db1a7dc880. I'm not sure it's the best way to achieve this (the diff is kind of weird; what I did is I wrapped the `ExpansionTile`s in a `Theme` for which I set the `deviderColor` to transparent (e.g. line 119-120))

tnotheis (dismissed):

tnotheis (dismissed):

tnotheis (approved):

PR #652, opened 2024-05-13
Integrate domain events into Domain (Quotas Module)

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/DeleteTierQuotaDefinition/HandlerTests.cs
```diff
@@ -47,14 +43,18 @@ public async Task Triggers_TierQuotaDefinitionDeletedDomainEvent()
         await handler.Handle(command, CancellationToken.None);
 
         // Assert
-        A.CallTo(() => _eventBus.Publish(A<DomainEvent>.That.IsInstanceOf(typeof(TierQuotaDefinitionDeletedDomainEvent)))).MustHaveHappenedOnceExactly();
+        tier.DomainEvents.Should().HaveCount(2);
+        tier.DomainEvents[1].Should().BeOfType<TierQuotaDefinitionDeletedDomainEvent>();
+        var domainEvent = (TierQuotaDefinitionDeletedDomainEvent)tier.DomainEvents[1]; //TODO: Timo (Should I make this a domain test as well or is this handler test enough?)
```
tnotheis, line 48: > Should I make this a domain test as well or is this handler test enough?

As I already said in the other PR: the domain events tests in general should ALWAYS be ONLY domain tests (obviously, as they are domain logic)

Modules/Quotas/test/Quotas.Domain.Tests/Tests/Tiers/TierTests.cs
```diff
@@ -187,12 +186,12 @@ public void AddQuotaForAllMetricsOnQueuedForDeletion_only_creates_missing_quotas
...
 
-
 file static class TierExtensions
 {
     public static Tier Clone(this Tier tier)
     {
-        var serialized = JsonConvert.SerializeObject(tier);
-        return JsonConvert.DeserializeObject<Tier>(serialized)!;
+        var newTier = new Tier(tier.Id, tier.Name);
+        newTier.Quotas.AddRange(tier.Quotas);
```
tnotheis, line 194: We don't modify entities directly. Only with the help of methods.

Was there something wrong with the json serialization+deserialization?

Modules/Quotas/src/Quotas.Domain/Aggregates/Tiers/Tier.cs
```diff
@@ -71,6 +75,8 @@ public IEnumerable<TierQuotaDefinition> AddQuotaForAllMetricsOnQueuedForDeletion
         var quotaDefinition = new TierQuotaDefinition(metricKey, max, period);
         Quotas.Add(quotaDefinition);
 
+        RaiseDomainEvent(new QuotaCreatedForTierDomainEvent(Id, quotaDefinition.Id));
```
tnotheis, line 78: I just realized that the naming of the events is inconsistent: `QuotaCreatedForTierDomainEvent` vs. `TierQuotaDefinitionDeletedDomainEvent`. Can you rename `QuotaCreatedForTierDomainEvent` to `TierQuotaDefinitionCreatedDomainEvent`?

tnotheis (changes requested):

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/DeleteTierQuotaDefinition/HandlerTests.cs
tnotheis, line 48: Reminder: change this test as well, as described in my other comment

tnotheis: Oops, GitHub played a trick on me here. Ignore this.

Modules/Quotas/test/Quotas.Application.Tests/Tests/Quotas/CreateQuotaForTier/HandlerTests.cs
```diff
@@ -79,13 +74,14 @@ public async Task Triggers_QuotaCreatedForTierDomainEvent()
         await handler.Handle(command, CancellationToken.None);
 
         // Assert
-        A.CallTo(() => _eventBus.Publish(A<DomainEvent>.That.IsInstanceOf(typeof(QuotaCreatedForTierDomainEvent)))).MustHaveHappened();
+        var domainEvent = tier.Should().HaveASingleDomainEvent<QuotaCreatedForTierDomainEvent>();
+        domainEvent.TierId.Should().Be(tierId);
```
tnotheis, line 78: Reminder: change this test as well, as described in my other comment

Modules/Quotas/src/Quotas.Domain/Aggregates/Tiers/TierQuotaDefinition.cs
```diff
@@ -1,9 +1,10 @@
+using Backbone.BuildingBlocks.Domain;
 using Backbone.Modules.Quotas.Domain.Aggregates.Identities;
-using Backbone.Modules.Quotas.Domain.Aggregates.Metrics;
+using MetricKey = Backbone.Modules.Quotas.Domain.Aggregates.Metrics.MetricKey;
```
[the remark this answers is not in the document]
MH321Productions: There are two `MetricKey` classes: One in `Backbone.BuildingBlocks.Domain` and the other in `Backbone.Modules.Quotas.Domain.Aggregates.Metrics`. Since I import both namespaces (bc `Entity` is in `Backbone.BuildingBlocks.Domain`), I have to specify which `MetricKey` to use.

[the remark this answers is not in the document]
tnotheis: Ah, I see. Weird. I wonder why MetricKey is at the root level of the BuildingBlocks. But okay, we keep it as it is for now.

Modules/Quotas/test/Quotas.Domain.Tests/Tests/Tiers/TierTests.cs
MH321Productions: It used `Newtonsoft.Json` and couldn't clone everything, which caused some errors. I now replaced it with the .NET serialization, which now works fine.

tnotheis: Okay, cool 👍

Modules/Quotas/src/Quotas.Domain/Aggregates/Tiers/TierId.cs
tnotheis, line 1: You should make the constructor private to enforce the usage of the Parse method, which leads to a better consistency.

Modules/Quotas/test/Quotas.Domain.Tests/Tests/Tiers/TierTests.cs
```diff
@@ -184,14 +187,50 @@ public void AddQuotaForAllMetricsOnQueuedForDeletion_only_creates_missing_quotas
...
+
+        // Act
+        tier.DeleteQuota(tierQuotaDefinitionId);
+
+        // Assert
+        tier.DomainEvents.Should().HaveCount(2);
+        tier.DomainEvents[1].Should().BeOfType<TierQuotaDefinitionDeletedDomainEvent>();
+        var domainEvent = (TierQuotaDefinitionDeletedDomainEvent)tier.DomainEvents[1];
+        domainEvent.TierId.Should().Be(tierId);
+        domainEvent.TierQuotaDefinitionId.Should().Be(tierQuotaDefinitionId);
```
tnotheis, line 209: You should use our extension method for this.

(to be able to do this, you should clear the domain events at the end of the Arrange block)

Modules/Quotas/test/Quotas.Domain.Tests/Tests/Tiers/TierTests.cs
```diff
@@ -184,14 +187,50 @@ public void AddQuotaForAllMetricsOnQueuedForDeletion_only_creates_missing_quotas
         addedQuotas.Should().HaveCount(1);
         tier.Quotas.Should().HaveCount(2);
     }
+
+    [Fact]
+    public void Triggers_TierQuotaDefinitionDeletedDomainEvent()
```
tnotheis, line 192:
```suggestion
    public void Deleting_a_quota_triggers_a_TierQuotaDefinitionDeletedDomainEvent()
```

Modules/Quotas/test/Quotas.Domain.Tests/Tests/Tiers/TierTests.cs
```diff
@@ -184,14 +187,50 @@ public void AddQuotaForAllMetricsOnQueuedForDeletion_only_creates_missing_quotas
...
+        // Assert
+        tier.DomainEvents.Should().HaveCount(2);
+        tier.DomainEvents[1].Should().BeOfType<TierQuotaDefinitionDeletedDomainEvent>();
+        var domainEvent = (TierQuotaDefinitionDeletedDomainEvent)tier.DomainEvents[1];
+        domainEvent.TierId.Should().Be(tierId);
+        domainEvent.TierQuotaDefinitionId.Should().Be(tierQuotaDefinitionId);
+    }
+
+    [Fact]
+    public void Triggers_TierQuotaDefinitionCreatedDomainEvent()
```
tnotheis, line 213:
```suggestion
    public void Creating_a_quota_triggers_a_TierQuotaDefinitionCreatedDomainEvent()
```

tnotheis (changes requested):

Modules/Quotas/src/Quotas.Domain/Aggregates/Tiers/TierId.cs
MH321Productions: This one breaks the Tier cloning method, bc the deserializer needs a constructor

tnotheis: Mhh, okay. Then we should maybe go with the solution under "Update in .NET 7.0" in the following comment: https://stackoverflow.com/a/73583938/4046585

MH321Productions: The .NET 7.0 solution isn't applicable for our use case, bc the Activator needs a private parameter-less constructor. But since the `TierId` is a record, the Value can't be changed afterwards, so the `TierId(string value)` constructor has to be called. However, the .NET 8.0 solution `[JsonConstructor]` works perfectly, without any further modifications.

tnotheis: There is a reason why I suggested the .NET 7 solution and not the .NET 8 one: we don't want to pollute our domain model with code from libraries or frameworks.

tnotheis (approved):

PR #659, opened 2024-05-17
Consumer API: Use Consumer API SDK for integration tests

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


ConsumerApi.Tests.Integration/StepDefinitions/DevicesStepDefinitions.cs
```diff
@@ -1,222 +1,104 @@
...
 namespace Backbone.ConsumerApi.Tests.Integration.StepDefinitions;
 
 [Binding]
 [Scope(Feature = "DELETE Device")]
-internal class DevicesStepDefinitions : BaseStepDefinitions
+internal class DevicesStepDefinitions
 {
-    private readonly KeyPair _keyPair1;
-    private readonly KeyPair _keyPair2;
+    private Client? _sdk;
```
tnotheis, line 15: Hmm, since we're using null suppression everywhere anyway, we might as well initialize the `_sdk` with `null`. And I know that I said otherwise in Daniel S.'s PR. But now that I see it heavily used, I don't think that it's a good idea to make it nullable.

Same for the other step definition files

ConsumerApi.Tests.Integration/StepDefinitions/PnsRegistrationStepDefinitions.cs
```diff
@@ -1,62 +1,57 @@
...
+    }
 
-    public PnsRegistrationStepDefinitions(
-        IOptions<HttpConfiguration> httpConfiguration, ISignatureHelper signatureHelper, ChallengesApi challengesApi, IdentitiesApi identitiesApi, DevicesApi devicesApi, PushNotificationsApi pnsRegistrationsApi) :
-        base(httpConfiguration, signatureHelper, challengesApi, identitiesApi, devicesApi)
+    [Given("the user is authenticated")]
+    public async Task GivenTheUserIsAuthenticated()
     {
-        _pnsRegistrationsApi = pnsRegistrationsApi;
+        _sdk = await Client.CreateForNewIdentity(_httpClient, _clientCredentials, PasswordHelper.GeneratePassword(20, 26));
```
tnotheis, line 30: We should use a hardcoded password instead. Same for any other occurrence that might exist

ConsumerApi.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
```diff
@@ -1,62 +1,83 @@
...
-    private HttpResponse<CreateTokenResponse>? _createTokenResponse;
-    private HttpResponse<Token>? _tokenResponse;
-    private HttpResponse<IEnumerable<Token>>? _tokensResponse;
-    private static readonly string TOMORROW_AS_STRING = DateTime.Now.AddDays(1).ToString("yyyy-MM-dd");
-    private static readonly string YESTERDAY_AS_STRING = DateTime.Now.AddDays(-1).ToString("yyyy-MM-dd");
-
-    public TokensApiStepDefinitions(IOptions<HttpConfiguration> httpConfiguration, TokensApi tokensApi, ISignatureHelper signatureHelper, ChallengesApi challengesApi, IdentitiesApi identitiesApi, DevicesApi devicesApi) :
-        base(httpConfiguration, signatureHelper, challengesApi, identitiesApi, devicesApi)
+    private ApiResponse<CreateTokenResponse>? _createTokenResponse;
+    private ApiResponse<EmptyResponse>? _createTokenResponseError;
```
tnotheis, line 27: Why do we need a separate response in case of an error?

tnotheis (changes requested):

ConsumerApi.Tests.Integration/Features/Challenges/POST.feature
```diff
@@ -5,7 +5,7 @@ User creates a Challenge
 
 Scenario: Creating a Challenge as an anonymous user
 	Given the user is unauthenticated
-	When a POST request is sent to the Challenges endpoint
+	When a POST request is sent to the Challenges endpoint without authentication
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: Because the SDK has `PostAuthenticated` and `PostUnauthenticated` method, which depending on the test need to be called.
Would you rather I change the logic of all endpoints to check whether the Client is authenticated?

ConsumerApi.Tests.Integration/Features/Devices/{id}/DELETE.feature
```diff
@@ -5,26 +5,23 @@ User deletes an un-onboarded Device
...
 	And an un-onboarded device d2
 	When a DELETE request is sent to the Devices/{id} endpoint with d2.Id
 	Then the response status code is 204 (Ok)
 	And d2 is deleted
 
 Scenario: Deleting an onboarded Device is not possible
-	Given an Identity i with a device d
-	And the current user uses d
-	When a DELETE request is sent to the Devices/{id} endpoint with d.Id
+	Given an Identity i with a device d1
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: To reuse the already existing step.
There are other tests where we mentioned a `device d1` and `device d2`.

ConsumerApi.Tests.Integration/Features/Devices/{id}/DELETE.feature
```diff
@@ -5,26 +5,23 @@ User deletes an un-onboarded Device
...
+	When a DELETE request is sent to the Devices/{id} endpoint with d1.Id
 	Then the response status code is 400 (Bad Request)
-	And the error code is "error.platform.validation.device.deviceCannotBeDeleted"
+	And the response content includes an error with the error code "error.platform.validation.device.deviceCannotBeDeleted"
```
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: "includes" is what we use on every single test when looking at errors.
would you like me to change both the consumer api and admin api integration tests?

ConsumerApi.Tests.Integration/StepDefinitions/TokensApiStepDefinitions.cs
daniel-almeida-konkconsulting: When met with a `401 Unauthorized`, the response received does not match a typical API Response, thus the Deserialization fails.
For this reason, we an `EmptyResponse`.

tnotheis: Damn... we really need a correct response in that case. :( 

What bothers me about your solution is that there's no special response in case of any other error. So it's really confusing to have this special type there.
Therefore I would suggest a `_createTokenResponse401` as a temporary workaround. In the long run we should really return our json error though.

ConsumerApi.Tests.Integration/Features/Devices/{id}/DELETE.feature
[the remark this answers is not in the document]
tnotheis: Wow, I never noticed that. ;) 
Yeah, you can change it everywhere.

ConsumerApi.Tests.Integration/Features/Devices/{id}/DELETE.feature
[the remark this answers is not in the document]
tnotheis: But reducing readability of the test just to reuse a step definition isn't the way to go.

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -56,13 +54,13 @@ public async Task GivenAChallengeC()
     [When("a POST request is sent to the Challenges endpoint with")]
     public async Task WhenAPOSTRequestIsSentToTheChallengesEndpointWith(Table table)
     {
-        _response = await _sdk!.Challenges.CreateChallenge();
+        _response = await _sdk.Challenges.CreateChallenge();
```
tnotheis, line 57: Table not used?

ConsumerApi.Tests.Integration/Features/Challenges/POST.feature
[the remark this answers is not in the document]
tnotheis: This is the only place where you added something like this. 

What's confusing me: I was just looking for a step definition for "When a POST request is sent to the Challenges endpoint without authentication", but I cannot find it anywhere. How is there no error in the tests? 😲  

What you should do though is: in the step definition for "When a POST request is sent to the Challenges endpoint", check for _isAuthenticated, and if it is `true`, call `sdk.Challenges.CreateChallenge()`. If it is `false`, call `sdk.Challenges.CreateChallengeUnauthenticated()`.

ConsumerApi.Tests.Integration/StepDefinitions/ChallengesApiStepDefinitions.cs
```diff
@@ -31,7 +30,7 @@ public ChallengesApiStepDefinitions(IOptions<HttpConfiguration> httpConfiguratio
     [Given("the user is authenticated")]
     public async Task GivenTheUserIsAuthenticated()
     {
-        _sdk = await Client.CreateForNewIdentity(_httpClient, _clientCredentials, PasswordHelper.GeneratePassword(20, 26));
+        _sdk = await Client.CreateForNewIdentity(_httpClient, _clientCredentials, "somePassword");
```
tnotheis, line 33: I think it makes sense to have the same password for all identities across all step definitions.

=> introduce a globally available constant

ConsumerApi.Tests.Integration/StepDefinitions/DevicesStepDefinitions.cs
```diff
@@ -28,8 +28,9 @@ public DevicesStepDefinitions(HttpClientFactory factory, IOptions<HttpConfigurat
         _clientCredentials = new ClientCredentials(httpConfiguration.Value.ClientCredentials.ClientId, httpConfiguration.Value.ClientCredentials.ClientSecret);
     }
 
+    [Given("an Identity i with a device d")]
     [Given("an Identity i with a device d1")]
```
tnotheis, line 32:
```suggestion
    [Given("an Identity i with a device d1?")]
```

ConsumerApi.Tests.Integration/StepDefinitions/DevicesStepDefinitions.cs
```diff
@@ -42,8 +43,9 @@ public async Task GivenAnUnOnboardedDeviceD2()
         _deviceIdD2 = client.DeviceData!.DeviceId;
     }
 
+    [When("a DELETE request is sent to the Devices/{id} endpoint with d.Id")]
     [When("a DELETE request is sent to the Devices/{id} endpoint with d1.Id")]
```
tnotheis, line 47:
```suggestion
    [When("a DELETE request is sent to the Devices/{id} endpoint with d1?.Id")]
```

tnotheis (changes requested):

tnotheis (approved):

tnotheis (approved):

PR #661, opened 2024-05-21
Consumer API: Decomposing a relationship as second Participant 

NikolaDmitrasinovic:
# Readiness checklist

-   [x] I labeled the PR.

# Description
Second participant of the relationship can now also decompose effectively transitioning relationship into `ReadyForDeletion` status.


Modules/Relationships/src/Relationships.Domain/Aggregates/Relationships/Relationship.cs
```diff
@@ -254,18 +254,27 @@ public void Decompose(IdentityAddress activeIdentity, DeviceId activeDevice)
     {
         EnsureHasParticipant(activeIdentity);
         EnsureRelationshipNotDecomposedBy(activeIdentity);
-        EnsureStatus(RelationshipStatus.Terminated);
 
         if (From == activeIdentity)
             FromHasDecomposed = true;
 
         if (To == activeIdentity)
             ToHasDecomposed = true;
```
tnotheis, line 262: It's weird that here part of the decomposition is already made, and down below there methods with the names DecomposeAs[First/Second]Participant are called.

Modules/Relationships/test/Relationships.Domain.Tests/TestHelpers/TestData.cs
```diff
@@ -70,4 +70,13 @@ public static Relationship CreateDecomposedRelationship(IdentityAddress? from =
         relationship.Decompose(decomposedBy, DEVICE_1);
         return relationship;
     }
+
+    public static Relationship CreateRelationshipInDecompositionByFirstParticipant(IdentityAddress from, IdentityAddress to)
```
tnotheis, line 74: 1. It should be "DecomposedBy" instead of Decomposition, right?
2. From the method signature you cannot see which participant is meant with "FirstParticipant". So I'd suggest to call it "...ByFrom"

Summary:

```suggestion
    public static Relationship CreateRelationshipInDecomposedByFrom(IdentityAddress from, IdentityAddress to)
```

tnotheis (changes requested):

Modules/Relationships/test/Relationships.Domain.Tests/TestHelpers/TestData.cs
NikolaDmitrasinovic: I agree completely

Modules/Relationships/src/Relationships.Domain/Aggregates/Relationships/Relationship.cs
NikolaDmitrasinovic: Yes it is, I should have given it another thought.
It's refactored now.

Modules/Relationships/src/Relationships.Domain/Aggregates/Relationships/Relationship.cs
```diff
@@ -253,25 +253,50 @@ private void EnsureRevocableReactivationRequestExistsFor(IdentityAddress activeI
     public void Decompose(IdentityAddress activeIdentity, DeviceId activeDevice)
     {
         EnsureHasParticipant(activeIdentity);
-        EnsureRelationshipNotDecomposedBy(activeIdentity);
+        EnsureRelationshipNotDecomposedBy(activeIdentity);        
+
+        if (FromHasDecomposed || ToHasDecomposed)
+            DecomposeAsSecondParticipant(activeIdentity, activeDevice);
+        else
+            DecomposeAsFirstParticipant(activeIdentity, activeDevice);
```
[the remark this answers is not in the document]
NikolaDmitrasinovic: Wish I thought of the second one...
Changed

tnotheis (approved):

tnotheis (approved):

PR #692, opened 2024-06-11
Admin UI: Use l10n translations everywhere

stamenione:
# Readiness checklist

-   [x] I labeled the PR.


AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
```diff
@@ -92,32 +92,32 @@ class _AddQuotaDialogState extends State<_AddQuotaDialog> {
                 value: _selectedMetric,
                 items: widget.availableMetrics.map((metric) => DropdownMenuItem(value: metric.key, child: Text(metric.displayName))).toList(),
                 onChanged: _saving ? null : (String? selected) => setState(() => _selectedMetric = selected),
-                decoration: const InputDecoration(
-                  border: OutlineInputBorder(),
-                  labelText: 'Metric*',
+                decoration: InputDecoration(
+                  border: const OutlineInputBorder(),
+                  labelText: context.l10n.metric,
```
Siolto, line 97: In the MeinBildungsraum app we don't want to have the `*` in the translation itself. We do: `'${context.l10n.metric}*'`

ping @jkoenig134: correct?

AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
```diff
@@ -92,32 +92,32 @@ class _AddQuotaDialogState extends State<_AddQuotaDialog> {
...
               TextField(
                 controller: _maxAmountController,
                 enabled: !_saving,
-                decoration: const InputDecoration(
-                  border: OutlineInputBorder(),
-                  labelText: 'Max Amount*',
-                  helperText: 'Only numbers greater or equal to 0 are valid.',
+                decoration: InputDecoration(
+                  border: const OutlineInputBorder(),
+                  labelText: context.l10n.maxEntities('Amount'),
```
Siolto, line 106: You are losing the `*` here

AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
```diff
@@ -153,7 +153,7 @@ class _AddQuotaDialogState extends State<_AddQuotaDialog> {
   Future<void> _addQuota() async {
     setState(() => _saving = true);
 
-    assert(_selectedMetric != null && _maxAmount != null && _selectedPeriod != null, 'Invalid state');
+    assert(_selectedMetric != null && _maxAmount != null && _selectedPeriod != null, 'Invalid State');
```
Siolto, line 156: Missing translation

AdminUi/apps/admin_ui/lib/core/widgets/app_title.dart
```diff
@@ -21,7 +21,7 @@ class AppTitle extends StatelessWidget {
           TextSpan(
             children: [
               TextSpan(text: 'enmeshed', style: textStyle.copyWith(fontWeight: FontWeight.bold)),
-              const TextSpan(text: ' Backbone Admin UI', style: textStyle),
+              const TextSpan(text: ' Backbone AdminUI', style: textStyle),
```
Siolto, line 24: missing translation

AdminUi/apps/admin_ui/lib/home/clients_overview/modals/create_client_dialog.dart
```diff
@@ -120,29 +120,29 @@ class _CreateClientDialogState extends State<_CreateClientDialog> {
...
               ),
               Gaps.h24,
               DropdownButtonFormField<String>(
                 isExpanded: true,
-                decoration: const InputDecoration(
-                  border: OutlineInputBorder(),
-                  labelText: 'Default Tier*',
+                decoration: InputDecoration(
+                  border: const OutlineInputBorder(),
+                  labelText: context.l10n.default_tier,
```
Siolto, line 145: `*` outside of the translation

AdminUi/apps/admin_ui/lib/home/tier_detail/tier_detail.dart
```diff
@@ -63,15 +63,18 @@ class _TierDetailState extends State<TierDetail> {
...
                                 TextSpan(text: tierDetails.id, style: Theme.of(context).textTheme.bodyLarge),
                               ],
                             ),
                           ),
                           Text.rich(
                             TextSpan(
                               children: [
-                                TextSpan(text: 'Name: ', style: Theme.of(context).textTheme.bodyLarge!.copyWith(fontWeight: FontWeight.bold)),
+                                TextSpan(
+                                  text: context.l10n.name,
```
Siolto, line 75: `:` outside of the translation

AdminUi/apps/admin_ui/lib/home/tier_detail/tier_detail.dart
```diff
@@ -63,15 +63,18 @@ class _TierDetailState extends State<TierDetail> {
                           Text.rich(
                             TextSpan(
                               children: [
-                                TextSpan(text: 'ID: ', style: Theme.of(context).textTheme.bodyLarge!.copyWith(fontWeight: FontWeight.bold)),
+                                TextSpan(text: context.l10n.id, style: Theme.of(context).textTheme.bodyLarge!.copyWith(fontWeight: FontWeight.bold)),
```
Siolto, line 66: `:` outside of the translation

Siolto (changes requested):
For me it seems  a little inconsistent when `Tiers` is written with a capital letter or not. Same thing with `Identities` <=> `identities`.

Maybe @tnotheis could have a look at these cases again.

AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
jkoenig134: correct, please do that.

AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
stamenione: I had a previous review by Timo and he told me to exclude the assert messages from translation, because there is not need to translate them.

AdminUi/apps/admin_ui/lib/core/widgets/app_title.dart
stamenione: For this text is the same, Timo's explanation was that this was a fixed texts and it should not be translated.

AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
stamenione: Changes made.

AdminUi/apps/admin_ui/lib/core/widgets/app_title.dart
tnotheis: If you (@jkoenig134 and @Siolto) think differently here, feel free to say so. I was just under the impression that "Backbone Admin UI" will never be translated.

jkoenig134: I'm with you!

tnotheis: On the other hand, it might be useful to extract the "product name" into some globally accessible thing, so that it's the same everywhere, and we can easily change it. I don't know though if you should do this with a constant, or with a translation.

Siolto (approved):

AdminUi/apps/admin_ui/lib/core/widgets/app_title.dart
stamenione: If it is planned to product name changes whenever language changes, then I recommend using localization. On the other hand it can be easily reverted to use constant.

AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
```diff
@@ -92,37 +92,37 @@ class _AddQuotaDialogState extends State<_AddQuotaDialog> {
...
               TextField(
                 controller: _maxAmountController,
                 enabled: !_saving,
-                decoration: const InputDecoration(
-                  border: OutlineInputBorder(),
-                  labelText: 'Max Amount*',
-                  helperText: 'Only numbers greater or equal to 0 are valid.',
+                decoration: InputDecoration(
+                  border: const OutlineInputBorder(),
+                  labelText: '${context.l10n.maxEntities('Amount')}*',
```
jkoenig134, line 106: There are many places where you use this kind of pattern. What you forget is: `Amount` never gets translated here. I think you should not mix translated and non-translated text in this way.

Please go over the whole PR with this information, there are a bunch of occurences and I will not review until this is fixed.

AdminUi/apps/admin_ui/lib/core/modals/add_quota_dialog.dart
```diff
@@ -103,7 +103,7 @@ class _AddQuotaDialogState extends State<_AddQuotaDialog> {
                 enabled: !_saving,
                 decoration: InputDecoration(
                   border: const OutlineInputBorder(),
-                  labelText: '${context.l10n.maxEntities('Amount')}*',
+                  labelText: '${context.l10n.maxEntities(context.l10n.amount)}*',
```
jkoenig134, line 106: I still don't like translations wrapped into translations. Why not just making one string per label?

stamenione: I hope that i have fixed the issue now. I thought that it was the way to do it, but clearly I was wrong.

tnotheis:
> I think we will have to refactor these translations again, i don't care if this happens in this PR or in another. When looking at [the translations in the enmeshed App](https://github.com/nmshd/app/blob/main/apps/enmeshed/lib/l10n/app_en.arb) you see that we scope translations.
> 
> * general translations like `cancel`, `add`, etc. are on global scope b/c they are re-used in the whole application
> * translations that belong to a specific screen / feature are scoped. Take the example `list_of_all_clients` - for this translation you have **no idea** where this translation is used. The key of the translation should not be just a _ separated version of the english translation. Instead use `clients_overview_title` if you have multiple word in one scope camelCase these words, for example `topLevel_pageWithMoreWords_aKeyThatNeedsMoreThanOneWord`
> 
> @tnotheis do you agree? If yes - this PR or another?

I totally agree, yes. And it should be done in this PR.

jkoenig134 (approved):

PR #695, opened 2024-06-12
Consumer API: Prevent Relationship Commands with Identities To Be Deleted

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


tnotheis:
@Dannyps I will wait with my review until you have made a first one.

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
Dannyps, line 1: I figure you could implement these changes as a PipelineBehavior instead, as it would reduce duplicate code. 
@tnotheis do you think it's a good fit here?

ConsumerApi.Tests.Integration/ConsumerApi.Tests.Integration.csproj
tnotheis, line 1: The formatting in this file is broken. Can you fix it please?

ConsumerApi.Tests.Integration/Features/Relationships/POST.feature
```diff
@@ -0,0 +1,64 @@
...
+User creates a Relationship
+
+Scenario: Creating a Relationship
+	Given Identities i1 and i2
+	And a Relationship Template rt created by i1
+	When a POST request is sent to the /Relationships endpoint by i2 with rt.id
+	Then the response status code is 201 (Created)
+	And the response contains a Relationship
+
+Scenario: Creating a Relationship with Identity "ToBeDeleted"
```
tnotheis, line 13: This sounds as if
* the identity was called "ToBeDeleted"
* you try to create the relationship **as** the identity in that status

The first bullet point also applies to the other tests.

```suggestion
Scenario: Creating a Relationship to an Identity in status "ToBeDeleted"
```

ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
```diff
@@ -0,0 +1,196 @@
...
+    {
+        _revokeRelationshipChangeResponse!.Should().BeASuccess();
+        _revokeRelationshipChangeResponse!.Should().ComplyWithSchema();
+    }
+
+    private async Task<ApiResponse<CreateRelationshipTemplateResponse>> CreateRelationshipTemplate(Client client)
+    {
+        var createRelationshipTemplateRequest = new CreateRelationshipTemplateRequest
+        {
+            Content = ConvertibleString.FromUtf8("AAA").BytesRepresentation
```
tnotheis, line 180: I start to see this `ConvertibleString.FromUtf8("AAA").BytesRepresentation` more and more often. But ConvertibleString wasn't created for this purpose. 
I'd suggest an extension method that allows us to do `"AAA".GetBytes()`.
After creating it, use it everywhere please.

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
tnotheis: Well, theoretically you could do this by putting the behavior into the consumer api class. But practically this would mean that you apply the behavior to **all** commands and queries, not just the ones related to relationships.
But we could at least extract the code in question into a method in the controller.

ConsumerApi.Tests.Integration/Features/Relationships/POST.feature
```diff
@@ -0,0 +1,64 @@
...
+	And a pending Relationship between i1 and i2 initiated by i1
+	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Accept endpoint by i1
+	Then the response status code is 200 (OK)
+	And the response contains an AcceptRelationshipChangeResponse
+
+Scenario: Accept Relationship Change with Identity "ToBeDeleted"
+	Given Identities i1 and i2
+	And a pending Relationship between i1 and i2 initiated by i1
+	And i1 is in status "ToBeDeleted"
+	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Accept endpoint by i1
```
tnotheis, line 32: This doesn't make sense. `i1` cannot accept its own change request.

```suggestion
	And a pending Relationship between i1 and i2 initiated by i1
	And i1 is in status "ToBeDeleted"
	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Accept endpoint by i2
```

ConsumerApi.Tests.Integration/Features/Relationships/POST.feature
```diff
@@ -0,0 +1,64 @@
...
+	And a pending Relationship between i1 and i2 initiated by i1
+	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Reject endpoint by i1
+	Then the response status code is 200 (OK)
+	And the response contains an RejectRelationshipChangeResponse
+
+Scenario: Reject Relationship Change with Identity "ToBeDeleted"
+	Given Identities i1 and i2
+	And a pending Relationship between i1 and i2 initiated by i1
+	And i1 is in status "ToBeDeleted"
+	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Reject endpoint by i1
```
tnotheis, line 47: See above.
```suggestion
	And a pending Relationship between i1 and i2 initiated by i1
	And i1 is in status "ToBeDeleted"
	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Reject endpoint by i2
```

ConsumerApi.Tests.Integration/Features/Relationships/POST.feature
```diff
@@ -0,0 +1,64 @@
...
+	And a pending Relationship between i1 and i2 initiated by i1
+	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Revoke endpoint by i2
+	Then the response status code is 200 (OK)
+	And the response contains an RevokeRelationshipChangeResponse
+
+Scenario: Revoke Relationship Change with Identity "ToBeDeleted"
+	Given Identities i1 and i2
+	And a pending Relationship between i1 and i2 initiated by i1
+	And i1 is in status "ToBeDeleted"
+	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Revoke endpoint by i2
```
tnotheis, line 62: Here it's the other way around: a relationship change request can **only** be revoked by the identity that created it.

```suggestion
	And a pending Relationship between i1 and i2 initiated by i2
	And i1 is in status "ToBeDeleted"
	When a POST request is sent to the /Relationships/{r.Id}/Changes/{r.Changes.Id}/Revoke endpoint by i2
```

ConsumerApi.Tests.Integration/Features/Relationships/POST.feature
tnotheis, line 1: I just wondered if the tests wouldn't be easier to understand if the identity that does something in the `When` is always the same (probably i1)

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
```diff
@@ -121,6 +130,12 @@ public async Task<IActionResult> CreateRelationship(CreateRelationshipCommand re
     public async Task<IActionResult> AcceptRelationshipChange([FromRoute] RelationshipId relationshipId,
         [FromRoute] RelationshipChangeId changeId, CompleteRelationshipChangeRequest request, CancellationToken cancellationToken)
     {
+        var relationship = await _mediator.Send(new GetRelationshipQuery { Id = relationshipId }, cancellationToken);
```
tnotheis, line 133: As the audit log of a relationship might contain a few entries, maybe it's better to create a new query `GetPeerOfActiveIdentityInRelationship` which only returns the address of the peer. Just to save some unncessary data from being transferred from the database to the application.

tnotheis (changes requested):

ConsumerApi.Tests.Integration/Features/Relationships/POST.feature
daniel-almeida-konkconsulting: `i1` initiated the Relationship, hence it was responsible for creating the Relationship Template.
`i2` made the call to create the Relationship, leaving it Pending and waiting for `i1´ to Accept it.

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
```diff
@@ -160,6 +176,13 @@ public async Task<IActionResult> CreateRelationship(CreateRelationshipCommand re
 
         return Ok(change);
     }
+
+    private async Task EnsurePeerIsValid(string peerIdentityAddress, CancellationToken cancellationToken)
```
Dannyps, line 180: It's not that the peer is invalid or does not exist. It's just that it cannot accept the action that is being carried out because it is pending deletion. Perhaps this could be renamed to something like `EnsurePeerIsNotToBeDeleted`, as that would make it much clearer what's going on here.

Dannyps (changes requested):

Dannyps (dismissed):

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
```diff
@@ -112,6 +116,9 @@ public async Task<IActionResult> GetChangeById(RelationshipChangeId id, Cancella
     [ProducesResponseType(typeof(HttpResponseEnvelopeResult<CreateRelationshipResponse>), StatusCodes.Status200OK)]
     public async Task<IActionResult> CreateRelationship(CreateRelationshipCommand request, CancellationToken cancellationToken)
     {
+        var relationshipTemplate = await _mediator.Send(new GetRelationshipTemplateQuery { Id = request.RelationshipTemplateId }, cancellationToken);
```
tnotheis, line 119: Why didn't you move this call to `EnsurePeerIsNotToBeDeleted` as well?

Modules/Relationships/src/Relationships.Infrastructure/Persistence/Database/Repository/RelationshipsRepository.cs
```diff
@@ -67,6 +67,15 @@ public RelationshipsRepository(RelationshipsDbContext dbContext)
         return changes;
     }
 
+    public async Task<IdentityAddress> FindRelationshipPeer(RelationshipId id, IdentityAddress identityAddress, CancellationToken cancellationToken)
+    {
+        var relationship = await _readOnlyRelationships
+            .WithParticipant(identityAddress)
+            .FirstWithId(id, cancellationToken);
+
+        return relationship.To;
```
tnotheis, line 76: ...... You're always returning `To`. But what if the peer is in `From`...?

tnotheis (changes requested):

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
daniel-almeida-konkconsulting: In this instance, we have a Relationship Template Id, which we use to retrieve the template itself and the identity address.
In other cases, we have a Relationship Id, and we retrieve an actual Relationship.

tnotheis (dismissed):

tnotheis (approved):

PR #697, opened 2024-06-12
Use Caching in GitHub Actions

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

The caching cuts the runtime of the workflow by up to 4 minutes (~11 minutes to ~7 minutes)


.github/workflows/test.yml
```diff
@@ -53,28 +73,168 @@ jobs:
...
+          - dockerfile: ConsumerApi/Dockerfile
+            tag: consumer-api
+          - dockerfile: AdminApi/src/AdminApi/Dockerfile
+            tag: admin-ui
+          - dockerfile: Modules/Devices/src/Devices.AdminCli/Dockerfile
+            tag: seed-client
+          - dockerfile: DatabaseMigrator/Dockerfile
+            tag: database-migrator
+          - dockerfile: EventHandlerService/src/EventHandlerService/Dockerfile
+            tag: event-handler-service
```
tnotheis, line 122: Why do we have tags that are named just like the image? If the tag is mandatory, we can just hardcode it to something like "test"

.github/workflows/test.yml
```diff
@@ -53,28 +73,168 @@ jobs:
...
+        run: docker compose -f ./.ci/docker-compose.test.yml -f ./.ci/docker-compose.test.postgres.yml logs > docker-log.txt
+
+      - name: Archive logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: integration-test-postgres-docker-logs
+          path: docker-log.txt
+        env:
+          DOTNET_CONSOLE_ANSI_COLOR: true
```
tnotheis, line 183: Why are you saving docker compose logs when you didn't even do anything with compose in this job?

.github/workflows/test.yml
```diff
@@ -53,28 +73,168 @@ jobs:
...
+          path: ~/.nuget/packages
+          key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj') }}
+          restore-keys: ${{ runner.os }}-nuget-
       - name: Run tests
         run: ./.ci/test.sh
         env:
           DOTNET_CONSOLE_ANSI_COLOR: true
+          
+  container-test-builds:
+    name: Build ${{matrix.tag}} Container
```
tnotheis, line 108: You only build the image. Not the container.

.github/workflows/test.yml
tnotheis, line 122: Is there a specific reason why you only build these five images instead of all?

Also, you should be able to remove the jobs that are specifically for building the images. That's just duplication, I think.

.github/workflows/test.yml
```diff
@@ -87,16 +247,48 @@ jobs:
...
+          pattern: docker-*
+          merge-multiple: true
+
+      - name: Load Docker images
+        run: |
+          docker load --input /tmp/consumer-api.tar
+          docker load --input /tmp/admin-ui.tar
+          docker load --input /tmp/seed-client.tar
+          docker load --input /tmp/database-migrator.tar
+          docker load --input /tmp/event-handler-service.tar
```
tnotheis, line 283: This seems to be very error prone. You can easily forget to add a new image here. And then you're running without cache

Isn't it possible to just load all tars?

.github/workflows/test.yml
```diff
@@ -87,16 +247,48 @@ jobs:
...
+
+      - name: Setup NuGet
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 8.x
+
       - name: Install script dependencies
         run: npm install --prefix ./.ci
+
+      - name: Download Docker images
```
tnotheis, line 270:
```suggestion
      - name: Download cached Docker images
```

Same for the other occurrences

tnotheis (changes requested):

.ci/integrationTest.postgres.sh
```diff
@@ -5,11 +5,17 @@ dockerCompose() {
...
+  dotnet restore "Backbone.sln";
+  dotnet build --no-restore "Backbone.sln"
+}
+wait
+
 export CONSUMER_API_BASE_ADDRESS="http://localhost:5000"
 export ADMIN_API_BASE_ADDRESS="http://localhost:5173"
-dotnet test --no-restore --no-build --filter "Category=Integration&TestCategory!~ignore" --logger "GitHubActions;summary.includePassedTests=true" "Backbone.sln"
+
+dotnet test --no-restore --no-build --filter "Category=Integration&TestCategory!~ignore" --logger "GitHubActions;summary.includePassedTests=true;summary.includeSkippedTests=true" "Backbone.sln"
```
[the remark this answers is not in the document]
MH321Productions: I just copied the arguments from another place as part of the usage of the GitHub Actions logger

.github/workflows/test.yml
MH321Productions: These are the only images containing a build instruction in the compose files (as this was my focus)

.github/workflows/test.yml
```diff
@@ -38,12 +38,32 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
+          check-latest: true
```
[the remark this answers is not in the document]
tnotheis: Yes.

.ci/integrationTest.postgres.sh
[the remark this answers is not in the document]
tnotheis: But not from our code, right? Because we don't use `summary.includeSkippedTests=true` anywhere, as far as I can see.

.github/workflows/test.yml
tnotheis: Yeah, I was expecting that. But I like your solution for two reasons:
1. It leads to less code
2. It then also benefits from the caching of the used base images.

.ci/integrationTest.postgres.sh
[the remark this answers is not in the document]
MH321Productions: I don't know anymore. I just removed it

.github/workflows/test.yml
tnotheis: Why did you ignore this comment?

MH321Productions: I didn't ignore it, I had to work on the release v5 PR, but thought about a solution

.ci/integrationTest.postgres.sh
```diff
@@ -5,11 +5,17 @@ dockerCompose() {
...
+  dotnet restore "Backbone.sln";
+  dotnet build --no-restore "Backbone.sln"
+}
+wait
+
 export CONSUMER_API_BASE_ADDRESS="http://localhost:5000"
 export ADMIN_API_BASE_ADDRESS="http://localhost:5173"
-dotnet test --no-restore --no-build --filter "Category=Integration&TestCategory!~ignore" --logger "GitHubActions;summary.includeNotFoundTests=false" "Backbone.sln"
+
+dotnet test --no-restore --no-build --filter "Category=Integration&TestCategory!~ignore" --logger "GitHubActions;summary.includePassedTests=true;summary.includeNotFoundTests=false" "Backbone.sln"
```
tnotheis, line 21: Why did you add `summary.includePassedTests=true`?

.ci/integrationTest.sqlserver.sh
```diff
@@ -5,11 +5,17 @@ dockerCompose() {
...
+  dotnet restore "Backbone.sln";
+  dotnet build --no-restore "Backbone.sln"
+}
+wait
+
 export CONSUMER_API_BASE_ADDRESS="http://localhost:5000"
 export ADMIN_API_BASE_ADDRESS="http://localhost:5173"
-dotnet test --no-restore --no-build --filter "Category=Integration&TestCategory!~ignore" --logger "GitHubActions;summary.includeNotFoundTests=false" "Backbone.sln"
+
+dotnet test --no-restore --no-build --filter "Category=Integration&TestCategory!~ignore" --logger "GitHubActions;summary.includePassedTests=true;summary.includeNotFoundTests=false" "Backbone.sln"
```
tnotheis, line 21: See above

tnotheis (changes requested):

.ci/integrationTest.postgres.sh
MH321Productions: I'm not sure what was copied in and what was merged in

tnotheis (approved):

PR #704, opened 2024-06-17
Consumer API: Endpoint for reading the current major version of the Backbone

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.


ConsumerApi/VersionService.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Backbone.ConsumerApi;
+
+public class VersionService
+{
+    public async Task<string?> GetBackboneMajorVersion()
+    {
+        var assembly = System.Reflection.Assembly.GetExecutingAssembly();
+        var version = System.Diagnostics.FileVersionInfo.GetVersionInfo(assembly.Location).FileVersion;
+
+        var majorVersion = version?.Split(['.'], 2)[0];
```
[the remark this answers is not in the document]
HunorTotBagi: I tested what the endpoint is giving me back (for version I have at the moment "1.0.0.0", checked with debugger) and it returns me 1. On my mind I had a scenario what if the major version has for example two digits. That's the reason for that 2 in the second parameter but I overlooked that. Split will be done based on `.` and later the [0] will get the first part, so it doesn't matter how many digit it contains. My bad. Changes made.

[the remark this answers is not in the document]
tnotheis: Ahh, I just tested it, and it seems the second parameter is there to stop searching for further elements after the given number of elements where found. Therefore I'd say we should keep your original solution. It's a bit better performance wise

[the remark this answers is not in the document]
HunorTotBagi: Changes made.

ConsumerApi/Dockerfile
```diff
@@ -56,6 +56,7 @@ COPY . .
 RUN dotnet restore "ConsumerApi/ConsumerApi.csproj"
 RUN dotnet publish --configuration Release --output /app/publish --no-restore "ConsumerApi/ConsumerApi.csproj"
 RUN dotnet publish --configuration Release --output /app/publish/health "HealthCheck/HealthCheck.csproj"
+RUN dotnet build --p:Version=5.6.0 "ConsumerApi/ConsumerApi.csproj"
```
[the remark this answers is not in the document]
HunorTotBagi: Changes made.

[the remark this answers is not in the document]
HunorTotBagi: Changes made.

ConsumerApi/Controllers/VersionController.cs
```diff
@@ -0,0 +1,30 @@
...
+        _versionService = versionService;
+    }
+
+    [HttpGet]
+    [AllowAnonymous]
+    [ProducesError(StatusCodes.Status404NotFound)]
+    public async Task<IActionResult> GetBackboneMajorVersion(CancellationToken cancellationToken)
+    {
+        var majorVersion = await _versionService.GetBackboneMajorVersion();
+        if (majorVersion != null)
```
tnotheis, line 25: It can never be null

ConsumerApi/Controllers/VersionController.cs
```diff
@@ -0,0 +1,30 @@
...
+    private readonly VersionService _versionService;
+
+    public VersionController(VersionService versionService, IMediator mediator) : base(mediator)
+    {
+        _versionService = versionService;
+    }
+
+    [HttpGet]
+    [AllowAnonymous]
+    [ProducesError(StatusCodes.Status404NotFound)]
```
tnotheis, line 21:
```suggestion
```

ConsumerApi/Controllers/VersionController.cs
```diff
@@ -0,0 +1,30 @@
...
+    public async Task<IActionResult> GetBackboneMajorVersion(CancellationToken cancellationToken)
+    {
+        var majorVersion = await _versionService.GetBackboneMajorVersion();
+        if (majorVersion != null)
+            return Ok(new { majorVersion });
```
tnotheis, line 26: You should document the response with the `ProducesResponseTypeAttribute`

tnotheis, line 26: Create a class for the response

ConsumerApi/Controllers/VersionController.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Backbone.BuildingBlocks.API.Mvc.ControllerAttributes;
+using Backbone.ConsumerApi.Mvc;
+using MediatR;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Backbone.ConsumerApi.Controllers;
+
+[Route("api/v1/[controller]")]
+public class VersionController : ApiControllerBase
```
tnotheis, line 10: You should't inherit from `ApiControllerBase`. It's not necessary in this case.

tnotheis (changes requested):

ConsumerApi/Program.cs
```diff
@@ -132,6 +132,8 @@ static void ConfigureServices(IServiceCollection services, IConfiguration config
     var parsedConfiguration = services.BuildServiceProvider().GetRequiredService<IOptions<BackboneConfiguration>>().Value;
 #pragma warning restore ASP0000
 
+    services.AddScoped<VersionService>();
```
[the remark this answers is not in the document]
HunorTotBagi: Changes made.

ConsumerApi/VersionService.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Backbone.ConsumerApi;
+
+public class VersionService
+{
+    public async Task<string?> GetBackboneMajorVersion()
```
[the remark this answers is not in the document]
HunorTotBagi: Changes made.

ConsumerApi/Controllers/VersionController.cs
HunorTotBagi: Changes made.

ConsumerApi/Controllers/VersionController.cs
HunorTotBagi: Changes made.

ConsumerApi/Controllers/VersionController.cs
HunorTotBagi: Changes made.

HunorTotBagi: Changes made.

ConsumerApi/Controllers/VersionController.cs
HunorTotBagi: Changes made.

tnotheis (approved):

PR #708, opened 2024-06-19
Use a central configuration file across all projects

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


AdminApi/src/AdminApi/AdminApi.csproj
```diff
@@ -8,74 +8,78 @@
...
+        <ProjectReference Include="..\..\..\Modules\Devices\src\Devices.Infrastructure\Devices.Infrastructure.csproj"/>
+        <ProjectReference Include="..\..\..\Infrastructure\Infrastructure.csproj"/>
+        <ProjectReference Include="..\..\..\Modules\Quotas\src\Quotas.Application\Quotas.Application.csproj"/>
+        <ProjectReference Include="..\..\..\Modules\Quotas\src\Quotas.Infrastructure\Quotas.Infrastructure.csproj"/>
+        <ProjectReference Include="..\AdminApi.Infrastructure.Database.Postgres\AdminApi.Infrastructure.Database.Postgres.csproj"/>
+        <ProjectReference Include="..\AdminApi.Infrastructure.Database.SqlServer\AdminApi.Infrastructure.Database.SqlServer.csproj"/>
+        <ProjectReference Include="..\AdminApi.Infrastructure\AdminApi.Infrastructure.csproj"/>
+
+        <Content Include="..\..\..\appsettings.override.json" Condition="$(Configuration) == Debug">
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
tnotheis, line 52: We don't want to make changes in the target dir, I think. So we can only copy it if it was updated.
Same for the other files

EventHandlerService/src/EventHandlerService/Dockerfile
```diff
@@ -77,7 +77,7 @@ WORKDIR "/src/EventHandlerService/src/EventHandlerService"
 RUN dotnet build "EventHandlerService.csproj" -c Release -o /app/build --no-restore
 
 FROM build AS publish
-RUN dotnet publish -c Release -o /app/publish --no-restore "EventHandlerService.csproj"
+RUN dotnet publish -c Release -o /app/publish --no-restore "EventHandlerService.csproj"
```
tnotheis, line 80: Remove the blank you added here

tnotheis (changes requested):

AdminApi/src/AdminApi/Properties/launchSettings.json
```diff
@@ -1,13 +1,14 @@
...
+  "$schema": "http://json.schemastore.org/launchsettings.json",
+  "profiles": {
+    "Default": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Local",
+        "ASPNETCORE_HOSTINGSTARTUPASSEMBLIES": "Microsoft.AspNetCore.SpaProxy",
+        "ASPNETCORE_URLS": "http://*:8082"
+      },
+      "workingDirectory": "bin/Debug/net8.0"
```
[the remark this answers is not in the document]
MH321Productions: As I told You, the override file is copied into the target directory. In order for the app to find the file, its working directory also needs to be the target directory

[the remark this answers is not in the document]
tnotheis: I don't quite get it. The dll is saved in `bin/Debug/net8.0`. So I would expect the working directory to already be that directory.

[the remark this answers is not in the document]
MH321Productions: It isn't. The default working directory is the source directory (at least in rider)

AdminApi/src/AdminApi/AdminApi.csproj
MH321Productions: I'm not sure what You mean with that, the documentation for the PreserveNewest property says:
> This item will only be copied if it is newer than the already existing item in the destination

tnotheis: What I meant is that if you edit the file in the output directory (intentionally or not), it will not be replaced until you updated the original one.
But since that's rather an edge case, and I can't find a value that's a better fit, I'm also okay with leaving it as it is.

AdminApi/src/AdminApi/Properties/launchSettings.json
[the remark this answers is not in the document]
tnotheis: Hm okay. But isn't there some kind of environment variable or similar which we can set the working directory to?
Because currently, if we update the .NET version, it will cause a lot of confusion, because it will still use the working directory of the old version. It might take a lot of time to find out the reason for this.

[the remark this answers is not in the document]
MH321Productions: Done. I used the `TargetFramework` property for the respective .net version

appsettings.override.json
```diff
@@ -0,0 +1,153 @@
...
+
+      "RabbitMQUsername": "guest", // only available for RabbitMQ
+      "RabbitMQPassword": "guest", // only available for RabbitMQ
+      "ConnectionRetryCount": 5, // only available for RabbitMQ
+
+      "GcpPubSubProjectId": "", // only available for Google Cloud Pub/Sub
+      "GcpPubSubTopicName": "" // only available for Google Cloud Pub/Sub
+    },
+
+    //Database migrator
```
tnotheis, line 29: It's also for the admin ui, afaik. I don't think we need this comment.

tnotheis: The one I commented on (=> `//Database migrator`)

MH321Productions: Ok

tnotheis (approved):

PR #728, opened 2024-07-04
Replace Newtonsoft.Json.Schema library

NikolaDmitrasinovic:
# Readiness checklist

-   [x] I labeled the PR.

# Description
Newtonsoft.Json.Schema library was replaced by the NJsonSchema open source library.

Small updates were also made to the integration tests as a result.


AdminApi/test/AdminApi.Tests.Integration/Validators/JsonValidators.cs
```diff
@@ -1,28 +1,38 @@
...
         {
             try
             {
                 var parsedJson = JObject.Parse(json);
-                return parsedJson.IsValid(schema, out errors);
+                var validationResults = schema.Validate(parsedJson);
+
+                errors.AddRange(validationResults.Select(validationResult => validationResult.ToString()));
+
+                return (validationResults.Count == 0, errors);
```
tnotheis, line 26: You have repeated this code 4 times in this method. That's not good.

AdminApi/test/AdminApi.Tests.Integration/Validators/JsonValidators.cs
```diff
@@ -1,28 +1,38 @@
 ﻿using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
-using Newtonsoft.Json.Schema;
+using NJsonSchema;
 using NJsonSchema.NewtonsoftJson.Generation;
 using JsonSchemaGenerator = NJsonSchema.Generation.JsonSchemaGenerator;
 
 namespace Backbone.AdminApi.Tests.Integration.Validators;
 
 public class JsonValidators
```
tnotheis, line 9: I just noticed the class name is plural for some reason.

AdminApi/test/AdminApi.Tests.Integration/Validators/JsonValidators.cs
tnotheis, line 1: I wonder why this class looks so much different than the corresponding one in the consumer api integration tests. Can you try to replace this implementation with the one from the consumer api and see if it works?

tnotheis (changes requested):

AdminApi/test/AdminApi.Tests.Integration/Validators/JsonValidators.cs
NikolaDmitrasinovic: Changed it, but there's another thing I now noticed. This class is placed in Validators folder in the AdminApi and in the Support folder in ConsumerApi? Should they both be in the Support or in the Validators (there is only one validator btw) folder?

tnotheis: Good catch. I leave it up to you which one to choose.

AdminApi/test/AdminApi.Tests.Integration/Validators/JsonValidators.cs
NikolaDmitrasinovic: Well, everything falls apart. I don't really understand it. It looks like some kind of a quick fix, but I can't figure out for what.

AdminApi/test/AdminApi.Tests.Integration/Validators/JsonValidators.cs
NikolaDmitrasinovic: I was a bit hesitant to mess too much with the original code, but this was really not good. Refactored.

AdminApi/test/AdminApi.Tests.Integration/Validators/JsonValidators.cs
tnotheis: Looking at the code, I would guess that it's a "fallback" for when the returned value is an array instead of an object. But there must be a way where you don't have to catch an exception.

tnotheis (approved):

PR #729, opened 2024-07-05
Ci: Validate licenses from referenced NuGet packages

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


.github/workflows/test.yml
```diff
@@ -426,3 +426,24 @@ jobs:
...
+      - name: Install nuget-licenses tool
+        run: dotnet tool install --global nuget-license
+      - name: Show Allowed Licenses
+        run: cat ./.ci/allowedLicenses.json
+      - name: Show Ignored Packages
+        run: cat ./.ci/ignoredPackages.json
+      - name: Restore Solution
+        run: dotnet restore ./Backbone.sln
+      - name: Validate Licenses
+        run: nuget-license -i ./Backbone.sln -a ./.ci/allowedLicenses.json -ignore ./.ci/ignoredPackages.json -o jsonPretty
```
tnotheis, line 449: Wouldn't the table output be nicer?

tnotheis, line 449: You should add the `--error-only` flag.

tnotheis, line 449: I usually prefer the descriptive flags instead of the abbreviations (e.g. `--input` instead of `-i`). This simplifies the interpretation of the command when you don't know the tool.
And I know that there are probably some commands out there where we don't use the descriptive flags yet. 😉

.github/workflows/test.yml
```diff
@@ -426,3 +426,24 @@ jobs:
         run: ./.ci/helm/buildChart.js
         env:
           VERSION: 0.0.0
+
+  validate-licenses:
+    name: Validate OpenSource Licenses
```
tnotheis, line 431:
```suggestion
    name: Validate Open Source Licenses
```

tnotheis (changes requested):
All in all, I'm quite happy with the results. :)

tnotheis (approved):

PR #734, opened 2024-07-10
Establishing a new relationship should not be possible before decomposition of the old one

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.


HunorTotBagi:
@NikolaDmitrasinovic Now the pull request is ready for review.

Dannyps:
When attempting to create a new relationship between two identities which have already established a relationship, the following error message is returned:

```json
{
	"error": {
		"id": "ERR4NGLGGXFH659UACK6",
		"code": "error.platform.validation.relationshipRequest.relationshipToTargetAlreadyExists",
		"message": "A relationship to 'did:e:localhost:dids:0f3e40164b6c495c28674f' already exists. If the relationship is **terminated**, you can reactivate it.",
		"docs": "",
		"time": "2024-07-11T07:50:39.657075Z",
		"data": null
	}
}
```
(emphasis added)

If one of the identities _terminates_ the relationship, the message is the same. I'm not sure this is correct.

tnotheis:
@Dannyps Sure, that's exactly what we expect. As long as no participant has decomposed the relationship, it can be reactivated.

You should check the task description to see a table with test cases. Make sure that all of them work as expected.

Dannyps:
## Manual Review

Upon testing, I have determined:
 - ✅ The creation of a second relationship between two identities cannot be done while there is already one present, unless its status is `ReadyForDeletion`, which is reached when both identities call the `Decompose` endpoint.
 - The creation of a second relationship between two identities cannot be done while there is already one present in one of the following statuses:
    - Pending
    - Active
    - Terminated
    - DeletionProposed

tnotheis:
> The creation of a second relationship between two identities cannot be done while there is already one present in one of the following statuses:
* [...]
* DeletionProposed

But this SHOULD actually be possible. It was the main goal of this task. @HunorTotBagi Did you not implement it?

Dannyps:
After reviewing the acceptance criteria, I have updated the review.

tnotheis:
> If the status is DeletionProposed, and the decomposer was P2, P1 cannot create a new relationship (this was what I tested initially).

@HunorTotBagi This seems to be a bug. Can you please check that?

tnotheis:
> @HunorTotBagi This seems to be a bug. Can you please check that?

Nvm. I misread Daniels report. :)

Dannyps:
Please consider the following scenario:

1. P1 creates relationship template `rlt1`
1. P2 creates relationship (using `rlt1`): Pending
1. P1 accepts: Active
1. P1 terminates: Terminated
1. P1 decomposes: DeletionProposed
--
6. P2 creates relationship template `rlt2`
1. P1 creates relationship (using `rlt2`): Pending
1. P2 rejects: Rejected
--
9. ❌ P2 tries to create relationship (using `rlt1`):  error: You haven't decomposed the old relationship.
1. ✔️ P1 creates relationship (using `rlt2`): Pending
---

**Step 8 should have failed.** Despite there being a Pending new relationship, the old one must be decomposed before proceeding:

P1 has decomposed old Relationship | P2 has decomposed old Relationship | Active Identity | POST Relationships | PUT (...)[Accept/Reject]| Fits the scenario
-- | -- | -- | -- | -- | --
true | true | P2 | true ⚠️| true ⚠️| up to step 5
true | true | P1 | true ⚠️| true ⚠️| up to step 5
true | false | P2 | false ✅ | false ❌| up to step 5, but bad step 8 ❌
true | false | P1 | true ✅ | true ⚠️| up to step 10
false | true | P2 | true ⚠️| true ⚠️| -
false | true | P1 | false ⚠️| false ⚠️| -
false | false | P2 | false ⚠️| false ⚠️| -
false | false | P1 | false ⚠️| false ⚠️| -

⚠️ - not tested in this scenario

Dannyps (changes requested):
As per my review above.

HunorTotBagi:
I can confirm that in this test case scenario that @Dannyps wrote, the Reject function isn't working as expected. I looked into it and found that it simply doesn't find all relationships for the active identity. I'm working on it.

tnotheis:
The runtime team decided they want this feature ater all.

PR #741, opened 2024-07-16
[AdminUI] Date filter text field acquires focus on interaction

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

--- 

This TextField is not a typical one and should not get focus on press. Instead the label will only move to the top when the field is filled.


tnotheis (approved):

tnotheis (changes requested):
Incorrect title format

tnotheis (approved):

PR #744, opened 2024-07-18
Unit Tests Next to Production Code - Relationships

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


Modules/Directory.Build.props
```diff
@@ -12,4 +12,21 @@
             CS8777,CS8819,CS8824,CS8825,CS8847
         </WarningsAsErrors>
     </PropertyGroup>
+
+    <ItemGroup Condition="'$(Configuration)' != 'Release' And !$(MSBuildProjectFile.Contains('Cleanup.csproj')) And !$(MSBuildProjectFile.Contains('SanityCheck.csproj')) And !$(MSBuildProjectFile.Contains('Devices.AdminCli.csproj'))">
+        <ProjectReference Include="..\..\..\..\BuildingBlocks\src\UnitTestTools\UnitTestTools.csproj" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(Configuration)' != 'Release' And !$(MSBuildProjectFile.Contains('Cleanup.csproj')) And !$(MSBuildProjectFile.Contains('SanityCheck.csproj')) And !$(MSBuildProjectFile.Contains('Devices.AdminCli.csproj'))">
```
Dannyps, line 20: Considering the two conditions are the same, couldn't you merge the two ItemGroups?

.filenesting.json
Dannyps, line 1: Is this intended?

Modules/Relationships/src/Relationships.Application/Relationships/Commands/RevokeRelationshipReactivation/Handler.Tests.cs
Dannyps, line 1: You should keep the named parameters as they help with code readability.

Modules/Directory.Build.props
tnotheis: The separation was on purpose, because you usually keep project and package references in different item groups.

.filenesting.json
daniel-almeida-konkconsulting: Yes, by having a solution-wide file nesting config file, it will get enabled by default and nest the test files.
Without, each person would have to enable/disable file nesting specifically, while choosing a configuration between Default and Web.

Modules/Relationships/src/Relationships.Application/Relationships/Commands/RevokeRelationshipReactivation/Handler.Tests.cs
daniel-almeida-konkconsulting: It was ReSharper's automatic formatting that removed it.
I'll add them back and ensure I turn that option off in my formatting profile.

Modules/Directory.Build.props
Dannyps: be that as it may, would there be a way of keeping the two item groups but providing the condition only once?

Dannyps (dismissed):

tnotheis:
@daniel-almeida-konkconsulting Did you compare the build times with the old structure compared to the new one?

tnotheis:
@daniel-almeida-konkconsulting Why is there only one reviewer for this PR? According to our new workflow, there should be one for a code review and one for a manual test.

NikolaVetnic:
Switching branches (e.g. from this feature to `main` or the other way around) sometimes does not "gather" all the tests under the relevant production code file. It also gets fixed on its own, I am not really sure what is causing the problem and could not reproduce it reliably so it may be a local issue. They do show up at all times though.

The tests fail if the production code is commented out.

`Rebuild` times comparison:
| Context | Feature Branch | Main Branch |
|--|--|--|
|`ConsumerApi`|03:35.769 minutes|01:05.063 minutes|
|`Solution`|05:55.007 minutes|02:32.772 minutes|
|`AdminApi`|52.678 seconds|34.589 seconds|

`bin` folder size comparison (`Debug` builds):
| Context | Feature Branch | Main Branch |
|--|--|--|
|`Relationships.Domain`|59.5 MB (62,480,384 bytes)|328 KB (335,872 bytes)|
|`Relationships.Application`|60.8 MB (63,827,968 bytes)|1.32 MB (1,384,448 bytes)|
|`Relationships.Infrastructure`|61.1 MB (64,143,360 bytes)|1.50 MB (1,581,056 bytes)|

Modules/Directory.Build.props
tnotheis: @daniel-almeida-konkconsulting ?

Modules/Directory.Build.props
```diff
@@ -12,4 +12,21 @@
             CS8777,CS8819,CS8824,CS8825,CS8847
         </WarningsAsErrors>
     </PropertyGroup>
+
+    <ItemGroup Condition="'$(Configuration)' != 'Release' And $(MSBuildProjectFile.Contains('Relationship'))">
```
tnotheis, line 16: We should add a comment explaining why we only add it for Relationship modules

tnotheis, line 16: Btw, shouldn't it be called "Relationship**s**"?
```suggestion
    <ItemGroup Condition="'$(Configuration)' != 'Release' And $(MSBuildProjectFile.Contains('Relationships'))">
```

Modules/Directory.Build.props
```diff
@@ -12,4 +12,21 @@
...
+        <ProjectReference Include="..\..\..\..\BuildingBlocks\src\UnitTestTools\UnitTestTools.csproj" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(Configuration)' != 'Release' And $(MSBuildProjectFile.Contains('Relationship'))">
```
tnotheis, line 20:
```suggestion
    <ItemGroup Condition="'$(Configuration)' != 'Release' And $(MSBuildProjectFile.Contains('Relationships'))">
```

Modules/Directory.Build.targets
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+    <ItemGroup Condition="'$(Configuration)' == 'Release' And !$(MSBuildProjectFile.Contains('Tests.csproj'))">
```
tnotheis, line 3: I could imagine that this condition causes the increased build times. 
For every single file in the solution, the compiler has to perform this check. I don't know how expensive such a check is during compilation, but it might be a possibility.

tnotheis (changes requested):

Modules/Directory.Build.props
daniel-almeida-konkconsulting: By looking at the docs (https://learn.microsoft.com/en-us/aspnet/web-forms/overview/deployment/web-deployment-in-the-enterprise/understanding-the-project-file#the-anatomy-of-a-project-file), there doesn't seem to be an aggregator we can use for the ItemGroups, so I would say we have to keep the duplicate condition.

Modules/Directory.Build.targets
tnotheis: To verify this, you could temporarily remove the check. While this will blow up the size of the dll, it should decrease the build time if my assumption is correct.
Btw: this also explains why the build time of the whole solution increases so drastically while the one for only the Consumer API only increases a little bit (the whole solution obviously has more files to check)

daniel-almeida-konkconsulting: Just did this check, build times are almost the exact same, so I'd say it's safe to assume this is not the culprit.

tnotheis: With "this check" you mean "removing the entire condition"?

daniel-almeida-konkconsulting: Yes, removed the `Directory.Build.targets` file.

tnotheis (dismissed):

tnotheis (approved):

PR #745, opened 2024-07-18
Fix copying of general `appsettings.override.json` file

NikolaVetnic:
# Readiness checklist

-   [X] I labeled the PR.

# Description

Due to work done on #740 the project on `main` branch could not even run, let alone run tests. The problem was in the Linux commands used in project files that should copy the joint `appsettings.override.json` file from the higher up the file tree to the project. This was remedied by reworking the commands like so:
```xml
    <Target Name="PreBuild" BeforeTargets="Build" Condition="$(Configuration) == Debug">
        <Delete Files="$(ProjectDir)appsettings.override.json" />
        <Copy SourceFiles="..\..\..\appsettings.override.json" DestinationFolder="$(ProjectDir)" />
    </Target>
```

The integration tests still don't run unfortunately. I am continuing to work on that.

The reason I am drafting this PR before repairing the integration tests is in case someone might need the halfway solution for their own work before I manage it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,7 +40,7 @@ jobs:
         uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: latest
+          node-version: "22.4.1"
           cache: "npm"
           cache-dependency-path: "**/package.json"
       - name: Cache npm dependencies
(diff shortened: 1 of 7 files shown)


tnotheis:
@MH321Productions This solution works locally? So you can
- run the integration tests
- run the solution with `dotnet run`

MH321Productions (approved):
I added the `UseHardlinksIfPossible` attribute to the `Copy` task to ensure that all appsettings.override.json files are synchronized.

tnotheis (approved):
Let's give it a try

PR #757, opened 2024-07-24
Replace value objects with strings

daniel-almeida-konkconsulting:
# Readiness checklist

-   [X] I labeled the PR.


MH321Productions:
After testing the endpoints, everything looks good

MH321Productions (dismissed):

Modules/Devices/src/Devices.Application/Clients/DTOs/ClientDTO.cs
```diff
@@ -17,6 +18,6 @@ public ClientDTO(OAuthClient client, int numberOfIdentities)
     public string DisplayName { get; set; }
     public string DefaultTier { get; set; }
     public DateTime CreatedAt { get; set; }
-    public int NumberOfIdentities { get; set; }
+    public int? NumberOfIdentities { get; set; }
```
Dannyps, line 21: Why?

Dannyps (commented):
You repeat this a lot:
```cs
validationResult.ShouldHaveValidationErrorFor(x => x.IdentityAddress);
validationResult.Errors.Should().HaveCount(1);
validationResult.Errors.First().ErrorCode.Should().Be("error.platform.validation.invalidPropertyValue");
validationResult.Errors.First().ErrorMessage.Should().Be("The ID is not valid. Check length, prefix and the used characters.");
```

Wouldn't it be possible to implement the same using an extension method?

BuildingBlocks/src/BuildingBlocks.Application/FluentValidation/ValueInValidator.cs
```diff
@@ -59,11 +62,28 @@ protected override string GetDefaultMessageTemplate(string errorCode)
...
         params TProperty[] validOptions)
     {
         return ruleBuilder
             .SetValidator(new ValueInValidator<T, TProperty>(validOptions))
             .WithErrorCode(GenericApplicationErrors.Validation.InvalidPropertyValue().Code);
     }
+
+    public static IRuleBuilderOptions<T, string?> ValidId<T, TId>(this IRuleBuilder<T, string?> ruleBuilder) where TId : StronglyTypedId
+    {
+        var method = IS_VALID_CACHE.FirstOrDefault(x => x.Key == typeof(TId)).Value;
```
tnotheis, line 77: Can't you access a ConcurrentDictionary with an indexer?

```suggestion
        var method = IS_VALID_CACHE[x.Key];
```

BuildingBlocks/src/BuildingBlocks.Application/FluentValidation/ValueInValidator.cs
```diff
@@ -59,11 +62,28 @@ protected override string GetDefaultMessageTemplate(string errorCode)
...
+        if (method == null)
+        {
+            method = typeof(TId).GetMethod("IsValid")!;
```
tnotheis, line 80:
```suggestion
            method = typeof(TId).GetMethod("IsValid", BindingFlags.Public | BindingFlags.Static)!;
```

tnotheis, line 80: You shouldn't suppress null. Instead, make a null check and throw a descriptive error message, informing the developer about the missing method.

BuildingBlocks/src/UnitTestTools/FluentValidation/Extensions.cs
```diff
@@ -31,3 +33,15 @@ public static class IValidatorExtensions
         return validationResult.Errors;
     }
 }
+
+public static class ValidationTestExtensions
+{
+    public static void ShouldHaveValidationErrorForItem<T>(this TestValidationResult<T> testValidationResult, string propertyName, string expectedErrorCode,
+        string expectedErrorMessage)
+    {
+        var validationError = testValidationResult.Errors.FirstOrDefault(r => r.PropertyName.Contains(propertyName)) ?? throw new XunitException($"Expected error for property '{propertyName}'.");
```
tnotheis, line 42: I don't think that `FirstOrDefault` is a good idea. If there is more than one error for the property, the assertion might fail even though there IS an error that matches the expected criteria

Modules/Challenges/test/Challenges.Application.Tests/Tests/Challenges/Commands/DeleteChallengesOfIdentity/ValidatorTests.cs
```diff
@@ -0,0 +1,40 @@
...
+        var validator = new Validator();
+
+        // Act
+        var validationResult = validator.TestValidate(new DeleteChallengesOfIdentityCommand("invalid-identity-address"));
+
+        // Assert
+        validationResult.ShouldHaveValidationErrorForItem(
+            propertyName: nameof(DeleteChallengesOfIdentityCommand.IdentityAddress),
+            expectedErrorCode: "error.platform.validation.invalidPropertyValue",
+            expectedErrorMessage: "The ID is not valid. Check length, prefix and the used characters.");
```
tnotheis, line 38: Since the error code and message are always the same, your current solution is unnecessarily verbose, imo. You might as well do this:

```suggestion
        validationResult.ShouldHaveValidationErrorDueToInvalidId(propertyWithInvalidId: nameof(DeleteChallengesOfIdentityCommand.IdentityAddress));
```

Modules/Devices/src/Devices.Application/Devices/Queries/ListDevices/Handler.cs
```diff
@@ -10,23 +9,17 @@ namespace Backbone.Modules.Devices.Application.Devices.Queries.ListDevices;
...
     public async Task<ListDevicesResponse> Handle(ListDevicesQuery request, CancellationToken cancellationToken)
     {
-
-        var dbPaginationResult = await _identitiesRepository.FindAllDevicesOfIdentity(_activeIdentity, request.Ids, request.PaginationFilter, cancellationToken);
-
-        var items = _mapper.Map<DeviceDTO[]>(dbPaginationResult.ItemsOnPage);
-
-        return new ListDevicesResponse(items, request.PaginationFilter, dbPaginationResult.TotalNumberOfItems);
+        var dbPaginationResult = await _identitiesRepository.FindAllDevicesOfIdentity(_activeIdentity, request.Ids.Select(DeviceId.Parse), request.PaginationFilter, cancellationToken);
+        return new ListDevicesResponse(dbPaginationResult.ItemsOnPage.Select(d => new DeviceDTO(d)), request.PaginationFilter, dbPaginationResult.TotalNumberOfItems);
```
tnotheis, line 23: Let the `ListDevicesResponse` do the mapping (same for all the other cases across the codebase)
```suggestion
        return new ListDevicesResponse(dbPaginationResult, request.PaginationFilter);
```

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcessId.cs
```diff
@@ -21,6 +21,16 @@ public static IdentityDeletionProcessId Generate()
         return new IdentityDeletionProcessId(PREFIX + randomPart);
     }
 
+    public static bool IsValid(string stringValue)
+    {
+        return UTILS.IsValid(stringValue);
+    }
```
tnotheis, line 27: Is it possible to write an ArchUnit test that asserts that a type inheriting from StronglyTypedId has a static method called `IsValid` with a single parameter of type `string`?

Modules/Files/src/Files.ConsumerApi/Controllers/FilesController.cs
```diff
@@ -39,6 +38,8 @@ public FilesController(IMediator mediator, IOptions<ApplicationOptions> options)
     [ProducesError(StatusCodes.Status400BadRequest)]
     public async Task<IActionResult> UploadFile([FromForm] CreateFileDTO dto, CancellationToken cancellationToken)
     {
+        //TODO model state is dto valid
```
tnotheis, line 41: What's that TODO?

Modules/Synchronization/src/Synchronization.Application/Datawallets/DTOs/PushDatawalletModificationItem.cs
```diff
@@ -1,11 +1,12 @@
 using Backbone.BuildingBlocks.Application.FluentValidation;
+using Backbone.Modules.Synchronization.Domain.Entities;
 using FluentValidation;
 
 namespace Backbone.Modules.Synchronization.Application.Datawallets.DTOs;
 
 public class PushDatawalletModificationItem
 {
-    public required DatawalletModificationDTO.DatawalletModificationType Type { get; set; }
+    public required DatawalletModificationType Type { get; set; }
```
tnotheis, line 9: Using a separate enum here was intended. The reason is that we don't want the public API to change when we change an enum of our Domain project.

Modules/Synchronization/src/Synchronization.Application/SyncRuns/Commands/StartSyncRun/Handler.cs
```diff
@@ -150,12 +147,10 @@ private long DetermineNextSyncRunIndex()
...
     private StartSyncRunResponse CreateResponse(StartSyncRunStatus status, SyncRun? newSyncRun = null)
     {
-        var syncRunDTO = _mapper.Map<SyncRunDTO>(newSyncRun);
-
         var response = new StartSyncRunResponse
         {
             Status = status,
-            SyncRun = syncRunDTO
+            SyncRun = newSyncRun != null ? new SyncRunDTO(newSyncRun) : null
         };
```
tnotheis, line 154: Extract the mapping into the constructor:

```c#
var response = new StartSyncRunResponse(status, newSyncRun);
```

Modules/Synchronization/src/Synchronization.Application/SyncRuns/Commands/StartSyncRun/StartSyncRunCommand.cs
```diff
@@ -1,24 +1,24 @@
...
-    public StartSyncRunCommand(SyncRunDTO.SyncRunType type, ushort? duration, ushort supportedDatawalletVersion)
+    public StartSyncRunCommand(SyncRun.SyncRunType type, ushort? duration, ushort supportedDatawalletVersion)
     {
         Type = type;
         Duration = duration;
         SupportedDatawalletVersion = supportedDatawalletVersion;
     }
 
-    public SyncRunDTO.SyncRunType Type { get; set; }
+    public SyncRun.SyncRunType Type { get; set; }
```
tnotheis, line 21: Should be a separate enum

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Application/FluentValidation/ValueInValidator.cs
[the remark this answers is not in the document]
daniel-almeida-konkconsulting: This file already held the `ValidatorExtensions` class.
But I agree, I'll promote it to a separate file under the `Extensions` folder.

BuildingBlocks/src/BuildingBlocks.Application/FluentValidation/ValueInValidator.cs
daniel-almeida-konkconsulting: I think you meant:
`var method = IS_VALID_CACHE[typeof(TId)];`
I could, but this throws when there are no matches, so we would be replacing the `FirstOrDefault` with a subsequent `null` check for a `try / catch` block where we use the indexer.
Thoughts?

BuildingBlocks/src/BuildingBlocks.Application/FluentValidation/ValueInValidator.cs
daniel-almeida-konkconsulting: Do you still want this if we are going to add an `ArchUnit` test that ensures all `StronglyTypedIds` have an `IsValid` method?

tnotheis: I'd say: why not?

BuildingBlocks/src/BuildingBlocks.Application/FluentValidation/ValueInValidator.cs
tnotheis: True. Then use `TryGetValue` instead:

```c#
if(IS_VALID_CACHE.TryGetValue(x.Key, out var method)
```

BuildingBlocks/src/BuildingBlocks.Application/FluentValidation/ValueInValidator.cs
daniel-almeida-konkconsulting: Roger that.

Modules/Challenges/test/Challenges.Application.Tests/Tests/Challenges/Commands/DeleteChallengesOfIdentity/ValidatorTests.cs
daniel-almeida-konkconsulting: It's not always the same, but I created a couple extension methods to avoid repetition of the error messages and codes.

BuildingBlocks/src/UnitTestTools/FluentValidation/Extensions.cs
tnotheis: Why not

```c#
errorsForProperty.Should().Contain(r => r.ErrorCode == expectedErrorCode && r.ErrorMessage == expectedErrorMessage);
```
?

Modules/Devices/src/Devices.Domain/Entities/Identities/IdentityDeletionProcessId.cs
tnotheis: What about this?

Modules/Files/src/Files.ConsumerApi/Controllers/FilesController.cs
tnotheis: What about this?

MH321Productions (dismissed):

Modules/Files/src/Files.ConsumerApi/Controllers/FilesController.cs
```diff
@@ -38,7 +39,9 @@ public FilesController(IMediator mediator, IOptions<ApplicationOptions> options)
     [ProducesError(StatusCodes.Status400BadRequest)]
     public async Task<IActionResult> UploadFile([FromForm] CreateFileDTO dto, CancellationToken cancellationToken)
     {
-        //TODO model state is dto valid
+        var validationResult = await new CreateFileDTOValidator().ValidateAsync(dto, cancellationToken);
+        if (!validationResult.IsValid)
+            return BadRequest(validationResult.Errors);
```
tnotheis, line 44: That's absolutely not what we want errors returned by our APIs to look like.

BuildingBlocks/src/UnitTestTools/FluentValidation/Extensions.cs
```diff
@@ -36,12 +36,30 @@ public static IEnumerable<ValidationFailure> ShouldHaveValidationErrorForItemWit
...
-    public static void ShouldHaveValidationErrorForItem<T>(this TestValidationResult<T> testValidationResult, string propertyName, string expectedErrorCode,
-        string expectedErrorMessage)
+    public static void ShouldHaveValidationErrorForItem<T>(this TestValidationResult<T> testValidationResult, string propertyName, string expectedErrorCode, string expectedErrorMessage)
     {
-        var validationError = testValidationResult.Errors.FirstOrDefault(r => r.PropertyName.Contains(propertyName)) ?? throw new XunitException($"Expected error for property '{propertyName}'.");
+        var errorsForProperty = testValidationResult.ShouldHaveValidationErrorFor(propertyName);
+        errorsForProperty.Should().Contain(r => r.ErrorCode == expectedErrorCode && r.ErrorMessage == expectedErrorMessage);
+    }
+
+    public static void ShouldHaveValidationErrorForItemInCollection<T>(this TestValidationResult<T> testValidationResult, string collectionWithInvalidId, int indexWithInvalidId,
```
tnotheis, line 45: This method isn't used, as far as I can see

daniel-almeida-konkconsulting: Thought it might add some value having an extension method such as this at the ready, considering the other 3 existing ones.
But I'll remove it if you think it doesn't make sense to have this in advance.

Modules/Files/src/Files.ConsumerApi/Controllers/FilesController.cs
daniel-almeida-konkconsulting: Updated to throw a `ValidationException` as the Mediatr Pipeline Behavior does.

tnotheis: Did you test it?

tnotheis (commented):
As you suggested via DM: write an ArchUnit test that makes sure we don't use strongly typed ids in our commands and queries.

Modules/Files/src/Files.ConsumerApi/Controllers/FilesController.cs
daniel-almeida-konkconsulting: image

tnotheis: The error in your screenshot does not come from the `CreateFileDTOValidator`.

daniel-almeida-konkconsulting: Uhm, it does, from the first rule on the `Owner` field which validates the `IdentityAddress`

tnotheis (approved):

PR #777, opened 2024-07-29
Remove redundant parameter from PnsConnector's Send method

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


Modules/Devices/test/Devices.Infrastructure.Tests/Tests/PushNotifications/Connectors/Sse/ServerSentEventsConnectorTests.cs
```diff
@@ -52,41 +52,22 @@ public async Task Uses_the_SseServerClient_to_send_the_event()
...
-
-        var mockSseServerClient = A.Fake<ISseServerClient>();
-
-        var sseConnector = CreateConnector(mockSseServerClient);
-
-        // Act
-        await sseConnector.Send([CreatePnsRegistrationForSse()], recipient, new TestPushNotification());
-
-        // Assert
-        A.CallTo(() => mockSseServerClient.SendEvent(A<string>._, "Test")).MustHaveHappenedOnceExactly();
```
tnotheis, line 75: Explanation: I combined this test with the one before

Modules/Devices/src/Devices.Infrastructure/PushNotifications/Connectors/Dummy/DummyConnector.cs
```diff
@@ -1,43 +1,43 @@
...
+        foreach (var registration in registrations)
         {
-            var (title, body) = await _notificationTextProvider.GetNotificationTextForDeviceId(notification.GetType(), device.Id);
-            _logger.LogInformation("Sending push notification to device with id '{deviceId}' of identity with address '{recipient}': {title}, {body}.", recipient, device.Id, title, body);
+            sendResults.AddSuccess(registration.DeviceId);
         }
 
-        return sendResults;
+        return Task.FromResult(sendResults);
     }
```
NikolaVetnic, line 28: So just to clarify in order to see if I understand the code correctly - since this is a dummy we are not really sending anything, but instead are adding a success to results for each of the registrations. Correct?

NikolaVetnic (approved):

Modules/Devices/src/Devices.Infrastructure/PushNotifications/Connectors/Dummy/DummyConnector.cs
tnotheis: Correct

PR #778, opened 2024-07-29
Consumer API: Create Metrics for the QueuedForDeletion Tier

NikolaVetnic:
# Readiness checklist

-   [X] I labeled the PR.

# Description

It was noticed that the `QUEUED_FOR_DELETION` does not have its associated metrics seeded on project start up. `QuotasDbContextSeeder` was introduce to seed the necessary metrics and the `SeedQueuedForDeletionTierCommand` was modified to take into account that the Tier itself is already created by the `DevicesDbContextSeeder`, thus only adding the associated metrics once the Tier is actually found in the database.


ConsumerApi/DevicesDbContextSeeder.cs
```diff
@@ -5,6 +5,7 @@
 using Backbone.Modules.Devices.Application.Users.Commands.SeedTestUsers;
 using Backbone.Modules.Devices.Domain.Aggregates.Tier;
 using Backbone.Modules.Devices.Infrastructure.Persistence.Database;
+using Backbone.Modules.Quotas.Application.Tiers.Commands.SeedQueuedForDeletionTier;
```
NikolaDmitrasinovic, line 8: line 8 can be removed

NikolaDmitrasinovic (changes requested):

ConsumerApi/DevicesDbContextSeeder.cs
NikolaVetnic: Changes made.

tnotheis (approved):

NikolaDmitrasinovic (approved):

PR #784, opened 2024-07-31
Admin UI: Upgrade dependencies

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Applications/AdminUi/apps/admin_ui/pubspec.yaml
+++ b/Applications/AdminUi/apps/admin_ui/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
 dependencies:
   admin_api_sdk: ^1.0.0
   admin_api_types: ^1.0.0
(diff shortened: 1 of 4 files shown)


Applications/AdminUi/apps/admin_ui/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
 dependencies:
   admin_api_sdk: ^1.0.0
   admin_api_types: ^1.0.0
-  collection: ^1.18.0
+  collection: any
```
tnotheis, line 13: Why don't we pin the version?

jkoenig134: There is another package (afaik test) that depends on collection. As we cannot bump it and we dont care for the version any is the solution!

tnotheis (approved):

PR #801, opened 2024-08-09
Determine the sequence of migrations across different modules

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


daniel-almeida-konkconsulting:
I have manually tested this, during which process I encountered a specific behavior with the initial migrations, which were not rolled back upon failure.
I've discussed this with @MH321Productions, who mentioned this is in accordance with EF Core's documentation and is expected behavior.
As such, I approve the current changes.

daniel-almeida-konkconsulting (dismissed):

NikolaDmitrasinovic (dismissed):

tnotheis:
> I have manually tested this, during which process I encountered a specific behavior with the initial migrations, which were not rolled back upon failure. I've discussed this with @MH321Productions, who mentioned this is in accordance with EF Core's documentation and is expected behavior. As such, I approve the current changes.

Can you clarify exactly what you mean? Because afaik everything should be rolled back in case of an error.

tnotheis:
@MH321Productions Are you aware of the failing pipelines?

MH321Productions:
> Can you clarify exactly what you mean? Because afaik everything should be rolled back in case of an error.

The documentation states for the `IMigrator.MigrateAsync` method that the latest migration is applied if no target migration is given. A method for clearing the first migration doesn't exist (AFAIK). The only way to entirely clear your DB is, according to the Microsoft docs, to delete all migrations and drop the database.

tnotheis:
> > Can you clarify exactly what you mean? Because afaik everything should be rolled back in case of an error.
> 
> The documentation states for the `IMigrator.MigrateAsync` method that the latest migration is applied if no target migration is given. A method for clearing the first migration doesn't exist (AFAIK). The only way to entirely clear your DB is, according to the Microsoft docs, to delete all migrations and drop the database.

@daniel-almeida-konkconsulting already answered me via DM. I didn't find a way to clear the latest migration either when I implemented the DatabaseMigrator.

daniel-almeida-konkconsulting (dismissed):

NikolaDmitrasinovic (approved):

daniel-almeida-konkconsulting (approved):

tnotheis (approved):

PR #802, opened 2024-08-12
Consumer API: Add missing majorVersion result wrapper

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.


Applications/ConsumerApi/src/Controllers/VersionController.cs
tnotheis, line 1: Please check out how we do it in all the other controllers. And then do it similarly.

tnotheis (changes requested):

tnotheis (approved):

PR #808, opened 2024-08-21
Check whether a relationship can be established with a given identity

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/CanEstablishRelationshipQuery.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Backbone.DevelopmentKit.Identity.ValueObjects;
+using MediatR;
+
+namespace Backbone.Modules.Relationships.Application.Relationships.Queries.CanEstablishRelationship;
+
+public class CanEstablishRelationshipQuery : IRequest<CanEstablishRelationshipResponse>
+{
+    public required IdentityAddress PeerAddress { get; set; }
```
daniel-almeida-konkconsulting, line 8: We do not want to use value objects in Mediatr Commands and Queries anymore.
This PR would benefit from https://github.com/nmshd/backbone/pull/757 to add the Validator with the `ValidId` extension method call.
I would say, use `string` and add a `Validator` regardless, as this might be merged before the linked PR.

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.Tests.cs
daniel-almeida-konkconsulting, line 1: I think these Handler tests fall under the description of "Unnecessary Handler Tests".
These should all instead be Integration tests.

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
```diff
@@ -0,0 +1,38 @@
...
+    public Handler(IUserContext userContext, IRelationshipsRepository relationshipsRepository)
+    {
+        _relationshipsRepository = relationshipsRepository;
+        _userContext = userContext;
+    }
+
+    public async Task<CanEstablishRelationshipResponse> Handle(CanEstablishRelationshipQuery request, CancellationToken cancellationToken)
+    {
+        var identity = _userContext.GetAddress();
+        var peer = request.PeerAddress;
```
daniel-almeida-konkconsulting, line 28: I would simply inline this as I don't believe it contributes to better readability.

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
```diff
@@ -0,0 +1,38 @@
...
+        var relationships = await _relationshipsRepository.FindRelationships(BetweenParticipants(identity, peer), cancellationToken);
+
+        return relationships.Any(relationship => FORBIDDEN_STATUSES.Contains(relationship.Status)) ? FALSE : TRUE;
```
daniel-almeida-konkconsulting, line 31: Maybe an Expression for this would be a good idea, especially considering we already do a similar check in the `Relationship` entity, specifically in the `EnsureNoOtherRelationshipToPeerExists` method.

daniel-almeida-konkconsulting, line 31: Actually, isn't there already a method `RelationshipBetweenTwoIdentitiesExists` in the repository that we can use?

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    private static Expression<Func<Relationship, bool>> BetweenParticipants(IdentityAddress a, IdentityAddress b)
+    {
+        return r => (r.From == a && r.To == b) || (r.From == b && r.To == a);
+    }
```
daniel-almeida-konkconsulting, line 37: There's already a queryable extension method for this in `RelationshipQueryableExtensions`

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
```diff
@@ -195,6 +197,17 @@ public async Task<IActionResult> DecomposeRelationship([FromRoute] string id, Ca
         return Ok(response);
     }
 
+    [HttpGet("CanCreate")]
+    [ProducesResponseType(typeof(HttpResponseEnvelopeResult<CanEstablishRelationshipResponse>), StatusCodes.Status200OK)]
+    [ProducesError(StatusCodes.Status400BadRequest)]
+    public async Task<IActionResult> CanEstablishRelationship([FromQuery] IdentityAddress peer, CancellationToken cancellationToken)
```
daniel-almeida-konkconsulting, line 203: Use `string` for `peer` query parameter

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
```diff
@@ -195,6 +197,17 @@ public async Task<IActionResult> DecomposeRelationship([FromRoute] string id, Ca
...
+    {
+        if (peer == null!) throw new ValidationException(GenericApplicationErrors.Validation.InvalidPropertyValue(nameof(peer)));
```
daniel-almeida-konkconsulting, line 205: You can remove this after adding the Validator to the Mediatr Query

daniel-almeida-konkconsulting (changes requested):

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
MH321Productions: I actually used this in the beginning, but this performs another check (the `Relationship.CountsAsActive` method), which returns true, when the relationship has the `ReadyForDeletion` status, which contradicts the Acceptance criteria.

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
MH321Productions: I know, but I can't use it, since I can't access the private `DbSet`

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
MH321Productions: Done

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
MH321Productions: Done

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
MH321Productions: Done

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.Tests.cs
MH321Productions: Ok, this will take a while. Never done this before

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/CanEstablishRelationshipQueryValidator.cs
```diff
@@ -0,0 +1,13 @@
...
+using FluentValidation;
+
+namespace Backbone.Modules.Relationships.Application.Relationships.Queries.CanEstablishRelationship;
+
+// ReSharper disable once UnusedType.Global
+public class CanEstablishRelationshipQueryValidator : AbstractValidator<CanEstablishRelationshipQuery>
+{
+    public CanEstablishRelationshipQueryValidator()
+    {
+        RuleFor(q => q.PeerAddress).DetailedNotEmpty();
```
daniel-almeida-konkconsulting, line 11: `RuleFor(q => q.PeerAddress).Must(IdentityAddress.IsValid);`

daniel-almeida-konkconsulting (changes requested):

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.Tests.cs
MH321Productions: done

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/CanEstablishRelationshipQueryValidator.cs
MH321Productions: done

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
MH321Productions: done

HunorTotBagi:
After manual testing I can confirm that everything is working as expected.

| Relationship status | `canCreate` |
|----------|----------|
| `Rejected`    | `true` ✅   |
| `Revoked`    | `true` ✅   |
| `ReadyForDeletion`    | `true` ✅  |
| `Pending`    | `false` ✅   |
| `Active`    | `false` ✅   |
| `Terminated`    | `false` ✅ |
| `DeletionProposed`    | `false` ✅    |

✅ - behaves as expected

HunorTotBagi (dismissed):

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/CanEstablishRelationshipQueryValidator.cs
daniel-almeida-konkconsulting: You can remove the rule for "DetailedNotEmpty" as it's now redundant

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        _relationshipsRepository = relationshipsRepository;
+        _userContext = userContext;
+    }
+
+    public async Task<CanEstablishRelationshipResponse> Handle(CanEstablishRelationshipQuery request, CancellationToken cancellationToken)
+    {
+        var hasActiveRelationship = await _relationshipsRepository.RelationshipBetweenTwoIdentitiesExists(_userContext.GetAddress(), IdentityAddress.Parse(request.PeerAddress), cancellationToken);
+
+        return hasActiveRelationship ? FALSE : TRUE;
```
daniel-almeida-konkconsulting, line 26: Might be a matter of preference, but I would remove the static responses and just use: 
`return new CanEstablishRelationshipResponse { CanCreate = !hasActiveRelationship };`

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
daniel-almeida-konkconsulting, line 1: There a couple of other Relationship statuses that aren't tested.
Is this because of limitations in terms of settings Relationships in those statuses?

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
```diff
@@ -65,6 +70,26 @@ public async Task GivenAPendingRelationshipBetweenI1AndI2CreatedByI1()
         _relationshipId = createRelationshipResponse.Result!.Id;
     }
 
+    [Given("an active Relationship between i1 and i2 created by i1")]
+    public async Task GivenAnActiveRelationshipBetweenI1AndI2()
+    {
+        var relationshipTemplateResponse = await CreateRelationshipTemplate(_client1);
+        var createRelationshipResponse = await CreateRelationship(_client2, relationshipTemplateResponse.Result!.Id);
+        var acceptRelationshipResponse = await AcceptRelationship(_client1, createRelationshipResponse.Result!.Id);
```
daniel-almeida-konkconsulting, line 78: There's a method in the `Utils.cs` class for establishing a relationship you can use `EstablishRelationshipBetween`.
Maybe even build on top of that method to establish Relationships in a given status (e.g., Accepted, Reject, etc.)

daniel-almeida-konkconsulting (changes requested):

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
MH321Productions: I tested the easiest statuses. And since the statuses of the relationships are checked in groups, it should be safe to assume that it will work for the other ones as well (especially because they itself are tested extensively).

Modules/Relationships/src/Relationships.Application/Relationships/Queries/CanEstablishRelationship/Handler.cs
MH321Productions: done

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
MH321Productions: done

daniel-almeida-konkconsulting: The step definition for `GivenARejectedRelationshipBetweenI1AndI2` could also benefit from this

MH321Productions: So should I just create a new method like `CreateRejectedRelationshipBetween` and then use it?

daniel-almeida-konkconsulting: Either that or change the current one to something like `CreateRelationshipInStatus` and add support for various ones you need.

MH321Productions: done

daniel-almeida-konkconsulting (approved):

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
```diff
@@ -0,0 +1,38 @@
+﻿@Integration
+Feature: GET Relationship
```
tnotheis, line 2:
```suggestion
Feature: GET Relationships/CanCreate
```

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
```diff
@@ -0,0 +1,38 @@
...
+    
+Scenario: Two identities without a relationship can create one
+    Given Identities i1 and i2
+    When a GET request is sent to the /Relationships/CanCreate?peer={i.id} endpoint by i1 for i2
+    Then the response status code is 200 (OK)
+    And a relationship can be established
+    
+Scenario: Two identities with a valid relationship can't create one
```
tnotheis, line 10:
```suggestion
Scenario: Two identities with an active relationship can't create another one
```

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
```diff
@@ -0,0 +1,38 @@
...
+    And a relationship can not be established
+    
+Scenario: Two identities with a rejected relationship can create one
+    Given Identities i1 and i2
+    And a rejected Relationship between i1 and i2 created by i1
+    When a GET request is sent to the /Relationships/CanCreate?peer={i.id} endpoint by i1 for i2
+    Then the response status code is 200 (OK)
+    And a relationship can be established
+
+Scenario: Two identities with a rejected and an active relationship can't create one
```
tnotheis, line 24: Does this scenario deliver any additional value compared to the first and second one?

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
```diff
@@ -0,0 +1,38 @@
...
+    Given Identities i1 and i2
+    And a rejected Relationship between i1 and i2 created by i1
+    And an active Relationship between i1 and i2 created by i1
+    When a GET request is sent to the /Relationships/CanCreate?peer={i.id} endpoint by i1 for i2
+    Then the response status code is 200 (OK)
+    And a relationship can not be established
+    
+Scenario: Two identities with two rejected relationships can create one
```
tnotheis, line 32: Does this scenario deliver any additional value compared to the second one?

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Helpers/Utils.cs
```diff
@@ -45,6 +45,39 @@ public static async Task<Relationship> EstablishRelationshipBetween(Client clien
         return getRelationshipResponse.Result!;
     }
 
+    public static async Task<Relationship> CreateRejectedRelationshipBetween(Client client1, Client client2)
```
tnotheis, line 48: You should extract common code from this method and `EstablishRelationshipBetween`. The extracted method could be called `CreatePendingRelationshipBetween`

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
```diff
@@ -115,6 +137,16 @@ public async Task WhenAPostRequestIsSentToTheRevokeRelationshipEndpointByI2()
         _revokeRelationshipResponse = await _client1.Relationships.RevokeRelationship(_relationshipId, revokeRelationshipRequest);
     }
 
+    [When("a GET request is sent to the /Relationships/CanCreate\\?peer={i.id} endpoint by i1 for i2")]
+    public async Task WhenAGetRequestIsSentToTheCanCreateEndpointByI1()
+    {
+        _canEstablishResponse = await _client1.Relationships.CanCreateRelationship(_client2.IdentityData?.Address ?? "");
```
tnotheis, line 143: I think in this case (since it's test code, and not production code) I would prefer the potential NullReferenceException over a silent fallback to an empty string.
```suggestion
        _canEstablishResponse = await _client1.Relationships.CanCreateRelationship(_client2.IdentityData!.Address);
```

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
```diff
@@ -187,6 +219,23 @@ public async Task ThenTheResponseContainsARelationship()
         }
     }
 
+    [Then("a relationship can be established")]
+    public void ThenARelationshipCanBeEstablished() => PerformCanEstablishRelationshipCheck(true);
```
tnotheis, line 223: Boolean parameters are really ugly. It's so hard to understand what they are used for.

I think I'd rather duplicate the code than having the boolean parameter.

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
```diff
@@ -187,6 +219,23 @@ public async Task ThenTheResponseContainsARelationship()
...
+
+    [Then("a relationship can not be established")]
+    public void ThenARelationshipCanNotBeEstablished() => PerformCanEstablishRelationshipCheck(false);
+
+    private void PerformCanEstablishRelationshipCheck(bool value)
+    {
+        if (_canEstablishResponse != null)
+        {
+            _canEstablishResponse!.Should().BeASuccess();
```
tnotheis, line 232: This assertion is redundant, right? Because in another `Then` step we check for the status code.

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
```diff
@@ -207,4 +256,13 @@ private async Task<ApiResponse<RelationshipMetadata>> CreateRelationship(Client
...
     }
+
+    private async Task<ApiResponse<RelationshipMetadata>> RejectRelationship(Client client, string relationshipId)
+    {
+        var rejectRelationshipRequest = new RejectRelationshipRequest
+        {
+            CreationResponseContent = "AAA".GetBytes()
+        };
+        return await client.Relationships.RejectRelationship(relationshipId, rejectRelationshipRequest);
+    }
```
tnotheis, line 267: Is this even used?

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
MH321Productions: I would say yes, since the two You mentioned test, if the handler can correctly detect multiple relationship entries.

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/GET.feature
MH321Productions: See above

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/RelationshipsStepDefinitions.cs
MH321Productions: No, forgot to remove it after using the new helper method `CreateRejectedRelationshipBetween`

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Helpers/Utils.cs
MH321Productions: done

tnotheis (approved):

PR #809, opened 2024-08-21
Consumer API: Correlation IDs are retained across domain event boundaries

NikolaVetnic:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Correlation IDs are propagated across domain event boundaries by means of using the `X-CORRELATION-ID` header if one is present in the request, or by generating a random one by Serilog if the request contains none.


NikolaVetnic:
Notes for manual testing: 
- Creating a message is a nice endpoint to use while testing this,
- Observe the EventHandler and ConsumerAPI terminal windows when running the whole solution, the Correlation IDs in both should of course match,
- Use request with and without setting the `X-CORRELATION-ID`.

NikolaVetnic:
Please note that Google Cloud Pub Sub and Azure Service Bus implementations *have not been tested*, neither manually nor automatically. Tests present in the solution for these two implementations are skipped due to the lack of emulator, and I did not have access to the infrastructure needed to deploy these and thus utilize the two stated event buses.

tnotheis (dismissed):

tnotheis:
Sorry for that spam. :D

tnotheis (approved):

PR #837, opened 2024-09-03
Identity deletion jobs fail due to InvalidOperationException when trying to send push notifications to multiple devices at once

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

I had to un-parallelize sending push notifications, because the `NotificationTextProvider` sends queries to the database, and EF Core does not allow multiple parallel queries - not even when trying to not track the entities. 😢 
I also got rid of the parallelization in the `ListQuotasForIdentityCommand` handler for the same reason.

Let's see if this leads to any notable performance decrease.

### DIFF
--- a/Modules/Tokens/src/Tokens.Application/Tokens/DTOs/TokenDTO.cs
+++ b/Modules/Tokens/src/Tokens.Application/Tokens/DTOs/TokenDTO.cs
@@ -12,6 +12,7 @@ public class TokenDTO
         CreatedAt = token.CreatedAt;
         ExpiresAt = token.ExpiresAt;
         Content = token.Content;
+        ForIdentity = token.ForIdentity?.Value;
     }
 
     public string Id { get; set; }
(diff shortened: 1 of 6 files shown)


ericbrunner (approved):

ericbrunner (approved):

Sdks/ConsumerApi.Sdk/src/Endpoints/Tokens/Types/Token.cs
```diff
@@ -7,5 +7,6 @@ public class Token
     public required string CreatedByDevice { get; set; }
     public required DateTime CreatedAt { get; set; }
     public required DateTime ExpiresAt { get; set; }
+    public string? ForIdentity { get; set; }
```
ericbrunner: Never used. Obsolete?

Modules/Tokens/src/Tokens.Application/Tokens/DTOs/TokenDTO.cs
```diff
@@ -12,6 +12,7 @@ public TokenDTO(Token token)
         CreatedAt = token.CreatedAt;
         ExpiresAt = token.ExpiresAt;
         Content = token.Content;
+        ForIdentity = token.ForIdentity?.Value;
```
ericbrunner: Should be removed. Never used

ericbrunner (commented):
done

PR #838, opened 2024-09-03
Error in Identity Deletion Jobs when using Google Cloud Pub/Sub

tnotheis:
…SubPersisterConnection

# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/GoogleCloudPubSub/DefaultGoogleCloudPubSubPersisterConnection.cs
+++ b/BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/GoogleCloudPubSub/DefaultGoogleCloudPubSubPersisterConnection.cs
@@ -2,16 +2,19 @@ using Backbone.Tooling.Extensions;
 using Google.Api.Gax;
 using Google.Apis.Auth.OAuth2;
 using Google.Cloud.PubSub.V1;
+using Microsoft.Extensions.Logging;
 
 namespace Backbone.BuildingBlocks.Infrastructure.EventBus.GoogleCloudPubSub;
 
 public class DefaultGoogleCloudPubSubPersisterConnection : IGoogleCloudPubSubPersisterConnection
 {
+    private readonly ILogger<DefaultGoogleCloudPubSubPersisterConnection> _logger;
     private bool _disposed;
 
-    public DefaultGoogleCloudPubSubPersisterConnection(string projectId, string topicId,
+    public DefaultGoogleCloudPubSubPersisterConnection(ILogger<DefaultGoogleCloudPubSubPersisterConnection> logger, string projectId, string topicId,
         string subscriptionName, string connectionInfo)
     {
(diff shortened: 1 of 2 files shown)


BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/GoogleCloudPubSub/DefaultGoogleCloudPubSubPersisterConnection.cs
```diff
@@ -39,8 +42,26 @@ public void Dispose()
         if (_disposed) return;
 
         _disposed = true;
-        PublisherClient.ShutdownAsync(CancellationToken.None).GetAwaiter().GetResult();
 
-        SubscriberClient.StopAsync(CancellationToken.None).GetAwaiter().GetResult();
+        try
+        {
+            PublisherClient.ShutdownAsync(CancellationToken.None).GetAwaiter().GetResult();
```
ericbrunner, line 48: If you need to await the returned Task, I would move that async code into an InitAsync() method and call and await it when the instance is created and the .ctor has run [https://blog.stephencleary.com/2013/01/async-oop-2-constructors.html](https://blog.stephencleary.com/2013/01/async-oop-2-constructors.html)]. Even if in backend .net core apps there is no Synchronization context and therefore the continuation is scheduled on a new background thread no dead-lock can occur, but it is hard to read and guess the intend. 

If the returned Task needs not to be awaited inside the .ctor it could be wrapped inside of the Task.Run( async () => try { await Publish. ...  ] catch { log.Error(e)} ); I would dispite that the returned Task needs not to be awaited, always log any possible exception (therefore the async-await in the lambda expression)

Dannyps (approved):

BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/GoogleCloudPubSub/DefaultGoogleCloudPubSubPersisterConnection.cs
tnotheis: > If you need to await the returned Task, I would move that async code into an InitAsync() method and call and await it when the instance is created and the .ctor has run https://blog.stephencleary.com/2013/01/async-oop-2-constructors.html]

Caution: the changed code is part of the `Dispose` function, not of the constructor. The diff GitHub shows is a bit misleading, as it hides some lines (see screenshot below).

image

> it could be wrapped inside of the Task.Run( async () => try { await Publish. ... ] catch { log.Error(e)} ); I would dispite that the returned Task needs not to be awaited, always log any possible exception (therefore the async-await in the lambda expression)

To be honest, I didn't even see the `GetAwaiter.GetResult` when I was catching the exceptions. 
I just changed the implementation so that the `IEventBus` implementations implement `IAsyncDisposable` instead of `IDisposable`. This should make the code much clearer. Here is the link to the corresponding PR: https://github.com/nmshd/backbone/pull/839

PR #839, opened 2024-09-03
Implement `DisposeAsync` in addition to `Dispose` in event buses

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/AzureServiceBus/EventBusAzureServiceBus.cs
+++ b/BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/AzureServiceBus/EventBusAzureServiceBus.cs
@@ -12,7 +12,7 @@ using Newtonsoft.Json;
 
 namespace Backbone.BuildingBlocks.Infrastructure.EventBus.AzureServiceBus;
 
-public class EventBusAzureServiceBus : IEventBus, IDisposable
+public class EventBusAzureServiceBus : IEventBus, IAsyncDisposable
 {
     private const string DOMAIN_EVENT_SUFFIX = "DomainEvent";
     private const string TOPIC_NAME = "default";
@@ -42,10 +42,10 @@ public class EventBusAzureServiceBus : IEventBus, IDisposable
         _handlerRetryBehavior = handlerRetryBehavior;
     }
 
-    public void Dispose()
+    public async ValueTask DisposeAsync()
     {
         _subscriptionManager.Clear();
-        _processor.CloseAsync().GetAwaiter().GetResult();
+        await _processor.CloseAsync();
     }
 
     public async void Publish(DomainEvent @event)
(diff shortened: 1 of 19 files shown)


daniel-almeida-konkconsulting:
As per Microsoft's official documentation:
_"If you implement the [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) interface but not the [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable) interface, your app can potentially leak resources. If a class implements [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable), but not [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable), and a consumer only calls Dispose, your implementation would never call DisposeAsync. This would result in a resource leak."_

So I would recommend that our classes implement both `IDisposable` and `IAsyncDisposable`.

You can find the referenced information here: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync

tnotheis:
> As per Microsoft's official documentation: _"If you implement the [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) interface but not the [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable) interface, your app can potentially leak resources. If a class implements [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable), but not [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable), and a consumer only calls Dispose, your implementation would never call DisposeAsync. This would result in a resource leak."_
> 
> So I would recommend that our classes implement both `IDisposable` and `IAsyncDisposable`.
> 
> You can find the referenced information here: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync

To be honest, the only time Dispose is called is when the application is shut down. So I think we wouldn't even need to implement _any_ Disposable interface.

But still, you're probably right. If we do it, we should do it right. Good catch!

BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/AzureServiceBus/EventBusAzureServiceBus.cs
```diff
@@ -43,9 +43,14 @@ public EventBusAzureServiceBus(IServiceBusPersisterConnection serviceBusPersiste
     }
 
     public void Dispose()
+    {
+        Task.Run(async () => await DisposeAsync()).GetAwaiter().GetResult();
```
ericbrunner, line 47: Looks strange to me. I would follow that pattern: [https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-both-dispose-and-async-dispose-patterns](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-both-dispose-and-async-dispose-patterns)

tnotheis: Looks really complicated. Especially if we consider that there will probably never be a case where we actually need to dispose, because, as I said in a previous comment, the disposal will only ever happen when shutting down the application.

tnotheis: But if you're interested in it, you can give it a shot. 😉 I need to focus on other things now and I don't know when I will be able to revisit it.

PR #856, opened 2024-09-09
Personalized Relationship Templates

NikolaDmitrasinovic:
# Readiness checklist

-   [x] I labeled the PR.

# Description
Relationship Templates now have optional `ForIdentity` property which can be used to address the template to a specific identity. If the property is specified during the creation of the template only the creator and the specified identity can retrieve it.


tnotheis:
@coderabbitai review

tnotheis (changes requested):
You should return the `ForIdentity` property as part of the `RelationshipTemplateDTO` (remember to add it to the SDK as well)

tnotheis (changes requested):
Ohh, now I see what was causing your confusion. You actually added the property to the `CreateRelationshipTemplateRequest` of the SDK. But not to the `CreateRelationshipTemplateResponse`.

tnotheis (approved):

PR #859, opened 2024-09-10
Consumer Api: Devices that have triggered an action receive a push notification for this action

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


tnotheis (changes requested):
Write a test please.

Modules/Devices/test/Devices.Application.Tests/Tests/PushNotifications/DatawalletModified/HandlerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using Backbone.Modules.Devices.Infrastructure.PushNotifications.Connectors;
+using Backbone.UnitTestTools.BaseClasses;
+using FakeItEasy;
+using FluentAssertions;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace Backbone.Modules.Devices.Application.Tests.Tests.PushNotifications.DatawalletModified;
+
+public class HandlerTests : AbstractTestsBase
```
tnotheis, line 18: Sorry, I should have been more clear: I was referring to a test for the PushService's new Send method.
The test you wrote isn't ideal because of the 2-layer-mocks/-fakes you built. That makes it way more complicated than it has to be. If you would test the PushService directly, you would save one layer of fakes. 

Also, instead of 
```c#
A.CallTo(() => fakeConnector.Send(A<IEnumerable<PnsRegistration>>._, A<IPushNotification>._))
            .Invokes((IEnumerable<PnsRegistration> registrations, IPushNotification _) => notifiedDevices.AddRange(registrations));
...
notifiedDevices.Should().HaveCount(1);
notifiedDevices.First().Should().Be(deviceB);
```

you should do:
```c#
A.CallTo(() => mockConnector.Send(...)).MustHaveHappendOnceExactly()
```

You can keep this test though. Just simplify it and only assert that `IPushService.Send` is called.

tnotheis (changes requested):

tnotheis (approved):

tnotheis:
@coderabbitai review

PR #861, opened 2024-09-12
Deletion of messages after sender and all recipients have been anonymized

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.

This pull request implements the automatic deletion of messages once both the sender and all recipients have been anonymized. Additionally, it introduces new `MessageOrphanedDomainEvent` and repository `Delete` function to support this functionality, ensuring that orphaned messages are handled.


tnotheis:
@coderabbitai review

Modules/Messages/src/Messages.Domain/Entities/Message.cs
```diff
@@ -111,5 +111,12 @@ public static Expression<Func<Message, bool>> WasExchangedBetween(IdentityAddres
             (m.CreatedBy == identityAddress2 && m.Recipients.Any(r => r.Address == identityAddress1));
     }
 
+    public static Expression<Func<Message, bool>> IsMessageOrphaned(string didDomainName)
+    {
+        return i =>
+            i.CreatedBy == IdentityAddress.GetAnonymized(didDomainName) &
```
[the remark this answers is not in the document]
HunorTotBagi: Good catch, my bad. && should have been used. Changes made.

HunorTotBagi:
Note to @MH321Productions:

I had a discussion with @tnotheis and I have to do a rework on this PR. Namely I have to use `DomainEvents` because in the current implementation orphaned messages are only being deleted during identity deletion, but messages should also be deleted when they are anonymized due to relationship termination or decomposition.

MH321Productions (approved):

Modules/Messages/src/Messages.Domain/DomainEvents/Incoming/MessageOrphanedDomainEvent.cs
```diff
@@ -0,0 +1,8 @@
+﻿using Backbone.BuildingBlocks.Domain.Events;
+
+namespace Backbone.Modules.Messages.Domain.DomainEvents.Incoming;
+
+public class MessageOrphanedDomainEvent : DomainEvent
+{
+    public required string IdentityAddress { get; set; }
```
tnotheis, line 7: Here the property is called `IdentityAddress`, but in the outgoing event, it is called `CreatedBy`. So this will not work.

But this also tells me that neither you nor @MH321Productions have tested the code manually... 😠

Modules/Messages/src/Messages.Domain/Entities/Message.cs
```diff
@@ -86,11 +86,13 @@ private void AnonymizeRecipient(IdentityAddress participantAddress, IdentityAddr
     {
         var recipient = Recipients.FirstOrDefault(r => r.Address == participantAddress);
         recipient?.UpdateAddress(anonymizedIdentityAddress);
+        RaiseDomainEvent(new MessageOrphanedDomainEvent(this));
```
tnotheis, line 89: Wtf. Why do you ALWAYS raise the `MessageOrphanedDomainEvent`, no matter if there are un-anonymized participants left?

Modules/Messages/src/Messages.Domain/Entities/Message.cs
```diff
@@ -86,11 +86,13 @@ private void AnonymizeRecipient(IdentityAddress participantAddress, IdentityAddr
...
     }
 
     private void AnonymizeSender(IdentityAddress anonymizedIdentityAddress)
     {
         CreatedBy = anonymizedIdentityAddress;
+        RaiseDomainEvent(new MessageOrphanedDomainEvent(this));
```
tnotheis, line 95: Wtf. Why do you ALWAYS raise the `MessageOrphanedDomainEvent`, no matter if there are un-anonymized participants left?

Modules/Messages/test/Messages.Domain.Tests/Messages/ReplaceIdentityAddressTests.cs
tnotheis, line 1: Once you've fixed the creation of the MessageOrphanedDomainEvent, you should add a test for it in this file.

tnotheis (changes requested):

Modules/Messages/src/Messages.Domain/DomainEvents/Incoming/MessageOrphanedDomainEvent.cs
HunorTotBagi: I swear to God that I have tested manually and it deleted the message, otherwise I wouldn't consider the PR ready for review.

tnotheis: Maybe you renamed the property afterwards then? Because as it currently is, it should throw an exception when trying to deserialize the event on the receiving side (I haven't tested this, though, so I could be wrong)

Modules/Synchronization/src/Synchronization.Domain/Entities/Sync/ExternalEvent.cs
```diff
@@ -61,6 +61,7 @@ public void SyncFailed(SyncError error)
 public enum ExternalEventType
 {
     MessageReceived = 0,
+    MessageOrphaned = 1,
```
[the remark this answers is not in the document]
HunorTotBagi: Changes made.

Modules/Synchronization/src/Synchronization.Domain/DomainEvents/Incoming/MessageOrphaned/MessageOrphanedDomainEvent.cs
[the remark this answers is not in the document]
HunorTotBagi: Changes made.

Modules/Messages/src/Messages.Domain/Entities/Message.cs
HunorTotBagi: I agree it's better to perform the check before raising the `MessageOrphanedDomainEvent`. Changes made.

tnotheis: Well, as soon as the event handler deletes the orphaned messages without any additional check it has to rely on the event to only be raised if the message is actually orphaned. With the current implementation the messages would be deleted as soon as the first participant is anonymized.

HunorTotBagi: In this implementation my logic for checking weather the message is orphaned or not is done in the `Delete` method in `MessageRepository` with the help of an expression,

```csharp
    public async Task Delete(Expression<Func<Message, bool>> filter, CancellationToken cancellationToken)
    {
        await _messages.Where(filter).ExecuteDeleteAsync(cancellationToken);
    }
 ```

the expression:

```csharp
    public static Expression<Func<Message, bool>> IsMessageOrphaned(string didDomainName)
    {
        var anonymizedAddress = IdentityAddress.GetAnonymized(didDomainName);

        return i =>
            i.CreatedBy == anonymizedAddress &&
            i.Recipients.All(r => r.Address == anonymizedAddress);
    }
 ```
 
 I raised every time the event when someone is anonymized (which now I understand it's not a good solution). Also I manually tested this. First I started a deletion process for identity A, run the job and the message wasn't deleted. Later on I started a deletion process for identity B and then the message was indeed deleted. As I mentioned in some of the previous comments, I tested everything manually otherwise I wouldn't consider the PR ready for review.

File: Modules/Messages/src/Messages.Domain/DomainEvents/Incoming/MessageOrphanedDomainEvent.cs (hunk 2 above)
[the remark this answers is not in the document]
Comment (HunorTotBagi) [reply Modules/Messages/src/Messages.Domain/DomainEvents/Incoming/MessageOrphanedDomainEvent.cs:7]: I didn't know that is possible. Changes made.

Comment (HunorTotBagi) [reply Modules/Messages/src/Messages.Domain/DomainEvents/Incoming/MessageOrphanedDomainEvent.cs:7]: No property renaming was done, and no exception was thrown.

Comment (tnotheis) [reply Modules/Messages/src/Messages.Domain/DomainEvents/Incoming/MessageOrphanedDomainEvent.cs]: Wow, interesting. I just did a small test, and it actually seems as if NewtonsoftJson doesn't throw if a `required` property does not exist. 😲 I really didn't expect that

File: Modules/Synchronization/src/Synchronization.Application/DomainEvents/Incoming/MessageOrphaned/MessageOrphanedDomainEventHandler.cs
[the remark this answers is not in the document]
Comment (HunorTotBagi) [reply Modules/Synchronization/src/Synchronization.Application/DomainEvents/Incoming/MessageOrphaned/MessageOrphanedDomainEventHandler.cs:1]: It appears I was mistaken in this case regarding the event being sent and received within the same module. I initially thought the `Synchronization` module needed to be included. Changes made.

File: Modules/Messages/src/Messages.Application/DomainEvents/Incoming/MessageOrphaned/MessageOrphanedDomainEventHandler.cs [hunk 6]
```diff
@@ -0,0 +1,24 @@
...
+
+    public MessageOrphanedDomainEventHandler(IMessagesRepository messagesRepository, IOptions<ApplicationOptions> applicationOptions)
+    {
+        _messagesRepository = messagesRepository;
+        _applicationOptions = applicationOptions.Value;
+    }
+
+    public async Task Handle(MessageOrphanedDomainEvent @event)
+    {
+        await _messagesRepository.Delete(Message.IsMessageOrphaned(_applicationOptions.DidDomainName), CancellationToken.None);
```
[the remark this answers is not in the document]
HunorTotBagi: I agree that I could have given it more thought. The approach of checking all messages was the first solution that came to my mind, and given our previous talk about performance concerns, I thought it would be an ok solution since it would run at night. However, I fully agree that passing the `Id` is a more efficient solution. Changes made.

Modules/Messages/src/Messages.Domain/Entities/Message.cs
HunorTotBagi: Changes made.

Modules/Messages/test/Messages.Domain.Tests/Messages/ReplaceIdentityAddressTests.cs
HunorTotBagi: Tests added.

tnotheis (approved):

PR #866, opened 2024-09-16
Set some useful MSBuild properties for all projects

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

Documentation for the added flags can be found here: https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props.


NikolaVetnic:
Just to see if I understood this correctly - instead of listing all the warnings that should be treated as errors, now you set all warnings to be treated thusly, and in addition you enable implicitly triggered build optimization and code style enforcement, correct?

tnotheis:
Exactly 👍

NikolaVetnic (approved):

PR #867, opened 2024-09-16
Send additional data about the identity deletion process to the peer of the identity to be deleted

HunorTotBagi:
# Readiness checklist

-   [x] I labeled the PR.

This pull request addresses the need to send additional data about the identity deletion process to the peer of the identity to be deleted. Specifically, it extends the following external event payloads:
- `PeerToBeDeleted`: The payload is extended by adding a `DeletionDate` property, which maps to the `GracePeriodEndsAt` value of the identity deletion process
- `PeerDeleted`: The payload is extended by adding a `DeletionDate` property, which is set to `UtcNow` at the time the external event is created


HunorTotBagi:
@MH321Productions The second acceptance criteria, as requested by @tnotheis, has been added. The pull request is now ready for review.

Modules/Devices/src/Devices.Domain/DomainEvents/Outgoing/IdentityToBeDeletedDomainEvent.cs
```diff
@@ -3,10 +3,12 @@
...
 {
-    public IdentityToBeDeletedDomainEvent(string identityAddress) : base($"{identityAddress}/IdentityToBeDeleted", randomizeId: true)
+    public IdentityToBeDeletedDomainEvent(string identityAddress, DateTime? gracePeriodEndsAt) : base($"{identityAddress}/IdentityToBeDeleted", randomizeId: true)
     {
         IdentityAddress = identityAddress;
+        GracePeriodEndsAt = gracePeriodEndsAt;
     }
 
     public string IdentityAddress { get; }
+    public DateTime? GracePeriodEndsAt { get; }
```
MH321Productions, line 13: Is there a specific reason (or case) why this has to be null? As far as I understood it, when the event is raised, the Deletion Process is either started by the owner or approved by them (therefore always approved). When approved, the `GracePeriodEndsAt` property is always a valid date (set in the `ApproveInternally` method).

Modules/Relationships/src/Relationships.Application/DomainEvents/Incoming/IdentityToBeDeleted/IdentityToBeDeletedDomainEventHandler.cs
```diff
@@ -19,7 +19,7 @@ public async Task Handle(IdentityToBeDeletedDomainEvent @event)
     {
         var relationships = await GetRelationshipsOf(@event.IdentityAddress);
 
-        NotifyRelationshipsOfPeerToBeDeleted(@event.IdentityAddress, relationships);
+        NotifyRelationshipsOfPeerToBeDeleted(@event.IdentityAddress, relationships, @event.CreationDate);
```
MH321Productions, line 22: Why do You use the Creation date of the event when the grace period is needed? I think, You need to change the `Relationships.Domain/DomainEvents/Incoming/IdentityToBeDeletedDomainEvent` as well as the outgoing one in the Devices Module. Then You can pass this value.

MH321Productions (changes requested):

Modules/Devices/src/Devices.Domain/DomainEvents/Outgoing/IdentityToBeDeletedDomainEvent.cs
HunorTotBagi: In the `IdentityDeletionProcess` file, the `GracePeriodEndsAt` property is declared as nullable, so I maintained that approach to pass it to the domain event. However, your question is valid but unfortunately I don't know the answer.

tnotheis: > I don't know the answer.

The answer is easy. If the property CAN be nullable, make it nullable. If it can NOT be nullable, make it not nullable.

And as Mika correctly mentioned, in case of this event the property will never be null.

HunorTotBagi: After giving it some thought, I now understand that the only scenario in which the `GracePeriodEndsAt` property is nullable occurs when the deletion process is initiated by support and is still in the approval phase. Changes made.

Modules/Relationships/src/Relationships.Application/DomainEvents/Incoming/IdentityToBeDeleted/IdentityToBeDeletedDomainEventHandler.cs
HunorTotBagi: You're absolutely right, my oversight :/ Changes made.

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -115,7 +115,7 @@ public IdentityDeletionProcess StartDeletionProcessAsOwner(DeviceId asDevice)
         DeletionGracePeriodEndsAt = deletionProcess.GracePeriodEndsAt;
         TierId = Tier.QUEUED_FOR_DELETION.Id;
         Status = IdentityStatus.ToBeDeleted;
-        RaiseDomainEvent(new IdentityToBeDeletedDomainEvent(Address, deletionProcess.GracePeriodEndsAt));
+        RaiseDomainEvent(new IdentityToBeDeletedDomainEvent(Address, (DateTime)deletionProcess.GracePeriodEndsAt!));
```
MH321Productions, line 118: This looks a bit wierd (though I understand that it somehow has to be converted, probably bc it's a property instead of a variable). My solution would be

```csharp
RaiseDomainEvent(new IdentityToBeDeletedDomainEvent(Address, deletionProcess.GracePeriodEndsAt!.Value));
```

But I don't know if there's a convention for this kind of situation, so I would leave that to @tnotheis to decide.

tnotheis: I would have done it just like you suggested it 👍

MH321Productions (approved):

tnotheis (approved):

tnotheis (approved):

PR #888, opened 2024-09-25
Add blob storage to health check

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

Note: The Consumer Api Health Check still shows the "old" json format (because the branch is based of main)


tnotheis (changes requested):
2 major issues:
1. Look where the SQL database health check is added and do it similarly for the blob storage health check
2. You should add a health check for Google Cloud Storage as well. I'm not sure if there is an emulator for it. If not, let me know so I can create a real one on the GCP.

PR #902, opened 2024-10-07
Add Blob Storage to Health Check

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
```diff
@@ -24,13 +26,18 @@ public static void AddBlobStorage(this IServiceCollection services, BlobStorageO
         {
             case AZURE_CLOUD_PROVIDER:
                 services.AddAzureStorageAccount(azureStorageAccountOptions => { azureStorageAccountOptions.ConnectionString = options.ConnectionInfo!; });
+                services.AddHealthChecks().AddAzureBlobStorage(
+                    clientFactory: _ => new BlobServiceClient(options.ConnectionInfo),
+                    optionsFactory: _ => new AzureBlobStorageHealthCheckOptions { ContainerName = options.Container }
+                );
```
tnotheis, line 32: Can you please check whether the result of the clientFactory is cached somehow or whether a new BlobServiceClient is created on every health check?
Because if it's the latter, we need a different solution.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorageServiceCollectionExtensions.cs
```diff
@@ -30,6 +30,8 @@ public static void AddGoogleCloudStorage(this IServiceCollection services, Googl
             return storageClient;
         });
 
+        services.AddSingleton(_ => options);
```
tnotheis, line 33: What's that?

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
tnotheis, line 32: This method should be called from `services.AddAzureStorageAccount`

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
```diff
@@ -24,13 +26,18 @@ public static void AddBlobStorage(this IServiceCollection services, BlobStorageO
         {
             case AZURE_CLOUD_PROVIDER:
                 services.AddAzureStorageAccount(azureStorageAccountOptions => { azureStorageAccountOptions.ConnectionString = options.ConnectionInfo!; });
+                services.AddHealthChecks().AddAzureBlobStorage(
+                    clientFactory: _ => new BlobServiceClient(options.ConnectionInfo),
+                    optionsFactory: _ => new AzureBlobStorageHealthCheckOptions { ContainerName = options.Container }
```
tnotheis, line 31:
```suggestion
                    optionsFactory: sp => new AzureBlobStorageHealthCheckOptions { ContainerName = sp.GetRequiredService<IOptions<BlobOptions>>().Value.RootFolder }
```

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorageHealthCheck.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                //Upload random Blob with predefined text
+                var filename = Guid.NewGuid().ToString();
+                await _storage.UploadObjectAsync(_options.BucketName, filename, "text/plain", new MemoryStream(FILE_TEXT.GetBytes()), null, cancellationToken);
+
+                //Download the Blob again and check the text
+                var downloadStream = new MemoryStream();
+                await _storage.DownloadObjectAsync(_options.BucketName, filename, downloadStream, null, cancellationToken);
+                var downloadedString = Encoding.UTF8.GetString(downloadStream.ToArray());
+                if (downloadedString != FILE_TEXT) _isHealthy = false;
```
tnotheis, line 36: Setting `isHealthy` to false here doesn't do anything. It's overwritten in line 41 or 45 (depending on whether an exception will be thrown by the deletion or not)

I'd suggest to refactor the whole method anyway:

```c#
        if (!_isHealthy.HasValue)
        {
            var isUploadPossible = await IsUploadPossible();
            var isDownloadPossible = await IsDownloadPossible();
            var isDeletionPossible = await IsDeletionPossible();

            _isHealthy = isUploadPossible && isDownloadPossible && isDeletionPossible;
        }

        return _isHealthy.Value ? HealthCheckResult.Healthy() : HealthCheckResult.Unhealthy();
```

This has 2 advantages:
1. It is more readable, because the methods are smaller.
2. You cannot make a mistake like the one I mentioned above, because each method is responsible for its own health check result.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorageHealthCheck.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        _storage = storage;
+        _options = options;
+    }
+
+    private readonly StorageClient _storage;
+    private readonly GoogleCloudStorageOptions _options;
+
+    private static bool? _isHealthy;
+    private const string FILE_TEXT = "healthcheck";
```
tnotheis, line 20: Fields should always be placed above constructors.

Also, const fields should be placed above static fields, and static fields should be placed above instance fields.

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorageServiceCollectionExtensions.cs
MH321Productions: The Health Check needs the options to get the bucket name. In this constellation it was the only way to make the Dependency Injection work. But I can rewrite that in the `services.AddGoogleCloudStorage` method

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
MH321Productions: `BlobOptions` is a part of the `Files.Application` project. Should I create a dependency (Since this is a common project)?

tnotheis: Damn. No, no dependency. We have to find another way then. 😞

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
MH321Productions: The client factory is created when a `AzureBlobStorageHealthCheck` instance is created. It is created with a factory during the creation of a `HealthCheckRegistration`. I don't think that the registration is recreated at every health check, but I'm not sure if the Health check class is created multiple times.

tnotheis: Testing this should be as simple as setting a breakpoint into the client factory and running the health check a bunch of times.

MH321Productions: The client factory gets created every time, but it can be configured to use the service provider, so I added the `BlobServiceClient` as a Singleton.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/BlobStorageServiceCollectionExtensions.cs
MH321Productions: I use `sp.GetRequiredService<IOptions<BlobStorageOptions>>().Value.Container` now

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorageHealthCheck.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public GoogleCloudStorageHealthCheck(StorageClient storage, GoogleCloudStorageOptions options)
+    {
+        _storage = storage;
+        _options = options;
+    }
+
+    public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
+    {
+        if (!_isHealthy.HasValue)
```
tnotheis, line 24: I just realized: if the previous health check was unsuccessful, we should give it a chance to retry. I'd say up to 5 times.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/AzureStorageAccount/AzureStorageAccountServiceCollectionExtensions.cs
```diff
@@ -20,6 +23,11 @@ public static void AddAzureStorageAccount(this IServiceCollection services, Azur
         services.Configure<AzureStorageAccountOptions>(opt => opt.ConnectionString = options.ConnectionString);
         services.AddSingleton<AzureStorageAccountContainerClientFactory>();
         services.AddScoped<IBlobStorage, AzureStorageAccount>();
+        services.AddSingleton(_ => new BlobServiceClient(options.ConnectionString));
```
tnotheis, line 26: Use the extension method provided by the blob storage library instead

```suggestion
        services.AddBlobServiceClient(options.ConnectionString);
```

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/AzureStorageAccount/AzureStorageAccountServiceCollectionExtensions.cs
```diff
@@ -20,6 +23,11 @@ public static void AddAzureStorageAccount(this IServiceCollection services, Azur
...
+
+        services.AddHealthChecks().AddAzureBlobStorage(
+            optionsFactory: sp => new AzureBlobStorageHealthCheckOptions { ContainerName = sp.GetRequiredService<IOptions<BlobStorageOptions>>().Value.Container }
+        );
```
tnotheis, line 30: Hmm, you know what? I think I like the self-implemented health check for Google Cloud more than the one of the library for Azure, because it also checks the permissions. I think we should create a custom implementation for Azure as well. 
You should even be able to only use a single implementation. This implementation could simply use our `IBlobStorage` to perform its checks.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/GoogleCloudStorage/GoogleCloudStorageServiceCollectionExtensions.cs
```diff
@@ -47,6 +48,9 @@ public static void AddGoogleCloudStorage(this IServiceCollection services, Googl
 
             return new GoogleCloudStorage(storageClient, logger);
         });
+
+        services.AddHealthChecks()
+            .Add(new HealthCheckRegistration("Google Cloud", sp => new GoogleCloudStorageHealthCheck(sp.GetRequiredService<StorageClient>(), options), HealthStatus.Unhealthy, null));
```
tnotheis, line 53: The health check shouldn't be called "Google Cloud", because
1. It doesn't check the health of the whole Google Cloud
2. It should be transparent to the caller whether we use Google Cloud Storage, Azure Storage Account, or whatever.

So just call it "blob_storage".

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/BlobStorage/AzureStorageAccount/AzureStorageAccountServiceCollectionExtensions.cs
MH321Productions: This doesn't work. The reason: `The type 'Microsoft.Extensions.DependencyInjection.IServiceCollection' must be convertible to 'Azure.Core.Extensions.IAzureClientFactoryBuilder'`. The method is generic and needs an Azure specific builder. Looking inside the documentation, it tells me to use
```c#
services.AddAzureClients(cb => cb.AddBlobServiceClient(options.ConnectionString))
```
but this method doesn't exist either.

tnotheis: Shouldn't matter anyway if we go with the approach I described in https://github.com/nmshd/backbone/pull/902#discussion_r1794689509.

BuildingBlocks/src/BuildingBlocks.Infrastructure/BuildingBlocks.Infrastructure.csproj
```diff
@@ -19,6 +19,7 @@
         <PackageReference Include="RabbitMQ.Client" Version="6.8.1" />
         <PackageReference Include="Serilog" Version="4.0.2" />
         <PackageReference Include="System.Interactive.Async" Version="6.0.1" />
+        <PackageReference Include="AspNetCore.HealthChecks.Azure.Storage.Blobs" Version="8.0.1" />
```
tnotheis, line 22: No more need for this packages

tnotheis (changes requested):

tnotheis (approved):

PR #903, opened 2024-10-08
Return the reason why the relationship cannot be created from CanCreateRelationship

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/CanCreate/GET.feature
+++ b/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/CanCreate/GET.feature
@@ -22,7 +22,7 @@ Feature: GET /Relationships/CanCreate
         When i1 sends a GET request to the /Relationships/CanCreate?peer={id} endpoint with id=i2.id
         Then the response status code is 200 (OK)
         And a Relationship can be established
-        And the response doesn not contain a relationship creation check code
+        And the response does not contain a relationship creation check code
 
     Scenario: Two identities with a rejected and an active Relationship can't create one
         Given Identities i1 and i2
--- a/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/CanCreate/GET.feature
+++ b/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/CanCreate/GET.feature
@@ -6,6 +6,7 @@ Feature: GET /Relationships/CanCreate
         When i1 sends a GET request to the /Relationships/CanCreate?peer={id} endpoint with id=i2.id
         Then the response status code is 200 (OK)
         And a Relationship can be established
+        And there is no code
 
     Scenario: Two identities with an active Relationship can't create another one
         Given Identities i1 and i2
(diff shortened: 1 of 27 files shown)


Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Relationships/CanCreate/GET.feature
```diff
@@ -6,20 +6,23 @@ Feature: GET /Relationships/CanCreate
...
         And a Relationship can not be established
+        And the code is "error.platform.validation.relationship.relationshipToTargetAlreadyExists"
 
     Scenario: Two identities with a rejected Relationship can create one
         Given Identities i1 and i2
         And a rejected Relationship r between i1 and i2
         When i1 sends a GET request to the /Relationships/CanCreate?peer={id} endpoint with id=i2.id
         Then the response status code is 200 (OK)
         And a Relationship can be established
+        And there is no code
```
NikolaVetnic, line 25: I don't like that this is simply called `code`.

When we discuss things the "code" is usually the "status code". Strictly speaking this is `code` and that is `status code` so they are different, but I would still be more specific here so as to avoid confusion. For example, something like `there is no relationship creation check code` might do the trick.

Modules/Relationships/src/Relationships.Application/ApplicationErrors.cs
```diff
@@ -6,10 +6,10 @@ public static class ApplicationErrors
 {
     public static class Relationship
     {
-        public static ApplicationError PeerIsToBeDeleted(string peerToBeDeleted)
+        public static ApplicationError PeerIsToBeDeleted()
         {
             return new ApplicationError("error.platform.validation.relationship.peerIsToBeDeleted",
-                $"Cannot establish relationship with '{peerToBeDeleted}' because they are in status 'ToBeDeleted'.");
+                "Cannot establish relationship with the owner of the template because they are in status 'ToBeDeleted'.");
```
NikolaVetnic, line 12: I would say `...because the owner...` instead.

I am aware of modern tendencies in English language, however, despite being highly skilled in the use of it, it still sounds weird when people use "they" for a single person.

This is however an aesthetical concern so you may do however you think is best.

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
```diff
@@ -203,7 +203,7 @@ public async Task<IActionResult> CanEstablishRelationship([FromQuery(Name = "pee
         var peerIdentity = await _mediator.Send(new GetIdentityQuery(peerAddress), cancellationToken);
 
         var response = peerIdentity.Status is IdentityStatus.ToBeDeleted
-            ? new CanEstablishRelationshipResponse { CanCreate = false }
+            ? new CanEstablishRelationshipResponse { CanCreate = false, Code = ApplicationErrors.Relationship.PeerIsToBeDeleted().Code }
             : await _mediator.Send(new CanEstablishRelationshipQuery { PeerAddress = peerAddress }, cancellationToken);
```
NikolaVetnic, line 207: Why is this piece of logic in the controller? I would've moved all this to the handler.

NikolaVetnic (changes requested):

Modules/Relationships/src/Relationships.ConsumerApi/Controllers/RelationshipsController.cs
tnotheis: I would love to have it there. But the problem is that it uses code from a different module (the Handler is in the Relationships module, but it needs code from the Devices module (the GetIdentityQuery)).

NikolaVetnic: Ok, we discussed that in chat extensively, thanks for the insight :)

NikolaVetnic (approved):

PR #907, opened 2024-10-11
Make external event payloads strongly typed

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Modules/Synchronization/src/Synchronization.Application/DomainEvents/Incoming/PeerDeleted/PeerDeletedDomainEventHandler.cs
+++ b/Modules/Synchronization/src/Synchronization.Application/DomainEvents/Incoming/PeerDeleted/PeerDeletedDomainEventHandler.cs
@@ -1,4 +1,5 @@
 ﻿using Backbone.BuildingBlocks.Application.Abstractions.Infrastructure.EventBus;
+using Backbone.DevelopmentKit.Identity.ValueObjects;
 using Backbone.Modules.Synchronization.Application.Infrastructure;
 using Backbone.Modules.Synchronization.Domain.DomainEvents.Incoming.PeerDeleted;
 using Backbone.Modules.Synchronization.Domain.Entities.Sync;
@@ -25,16 +26,13 @@ public class PeerDeletedDomainEventHandler : IDomainEventHandler<PeerDeletedDoma
 
     private async Task CreateExternalEvent(PeerDeletedDomainEvent @event)
     {
-#pragma warning disable IDE0037
-        var payload = new
-        {
-            RelationshipId = @event.RelationshipId,
-            DeletionDate = SystemTime.UtcNow
-        };
-#pragma warning restore IDE0037
         try
         {
-            await _dbContext.CreateExternalEvent(@event.PeerOfDeletedIdentity, ExternalEventType.PeerDeleted, payload);
+            var payload = new PeerDeletedExternalEvent.PayloadT { RelationshipId = @event.RelationshipId, DeletionDate = SystemTime.UtcNow };
+
+            var externalEvent = new PeerDeletedExternalEvent(IdentityAddress.Parse(@event.PeerOfDeletedIdentity), payload);
+
+            await _dbContext.CreateExternalEvent(externalEvent);
         }
         catch (Exception ex)
         {
--- a/Modules/Synchronization/src/Synchronization.Application/Infrastructure/ISynchronizationDbContext.cs
+++ b/Modules/Synchronization/src/Synchronization.Application/Infrastructure/ISynchronizationDbContext.cs
@@ -13,7 +13,7 @@ public interface ISynchronizationDbContext : IDbContext
 
     Task<Datawallet?> GetDatawalletForInsertion(IdentityAddress owner, CancellationToken cancellationToken);
     Task<Datawallet?> GetDatawallet(IdentityAddress owner, CancellationToken cancellationToken);
-    Task<ExternalEvent> CreateExternalEvent(IdentityAddress owner, ExternalEventType type, object payload);
+    Task CreateExternalEvent(ExternalEvent externalEvent);
     Task<SyncRun> GetSyncRun(SyncRunId syncRunId, IdentityAddress createdBy, CancellationToken cancellationToken);
     Task<bool> IsActiveSyncRunAvailable(IdentityAddress createdBy, CancellationToken cancellationToken);
 
--- a/Modules/Synchronization/src/Synchronization.Domain/Entities/Sync/ExternalEvent.cs
+++ b/Modules/Synchronization/src/Synchronization.Domain/Entities/Sync/ExternalEvent.cs
@@ -18,11 +18,10 @@ public class ExternalEvent : Entity
         Payload = null!;
     }
 
-    public ExternalEvent(ExternalEventType type, IdentityAddress owner, long index, object payload)
+    protected ExternalEvent(ExternalEventType type, IdentityAddress owner, object payload)
     {
         Id = ExternalEventId.New();
         Type = type;
-        Index = index;
         Owner = owner;
         CreatedAt = SystemTime.UtcNow;
         Payload = payload;
@@ -32,7 +31,7 @@ public class ExternalEvent : Entity
 
     public ExternalEventId Id { get; }
     public ExternalEventType Type { get; }
-    public long Index { get; }
+    public long Index { get; private set; }
 
     public IdentityAddress Owner { get; }
     public DateTime CreatedAt { get; }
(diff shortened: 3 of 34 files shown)


Modules/Synchronization/src/Synchronization.Domain/Entities/Sync/ExternalEvent.cs
```diff
@@ -44,6 +43,11 @@ public ExternalEvent(ExternalEventType type, IdentityAddress owner, long index,
     public SyncRunId? SyncRunId { get; private set; }
     public IReadOnlyCollection<SyncError> Errors => _errors;
 
+    public void UpdateIndex(long newIndex)
+    {
+        Index = newIndex;
+    }
+
```
tnotheis, line 50: Hint for the reviewer: this method is necessary now, because we cannot recreate the External Event in the SynchronizationDbContext anymore (because the generic SynchronizationDbContext.CreateExternalEvent method doesn't know which external event to create).

But even though letting the Index default to `0` in the cosntructor is kind of ugly, I still think this apporach is somewhat better than the old one, because now the SynchronizationDbContext.CreateExternalEvent is way simpler than before.

Modules/Synchronization/src/Synchronization.Infrastructure/Persistence/Database/Configurations/ExternalEventEntityTypeConfiguration.cs
```diff
@@ -1,3 +1,4 @@
+using Autofac.Builder;
```
HunorTotBagi, line 1: Using directive is not required by the code and can be safely removed

Modules/Synchronization/src/Synchronization.Infrastructure/Persistence/Database/ValueConverters/ExternalEventPayloadEntityFrameworkValueConverter.cs
```diff
@@ -3,9 +3,9 @@
...
 
 namespace Backbone.Modules.Synchronization.Infrastructure.Persistence.Database.ValueConverters;
-public class ExternalEventIdPayloadEntityFrameworkValueConverter : ValueConverter<object, string>
+public class ExternalEventPayloadEntityFrameworkValueConverter : ValueConverter<object, string>
 {
-    public ExternalEventIdPayloadEntityFrameworkValueConverter() : base(
+    public ExternalEventPayloadEntityFrameworkValueConverter() : base(
         o => JsonConvert.SerializeObject(o),
         s => JsonConvert.DeserializeObject<ExpandoObject>(s)!)
     { }
```
HunorTotBagi, line 11:
```suggestion
    {
    }
```

Modules/Synchronization/test/Synchronization.Domain.Tests/ExternalEventTests.cs
HunorTotBagi, line 1: For some reason I can't highlight line 4, but here is an unused using directive `using Backbone.UnitTestTools.FluentAssertions.Extensions;`

HunorTotBagi (changes requested):
I noticed that in `DomainEvents.Incoming` most of the files we have a `Handle` method and a `CreateExternalEvents` but in 

- `IdentityDeletionProcessStartedDomainEventHandler`
- `IdentityDeletionProcessStatusChangedDomainEventHandler`
- `RelationshipStatusChangedDomainEventHandler`

we don't follow this "pattern". So for consistency sake I would refactor the above three mentioned files to have a `Handle` and a `CreateExternalEvents` method. What do you think about this?

Modules/Synchronization/src/Synchronization.Application/DomainEvents/Incoming/PeerDeleted/PeerDeletedDomainEventHandler.cs
```diff
@@ -25,16 +26,13 @@ public async Task Handle(PeerDeletedDomainEvent domainEvent)
...
-        var payload = new
-        {
-            RelationshipId = @event.RelationshipId,
-            DeletionDate = SystemTime.UtcNow
-        };
-#pragma warning restore IDE0037
         try
         {
-            await _dbContext.CreateExternalEvent(@event.PeerOfDeletedIdentity, ExternalEventType.PeerDeleted, payload);
+            var payload = new PeerDeletedExternalEvent.PayloadT { RelationshipId = @event.RelationshipId, DeletionDate = SystemTime.UtcNow };
```
ericbrunner, line 31: I would name the `PayloadT` class either `Payload` or `EventPayload` or has the `T` suffix a special meaning?

Modules/Synchronization/src/Synchronization.Application/Infrastructure/ISynchronizationDbContext.cs
```diff
@@ -13,7 +13,7 @@ Task<DbPaginationResult<DatawalletModification>> GetDatawalletModifications(Iden
 
     Task<Datawallet?> GetDatawalletForInsertion(IdentityAddress owner, CancellationToken cancellationToken);
     Task<Datawallet?> GetDatawallet(IdentityAddress owner, CancellationToken cancellationToken);
-    Task<ExternalEvent> CreateExternalEvent(IdentityAddress owner, ExternalEventType type, object payload);
+    Task CreateExternalEvent(ExternalEvent externalEvent);
```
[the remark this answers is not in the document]
tnotheis: I removed some of the newlines. I only kept the two that were too long for a single line.

Modules/Synchronization/src/Synchronization.Application/DomainEvents/Incoming/PeerDeleted/PeerDeletedDomainEventHandler.cs
tnotheis: The reason for the T was that `Payload` is an inner class of a class that has a `Payload` property. And you cannot have a class and a property with the same name in the same class.

But I like `EventPayload`. Will rename it.

Modules/Synchronization/src/Synchronization.Infrastructure/Persistence/Database/SynchronizationDbContext.cs
```diff
@@ -94,24 +94,20 @@ public async Task<DbPaginationResult<DatawalletModification>> GetDatawalletModif
...
 
-    public async Task<ExternalEvent> CreateExternalEvent(IdentityAddress owner, ExternalEventType type, object payload)
+    public async Task CreateExternalEvent(ExternalEvent externalEvent)
     {
-        ExternalEvent? externalEvent = null;
-
         await RunInTransaction(async () =>
         {
-            if (externalEvent != null)
-                // if the transaction is retried, the old event has to be removed from the DbSet, because a new one with a new index is added
```
ericbrunner, line 104: Even with a single statement, surrounding that statement with curly braces enhances readability and even debugging

tnotheis: You commented on removed code. :) The new code doesn't contain this problem anymore.

HunorTotBagi:
> I noticed that in `DomainEvents.Incoming` most of the files we have a `Handle` method and a `CreateExternalEvents` but in
> 
> * `IdentityDeletionProcessStartedDomainEventHandler`
> * `IdentityDeletionProcessStatusChangedDomainEventHandler`
> * `RelationshipStatusChangedDomainEventHandler`
> 
> we don't follow this "pattern". So for consistency sake I would refactor the above three mentioned files to have a `Handle` and a `CreateExternalEvents` method. What do you think about this?

@tnotheis Thoughts on this? It seems you missed this one.

ericbrunner (dismissed):

tnotheis:
@HunorTotBagi Sorry, you're right. I didn't see that.

I just fixed it. I also renamed the methods so they contain the name of the event that is created. So if there is another external event created by the same domain event handler, we would do this in another private method.

Further, I moved the try/catch block to the `Handle` method, so that we don't have to catch exceptions in all of the private methods.

Btw: I think we could even remove the try/catch blocks, because the EventBus classes already log descriptive error messages. But I will not remove them in this PR.

HunorTotBagi:
> @HunorTotBagi Sorry, you're right. I didn't see that.
> 
> I just fixed it. I also renamed the methods so they contain the name of the event that is created. So if there is another external event created by the same domain event handler, we would do this in another private method.
> 
> Further, I moved the try/catch block to the `Handle` method, so that we don't have to catch exceptions in all of the private methods.
> 
> Btw: I think we could even remove the try/catch blocks, because the EventBus classes already log descriptive error messages. But I will not remove them in this PR.

Thanks, I like the changes.

I also noticed that the naming of the parameters in the `Handle` and in the private methods are all different for example we can see `domainEvent`, `@event` and `integrationEvent`. We should choose in what method which naming to use and apply all over them.

tnotheis:
> I also noticed that the naming of the parameters in the `Handle` and in the private methods are all different for example we can see `domainEvent`, `@event` and `integrationEvent`. We should choose in what method which naming to use and apply all over them.

Done

HunorTotBagi (approved):

PR #915, opened 2024-10-17
Error message for `error.platform.validation.message.relationshipToRecipientNotActive` contains wrong address

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Modules/Messages/src/Messages.Domain/DomainErrors.cs
+++ b/Modules/Messages/src/Messages.Domain/DomainErrors.cs
@@ -10,7 +10,7 @@ public static class DomainErrors
 
         return new DomainError(
             "error.platform.validation.message.relationshipToRecipientNotActive",
-            $"Cannot send message to {recipientText} because the relationship to it is not active. In order to be able to send messages again, you have to reactivate the relationship.");
+            $"Cannot send message to {recipientText} because the relationship to it is not active. If it's terminated, you'll need to reactivate it to be able to send messages again.");
     }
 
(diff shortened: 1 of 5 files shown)


Modules/Messages/src/Messages.Domain/DomainErrors.cs
```diff
@@ -10,7 +10,7 @@ public static DomainError RelationshipToRecipientNotActive(string recipient = ""
 
         return new DomainError(
             "error.platform.validation.message.relationshipToRecipientNotActive",
-            $"Cannot send message to {recipientText} because the relationship to it is not active. In order to be able to send messages again, you have to reactivate the relationship.");
+            $"Cannot send message to {recipientText} because the relationship to it is not active. If it's terminated, you'll need to reactivate it to be able to send messages again.");
```
jkoenig134, line 13: you should know if it is terminated.. why not specify what exactly is going wrong here?

jkoenig134 (approved):
If you don't want to have the error message detailed this LGTM.

Modules/Messages/src/Messages.Domain/DomainErrors.cs
tnotheis: I don't know if it is necessary to be THAT specific. It should be easy to figure that out on the client. And the code to achieve that on the Backbone would be ugly (Currently we only check if the status is Active, and if not, we throw the error. If we wanted to change the error message, we would have to check if the status is Terminated, and if yes, set the message1, otherwise message2.

Also: soon it will be possible to send messages to terminated relationships. So this part will be removed anyway. :)

PR #917, opened 2024-10-18
Consumer Api: Add get Tags endpoint

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

# Internal save format

I save the tags internally in a different format:

```json
"Tags": {
    "Application": {
        "Attributes": [
            {
                "name": "AttributeName",
                "tags": [
                    {
                        "tag": "tagName",
                        "displayNames": {
                            "de": "Name",
                            "en": "Name"
                        },
                        "children": [
                          {
                                "tag": "a",
                                "displayNames": {
                                    "de": "Name",
                                    "en": "Name"
                                }
                            },
                            {
                                "tag": "b",
                                "displayNames": {
                                    "de": "Name",
                                    "en": "Name"
                                }
                            }
                        ]
                    }
                ]
            }
        ]
    }
}
```

I did this because it is easier to read and decode (without needing to write a custom converter)


Modules/Tags.Application/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+    public static void AddApplication(this IServiceCollection services)
+    {
+        services.AddMediatR(c => c
+            .RegisterServicesFromAssemblyContaining<ListTagsQuery>()
+        );
+    }
+
+    public static void AddTags(this IServiceCollection services, ApplicationOptions applicationOptions)
+    {
+        services.AddSingleton(_ => new TagProvider(applicationOptions.Attributes));
```
tnotheis, line 17:
```suggestion
        services.AddSingleton(new TagProvider(applicationOptions.Attributes));
```

tnotheis, line 17: You should rather use a `(I)TagsRepository` to match the existing pattern. The only difference to the other repositories would be that it reads from configuration instead of from the database.
This way, we don't have to change any existing code when we want to read the tags from the database one day

Modules/Tags.Application/ApplicationOptions.cs
tnotheis, line 1: This should probably be moved to the `Domain` project

Modules/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
```diff
@@ -0,0 +1,65 @@
...
+            .SelectMany(tags => tags)
+            .SelectMany(tag => tag.DisplayNames.Keys)
+            .Distinct();
+
+    public ListTagsResponse(IReadOnlyDictionary<string, IEnumerable<TagInfo>> tags) : base(tags)
+    {
+    }
+}
+
+public class ListTagsResponseConverter : JsonConverter<ListTagsResponse>
```
tnotheis, line 20: If I don't miss anything, then a JsonConverter shouldn't be necessary. The JsonSerializer should be able to handle everything on its own - given the model is built correctly.

Modules/Tags.ConsumerApi/Configuration.cs
tnotheis, line 1: What's this?

Modules/Tags.ConsumerApi/Controllers/TagsController.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Backbone.BuildingBlocks.API.Mvc;
+using Backbone.Modules.Tags.Application.Tags.Queries.ListTags;
+using MediatR;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Backbone.Modules.Tags.ConsumerApi.Controllers;
+
+[Route("api/v1/[controller]")]
+public class TagsController : ApiControllerBase
```
tnotheis, line 9: You should probably add the `[Authorize("OpenIddict.Validation.AspNetCore")]` attribute to the controller, and the `AllowAnonymous` attribute to the controller method. This makes sure that when adding a new route we don't accidentally make it public.

Modules/Tags.ConsumerApi/TagsModule.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        var parsedConfiguration = services.BuildServiceProvider().GetRequiredService<IOptions<ApplicationOptions>>().Value;
+
+        services.AddApplication();
+        services.AddTags(parsedConfiguration);
+    }
+
+    public override void ConfigureEventBus(IEventBus eventBus)
+    {
+        //No Event bus needed here
```
tnotheis, line 28:
```suggestion
        // No Event bus needed here
```

appsettings.override.json
tnotheis, line 1: This file doesn't contain the latest structure.

tnotheis (changes requested):

Modules/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
MH321Productions: Since the tag and attribute names are properties, they would be converted to camel case when serializing. Therefore I had to write a custom converter.

Modules/Tags.ConsumerApi/Configuration.cs
MH321Productions: I wasn't sure whether we would need it, since it seemed to be one giant json object getting serialized.

Modules/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
tnotheis: I'm would imagine there is a way to avoid that.

.ci/appsettings.override.postgres.docker.json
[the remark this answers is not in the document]
MH321Productions: - I generate the `supportedLanguages` field by parsing the given entries
- As I wrote in the first comment, this format is easier to decode, but I can rewrite it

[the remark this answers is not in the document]
tnotheis: > I generate the `supportedLanguages` field by parsing the given entries

I think that's not how it's supposed to work. :) The `supportedLanguages`, as far as I understood, should be some kind of safety net. You want to define which languages should be supported, and the backbone should give you an error if one of the attributes doesn't have a translation for one of the languages.
@jkoenig134 Correct me if I'm wrong.

> As I wrote in the first comment, this format is easier to decode, but I can rewrite it

I think it might make sense to use the same format in the configuration as we return from the API. This avoids any mental mapping for the person responsible for the configuration. 
And since the implementation only has to be done once, but the configuration potentially multiple times, I would try to minimize the effort for configuration over the effort for implementation.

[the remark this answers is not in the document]
jkoenig134: You are right :)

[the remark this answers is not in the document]
MH321Productions: Ok, working on it

tnotheis (changes requested):

Modules/Tags/src/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        WriteSupportedLanguages(writer, value.SupportedLanguages);
+        WriteAttributes(writer, value.TagsForAttributeValueTypes);
+
+        writer.WriteEndObject();
+    }
+
+    private void WriteSupportedLanguages(Utf8JsonWriter writer, IEnumerable<string> supportedLanguages)
+    {
+        writer.WriteStartArray("supportedLanguages");
```
tnotheis, line 30: You shouldn't hardcode the name:

```suggestion
        writer.WriteStartArray(nameof(ListTagsResponse).ToCamelCase());
```
(`ToCamelCase` has to be implemented by you)

Modules/Tags/src/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
```diff
@@ -0,0 +1,65 @@
...
+        foreach (var lang in supportedLanguages) writer.WriteStringValue(lang);
+        writer.WriteEndArray();
+    }
+
+    private void WriteAttributes(Utf8JsonWriter writer, Dictionary<string, Dictionary<string, TagInfo>> attributes)
+    {
+        writer.WriteStartObject("tagsForAttributeValueTypes");
```
tnotheis, line 37: See above

Modules/Tags/src/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        writer.WriteEndObject();
+    }
+
+    private void WriteTag(Utf8JsonWriter writer, string tagName, TagInfo tag)
+    {
+        writer.WriteStartObject(tagName);
+
+        writer.WriteStartObject("displayNames");
```
tnotheis, line 52: See above

Modules/Tags/src/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
```diff
@@ -0,0 +1,65 @@
...
+        foreach (var (lang, displayName) in tag.DisplayNames) writer.WriteString(lang, displayName);
+        writer.WriteEndObject();
+
+        if (tag.Children.Count > 0)
+        {
+            writer.WriteStartObject("children");
```
tnotheis, line 58: See above

tnotheis (changes requested):

Modules/Tags/src/Tags.Application/Tags/Queries/ListTags/ListTagsResponse.cs
tnotheis: I just updated the suggestion to actually contain the property name instead of just `nameof(ListTagsResponse)`.

PR #950, opened 2024-11-26
Add a hint to the error message of the `error.platform.recordNotFound` regarding a potentially incorrect password

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


Magnus-Kuhn (approved):
I originally didn't want to have the password warning show up for things that have nothing to do with passwords, but I guess it's consistent with expired record warnings showing up for things that can't expire. So I think it's fine.

tnotheis:
Yeah, I didn't like it at first either. But the alternative would require a lot of effort, which isn't worth it just to get a slightly better error message.

jkoenig134 (approved):

PR #962, opened 2024-12-04
`ForIdentity` column of Relationship Templates and Tokens are not anonymized during identity deletion

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Modules/Relationships/src/Relationships.Application/RelationshipTemplates/Commands/AnonymizeRelationshipTemplatesForIdentity/Handler.cs
+++ b/Modules/Relationships/src/Relationships.Application/RelationshipTemplates/Commands/AnonymizeRelationshipTemplatesForIdentity/Handler.cs
@@ -1,0 +1,28 @@
+﻿using Backbone.DevelopmentKit.Identity.ValueObjects;
+using Backbone.Modules.Relationships.Application.Infrastructure.Persistence.Repository;
+using Backbone.Modules.Relationships.Domain.Aggregates.RelationshipTemplates;
+using MediatR;
+using Microsoft.Extensions.Options;
+
+namespace Backbone.Modules.Relationships.Application.RelationshipTemplates.Commands.AnonymizeRelationshipTemplatesForIdentity;
+
+public class Handler : IRequestHandler<AnonymizeRelationshipTemplatesForIdentityCommand>
+{
+    private readonly IRelationshipTemplatesRepository _relationshipTemplatesRepository;
+    private readonly ApplicationOptions _applicationOptions;
+
+    public Handler(IRelationshipTemplatesRepository relationshipTemplatesRepository, IOptions<ApplicationOptions> options)
+    {
+        _relationshipTemplatesRepository = relationshipTemplatesRepository;
+        _applicationOptions = options.Value;
+    }
+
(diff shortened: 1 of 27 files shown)


Modules/Relationships/src/Relationships.Application/RelationshipTemplates/Commands/AnonymizeRelationshipTemplatesForIdentity/Handler.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        _relationshipTemplatesRepository = relationshipTemplatesRepository;
+        _applicationOptions = options.Value;
+    }
+
+    public async Task Handle(AnonymizeRelationshipTemplatesForIdentityCommand request, CancellationToken cancellationToken)
+    {
+        var relationshipTemplates = (await _relationshipTemplatesRepository.FindTemplates(RelationshipTemplate.IsFor(IdentityAddress.Parse(request.IdentityAddress)), cancellationToken)).ToList();
+
+        foreach (var relationshipTemplate in relationshipTemplates) relationshipTemplate.AnonymizeForIdentity(_applicationOptions.DidDomainName);
```
tnotheis, line 24: It's really hard to read if the main purpose of the loop is not at the beginning of the line.
```suggestion
        foreach (var relationshipTemplate in relationshipTemplates) 
            relationshipTemplate.AnonymizeForIdentity(_applicationOptions.DidDomainName);
```

Modules/Relationships/src/Relationships.Domain/Aggregates/RelationshipTemplates/RelationshipTemplate.cs
```diff
@@ -69,6 +69,15 @@ public void AllocateFor(IdentityAddress identity, DeviceId device)
         Allocations.Add(new RelationshipTemplateAllocation(Id, identity, device));
     }
 
+    public void AnonymizeForIdentity(string didDomainName)
+    {
+        EnsureIsPersonalized();
```
tnotheis, line 74: I don't think that an un-personalized template has to throw here. We should just ignore it.

Modules/Relationships/src/Relationships.Domain/DomainErrors.cs
```diff
@@ -82,4 +82,10 @@ public static DomainError RelationshipTemplateNotAllocated()
         return new DomainError("error.platform.validation.relationship.relationshipTemplateNotAllocated",
             "The relationship template has to be allocated before it can be used to establish a relationship. Send a GET request to the /RelationshipTemplates/{id} endpoint to allocate the template.");
     }
+
+    public static DomainError RelationshipTemplateNotPersonalized()
```
tnotheis, line 86: Remember to remove this after removing the code that throws the error.

tnotheis (changes requested):

Modules/Relationships/src/Relationships.Domain/Aggregates/RelationshipTemplates/RelationshipTemplate.cs
MH321Productions: So it should just return silently?

tnotheis (approved):

PR #997, opened 2024-12-18
Consumer Api: Lock password protected Tokens after too many failed access attempts

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


Modules/Tokens/src/Tokens.Application/Tokens/Queries/GetToken/Handler.cs
```diff
@@ -29,9 +31,36 @@ private async Task<Token> GetToken(string tokenId, byte[]? password, Cancellatio
...
+        if (token.IsLocked) throw new ApplicationException(ApplicationErrors.TokenIsLocked());
+
+        if (!token.CanBeCollectedUsingPassword(address, password))
+        {
+            token.IncrementAccessFailedCount();
+            await _tokensRepository.Update(token, cancellationToken);
+
+            if (token.IsLocked) throw new ApplicationException(ApplicationErrors.TokenIsLocked());
+            throw new NotFoundException(nameof(Token));
+        }
```
tnotheis, line 64: I'd suggest to extract more logic into the Token class. For example by doing something like this:

```c#
var error = token.TryToAccess(address, password);

if (error == null)
    return;

if(error == DomainErrors.WrongPassword()){
    await _tokensRepository.Update(token, cancellationToken);
    throw new NotFoundException(nameof(Token));
}

throw new DomainException(error);
```

This was just a quick try. I may have forgotten something or there may be a better way. But the big advantage of this is that the main part of the logic is now in Domain, which makes it a lot easier to test.

Modules/Tokens/src/Tokens.Application/Tokens/Queries/GetToken/Handler.cs
```diff
@@ -29,9 +31,36 @@ private async Task<Token> GetToken(string tokenId, byte[]? password, Cancellatio
...
+        else if (identity != token.CreatedBy)
+        {
+            if (!token.HasAllocationForIdentity(identity))
+            {
+                await EnsureTokenCanBeCollected(token, identity, password, cancellationToken);
+
+                token.AddAllocationFor(identity, _userContext.GetDeviceId());
+                await _tokensRepository.Update(token, cancellationToken);
+            }
```
tnotheis, line 47: I think this stuff (except for line 46, of course) should also be placed into the `TryToAccess` method I suggested in the previous comment.

Modules/Tokens/src/Tokens.Application/Tokens/Queries/ListTokens/ListTokensQuery.cs
tnotheis, line 1: Do we still need this? Can't we just pass a list of strings instead?

Modules/Tokens/src/Tokens.Infrastructure/Persistence/Repository/TokensRepository.cs
```diff
@@ -29,19 +29,17 @@ public async Task<DbPaginationResult<Token>> FindTokens(IEnumerable<ListTokensQu
...
-        }
+            idFilter = idFilter.Or(Token.HasId(TokenId.Parse(inputQuery.Id)));
 
         var query = (track ? _tokensDbSet : _readonlyTokensDbSet)
+            .IncludeAll(_dbContext)
             .Where(Token.IsNotExpired)
             .Where(Token.CanBeCollectedBy(activeIdentity))
-            .Where(idAndPasswordFilter);
+            .Where(Token.HasAllocationFor(activeIdentity))
+            .Where(idFilter);
```
tnotheis, line 42: You should be able to simplify this query, now that we don't need to check the password anymore.

```c#
var query = (track ? _tokensDbSet : _readonlyTokensDbSet)
            .IncludeAll(_dbContext)
            .Where(Token.IsNotExpired)
            .Where(t => queryItems.Contains(t.Id)); // this line is new and replaces the foreach loop above)
```

Modules/Tokens/test/Tokens.Domain.Tests/Tests/Token.AllocationTests.cs
tnotheis, line 1: All the tests need to be updated, of course, after extracting everything into a single domain method, as suggested above.

tnotheis (changes requested):
In the original task description, I forgot to mention that during identity deletion, the `TokenAllocation`s have to get anonymized. I updated the task description accordingly.

Modules/Devices/src/Devices.Infrastructure/Translations/Resources.pt.resx
```diff
@@ -177,4 +177,10 @@
     <data name="BackupDeviceUsedPushNotification.Body" xml:space="preserve">
         <value>O seu dispositivo de cópia de segurança foi utilizado.</value>
     </data>
+    <data name="TokenLockedPushNotification.Body" xml:space="preserve">
+        <value>O seu token foi bloqueado devido a demasiadas tentativas erradas de introdução de palavra-passe</value>
+    </data>
+    <data name="TokenLockedPushNotification.Title" xml:space="preserve">
+        <value>Token bloqueado</value>
+    </data>
```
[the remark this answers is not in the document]
MH321Productions: I've used the designer. I don't know why it is reversed 🤷

Modules/Tokens/src/Tokens.Application/Tokens/Queries/ListTokens/ListTokensQuery.cs
MH321Productions: We don't need the query item, but we need the query. With the removal of the query item we can also remove them from the controller.

MH321Productions: Does something use the new `tokens` query entry in `List Tokens`? Since they aren't needed anymore, I could remove it, but if some component depends on it, it would need to be kept for "backwards compatibility".

tnotheis (approved):

PR #999, opened 2024-12-19
Admin API for listing all tokens of an identity

ericbrunner:
# Readiness checklist

-   [x] I labeled the PR.


Applications/AdminApi/src/AdminApi/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -151,6 +128,31 @@ public static IServiceCollection AddCustomAspNetCore(this IServiceCollection ser
...
+        var formattedMessage = string.IsNullOrEmpty(nameOfPropertyWithError)
+            ? firstErrorMessage
+            : $"'{nameOfPropertyWithError}': {firstErrorMessage}";
+
+        context.HttpContext.Response.ContentType = "application/json";
+
+        var responsePayload = new HttpResponseEnvelopeError(
+            HttpError.ForProduction(GenericApplicationErrors.Validation.InputCannotBeParsed().Code,
+                formattedMessage,
+                ""));
```
tnotheis, line 151: Could you please add the TODO comment again?

Applications/AdminApi/src/AdminApi/appsettings.json
tnotheis, line 1: Please keep the original formatting (4 space indentation).

Applications/AdminApi/src/AdminApi/Extensions/TokensServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        var infrastructureConfiguration = services.BuildServiceProvider().GetRequiredService<IOptions<TokensConfiguration.InfrastructureConfiguration>>().Value;
+
+        services.AddDatabase(options =>
+        {
+            options.Provider = infrastructureConfiguration.SqlDatabase.Provider;
+            options.DbConnectionString = infrastructureConfiguration.SqlDatabase.ConnectionString;
+        });
+
+        // Note: Registration required. Only Modules have their used repositories registered in the DI container.
+        services.AddTransient<ITokensRepository, TokensRepository>();
```
tnotheis, line 27: Replace all of this with a call to `AddPersistence`:

```suggestion
        services.AddPersistence();
```

Modules/Tokens/src/Tokens.Application/Tokens/Queries/ListTokensByIdentity/Validator.cs
tnotheis, line 1: Why do you use `CascadeMode.Stop` on every rule?

tnotheis (changes requested):
I know I told you it's not necessary to write an integration test for this  functionality (because we can't create tokens via the Admin API), but I just thought about it, and I think we could at least add one like "Listing all tokens of an identity that doesn't have any tokens".
This ensures that all DI wiring has been done correctly and that all dependencies can be resolved.

Modules/Tokens/src/Tokens.Application/Tokens/Queries/ListTokensByIdentity/Validator.cs
ericbrunner: I grabbed the configuration from there: `Modules/Tokens/src/Tokens.Application/Tokens/Queries/ListTokens/Validator.cs`
probably it is not required. So I'll remove the call to Cascade 'cause in that Validator case its not required

Applications/AdminApi/src/AdminApi/appsettings.json
ericbrunner: Done. BTW: Thats a local Rider setting, which was Indent: 2 on my machine. 
image

Modules/Tokens/src/Tokens.Application/Tokens/Queries/ListTokensByIdentity/Validator.cs
tnotheis: > in that Validator case its not required

Yeah, exactly. We usually only set it to `Cascade` when a following rule would throw an exception if the current one fails. For example, if there is a collection property, for which there are two rules:
* Not null
* Count > X

Then `Cascade` on the first rule would be necessary, because otherwise the second rule would throw a NullReferenceException because it tries to read `.Count` of `null`.

Applications/AdminApi/src/AdminApi/Extensions/IServiceCollectionExtensions.cs
ericbrunner: Done

Applications/AdminApi/src/AdminApi/Extensions/TokensServiceCollectionExtensions.cs
ericbrunner: Done

ericbrunner:
> I know I told you it's not necessary to write an integration test for this functionality (because we can't create tokens via the Admin API), but I just thought about it, and I think we could at least add one like "Listing all tokens of an identity that doesn't have any tokens". This ensures that all DI wiring has been done correctly and that all dependencies can be resolved.

Check last commit

Applications/AdminApi/src/AdminApi/Extensions/TokensServiceCollectionExtensions.cs
tnotheis: Sorry, I guess I wasn't clear. The `AddPersistence` method already exists in:

```
Modules/Tokens/src/Tokens.Infrastructure/Persistence/IServiceCollectionExtensions.cs
```

Applications/AdminApi/test/AdminApi.Tests.Integration/StepDefinitions/TokensStepDefinitions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Backbone.AdminApi.Tests.Integration.StepDefinitions;
+
+[Binding]
+[Scope(Feature = "GET /Tokens?createdBy={identity-address}")]
+internal class TokensStepDefinitions(HttpClientFactory factory, IOptions<HttpClientOptions> options) : BaseStepDefinitions(factory, options)
+{
+    private ApiResponse<CreateIdentityResponse>? _createIdentityResponse;
+    private string _newIdentity = string.Empty;
+    private ApiResponse<ListTokensTestResponse> _tokensResult = null!;
```
tnotheis, line 19: You should be consistent with the naming. At least within one file.

You call it `_createIdentityResponse`, but `_tokensResult`, even though both are of type `ApiResponse<T>`

Applications/AdminApi/test/AdminApi.Tests.Integration/StepDefinitions/TokensStepDefinitions.cs
```diff
@@ -0,0 +1,51 @@
...
+using PaginationFilter = Backbone.BuildingBlocks.SDK.Endpoints.Common.Types.PaginationFilter;
+
+
+namespace Backbone.AdminApi.Tests.Integration.StepDefinitions;
+
+[Binding]
+[Scope(Feature = "GET /Tokens?createdBy={identity-address}")]
+internal class TokensStepDefinitions(HttpClientFactory factory, IOptions<HttpClientOptions> options) : BaseStepDefinitions(factory, options)
+{
+    private ApiResponse<CreateIdentityResponse>? _createIdentityResponse;
```
tnotheis, line 17: Why do we need a field for that? As far as I can see, it's only used in one method.

Applications/AdminApi/test/AdminApi.Tests.Integration/StepDefinitions/TokensStepDefinitions.cs
```diff
@@ -0,0 +1,51 @@
...
+    private string _newIdentity = string.Empty;
```
tnotheis, line 18:
```suggestion
    private string _newIdentityAddress = string.Empty;
```

tnotheis (changes requested):

Applications/AdminApi/src/AdminApi/Extensions/TokensServiceCollectionExtensions.cs
ericbrunner: done

tnotheis (approved):

PR #1016, opened 2025-01-10
SQL injection via push notification configuration is possible

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Modules/Devices/src/Devices.Infrastructure/Persistence/Database/DevicesDbContext.cs
+++ b/Modules/Devices/src/Devices.Infrastructure/Persistence/Database/DevicesDbContext.cs
@@ -176,20 +176,9 @@ public class DevicesDbContext : IdentityDbContext<ApplicationUser>, IDevicesDbCo
 
     private List<string> GetAppIdsForWhichNoConfigurationExists(string platform, ICollection<string> supportedAppIds)
     {
-        var query = PnsRegistrations.FromSqlRaw(
-            Database.IsNpgsql()
-                ? $"""
-                     SELECT "AppId"
-                     FROM "Devices"."PnsRegistrations"
-                     WHERE "Handle" LIKE '{platform}%'
-                   """
-                : $"""
-                     SELECT "AppId"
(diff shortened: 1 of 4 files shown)


Modules/Devices/src/Devices.Infrastructure/Persistence/Database/DevicesDbContext.cs
```diff
@@ -176,20 +176,9 @@ public override int SaveChanges(bool acceptAllChangesOnSuccess)
...
-                : $"""
-                     SELECT "AppId"
-                     FROM [Devices].[PnsRegistrations]
-                     WHERE Handle LIKE '{platform}%'
-                   """);
-
-        return query
+        return PnsRegistrations
+            .AsNoTracking()
+            .Where(x => ((string)(object)x.Handle).StartsWith(platform))
```
ericbrunner, line 181: Handle is of Type PnsHandle and has a property Value. Followning will work:
.Where(x => x.Handle.Value.StartsWith(platform))

ericbrunner (dismissed):

Modules/Devices/src/Devices.Infrastructure/Persistence/Database/DevicesDbContext.cs
tnotheis: Unfortunately not. `PnsHandle` has a value converter which converts both properties of the class (`Value` and `Platform`) into a single string before saving it into the database, so that there is only one column instead of two.

I just tried it anyway, but as expected, you get the following error when trying to compile the query:

```
.Where(p => p.Handle.Value.StartsWith(__platform_0))' could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'.
```

ericbrunner (approved):

ericbrunner (approved):

ericbrunner (approved):

ericbrunner (approved):

PR #1027, opened 2025-01-20
Abl 490 limit the maximum number of message recipients

ericbrunner:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Modules/Messages/src/Messages.ConsumerApi/Controllers/MessagesController.cs
+++ b/Modules/Messages/src/Messages.ConsumerApi/Controllers/MessagesController.cs
@@ -59,6 +59,10 @@ public class MessagesController : ApiControllerBase
     [ProducesError(StatusCodes.Status400BadRequest)]
     public async Task<IActionResult> SendMessage(SendMessageCommand request, CancellationToken cancellationToken)
     {
+        if (request.Recipients.Count > _options.MaxNumberOfMessageRecipients)
+            throw new ApplicationException(
+                GenericApplicationErrors.Validation.InvalidNumberOfRecipients(_options.MaxNumberOfMessageRecipients));
+
         var response = await _mediator.Send(request, cancellationToken);
         return CreatedAtAction(nameof(GetMessage), new { id = response.Id }, response);
     }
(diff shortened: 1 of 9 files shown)


Modules/Messages/src/Messages.ConsumerApi/Controllers/MessagesController.cs
```diff
@@ -59,6 +59,10 @@ public async Task<IActionResult> GetMessage(string id, [FromQuery] bool? noBody,
     [ProducesError(StatusCodes.Status400BadRequest)]
     public async Task<IActionResult> SendMessage(SendMessageCommand request, CancellationToken cancellationToken)
     {
+        if (request.Recipients.Count > _options.MaxNumberOfMessageRecipients)
```
tnotheis, line 62: This is validation, so it should be part of command validator (`Backbone.Modules.Messages.Application.Messages.Commands.SendMessage.Validator`).

tnotheis (changes requested):

Modules/Messages/src/Messages.ConsumerApi/Controllers/MessagesController.cs
ericbrunner: how should I read the runtime instance property _options.MaxNumberOfMessageRecipients in the Validator? BTW on L:40 the pagesize gets checked the same way.

Modules/Messages/src/Messages.Application/Messages/Commands/SendMessage/Validator.cs
```diff
@@ -21,7 +22,9 @@ public Validator()
                 .SetValidator(new SendMessageCommandRecipientInformationValidator()));
 
         RuleFor(m => m.Recipients.Count)
-            .InclusiveBetween(1, 50).WithErrorCode(GenericApplicationErrors.Validation.InvalidPropertyValue().Code);
+            .LessThanOrEqualTo(options.Value.MaxNumberOfMessageRecipients)
+            .WithName("Recipients")
```
tnotheis, line 26: No need for a custom name, if I am not mistaken.

I guess you copied it from above. But there we only add it because due to the fact that it's an element of a list, the error message would be weird otherwise.

PR #1042, opened 2025-01-31
Push notifications all override each other

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/NotificationIdAttribute.cs
+++ b/Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/NotificationIdAttribute.cs
@@ -4,10 +4,8 @@ public class NotificationIdAttribute : Attribute
 {
     public NotificationIdAttribute(string value)
     {
-#if DEBUG
         if (value.Length > 64)
             throw new Exception("The value of the NotificationIdAttribute must be less than or equal to 64 characters.");
-#endif
 
         Value = value;
     }
--- a/Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/NotificationIdAttribute.cs
+++ b/Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/NotificationIdAttribute.cs
@@ -2,10 +2,15 @@ namespace Backbone.Modules.Devices.Application.Infrastructure.PushNotifications;
 
 public class NotificationIdAttribute : Attribute
 {
-    public NotificationIdAttribute(int value)
+    public NotificationIdAttribute(string value)
     {
+#if DEBUG
+        if (value.Length > 64)
+            throw new Exception("The value of the NotificationIdAttribute must be less than or equal to 64 characters.");
+#endif
+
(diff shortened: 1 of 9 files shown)


Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/NotificationIdAttribute.cs
```diff
@@ -2,10 +2,15 @@ namespace Backbone.Modules.Devices.Application.Infrastructure.PushNotifications;
 
 public class NotificationIdAttribute : Attribute
 {
-    public NotificationIdAttribute(int value)
+    public NotificationIdAttribute(string value)
     {
+#if DEBUG
+        if (value.Length > 64)
+            throw new Exception("The value of the NotificationIdAttribute must be less than or equal to 64 characters.");
+#endif
```
jkoenig134, line 10: what's that?

Modules/Devices/src/Devices.Infrastructure/PushNotifications/Connectors/Apns/ApnsMessageBuilder.cs
```diff
@@ -71,7 +75,8 @@ public HttpRequestMessage Build()
     private class Payload
     {
         [JsonPropertyName("notId")]
-        public int NotificationId { get; set; }
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        public string? NotificationId { get; set; }
```
jkoenig134, line 79: as discussed - we don't need this at all.

Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/NotificationIdAttribute.cs
tnotheis: That's because APNs allows not more than 64 characters for the x-collapse-id header (see https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html for reference)

Modules/Devices/src/Devices.Infrastructure/PushNotifications/Connectors/Apns/ApnsMessageBuilder.cs
tnotheis: Oops. I forgot about that.

Modules/Devices/src/Devices.Application/Infrastructure/PushNotifications/NotificationIdAttribute.cs
jkoenig134: But Why is this enabled only for debugging?

tnotheis: I had a stupid idea, which doesn't make any sense. :) Removed it.

jkoenig134 (approved):

PR #1062, opened 2025-02-27
Consumer Api: Support Http ETags and If-None-Match Header

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Contexts/TagsContext.cs
+++ b/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Contexts/TagsContext.cs
@@ -1,0 +1,6 @@
+﻿namespace Backbone.ConsumerApi.Tests.Integration.Contexts;
+
+public class TagsContext
+{
+    public Dictionary<string, string> TagHashes { get; set; } = [];
+}
--- a/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Tags/GET.feature
+++ b/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Tags/GET.feature
@@ -8,3 +8,8 @@ User requests available Tags
         Then the response status code is 200 (OK)
         And the response supports the English language
         And the response attributes contain tags
+
+    Scenario: Requesting the tags with the current hash
+        Given the most current hash h
+        When A GET request to the /Tags endpoint gets sent with hash h
+        Then the response status code is 304 (Not modified)
--- a/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/TagsStepDefinitions.cs
+++ b/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/TagsStepDefinitions.cs
@@ -1,6 +1,7 @@
 ﻿using Backbone.BuildingBlocks.SDK.Endpoints.Common.Types;
 using Backbone.ConsumerApi.Sdk.Endpoints.Tags.Types.Responses;
 using Backbone.ConsumerApi.Tests.Integration.Contexts;
+using Backbone.ConsumerApi.Tests.Integration.Extensions;
 using Backbone.ConsumerApi.Tests.Integration.Helpers;
 
 namespace Backbone.ConsumerApi.Tests.Integration.StepDefinitions;
@@ -10,14 +11,33 @@ public class TagsStepDefinitions
 {
     private readonly ResponseContext _responseContext;
     private readonly ClientPool _clientPool;
+    private readonly TagsContext _tagsContext;
 
-    private ApiResponse<ListTagsResponse>? _listTagsResponse;
+    private CachedApiResponse<ListTagsResponse>? _listTagsResponse;
 
     public TagsStepDefinitions(ResponseContext responseContext, ClientPool clientPool)
     {
         _responseContext = responseContext;
         _clientPool = clientPool;
+        _tagsContext = new TagsContext();
     }
+
+    #region Given
+
+    [Given($@"the most current hash {RegexFor.SINGLE_THING}")]
+    public async Task GivenTheMostCurrentHash(string hash)
+    {
+        await WhenAGETRequestToTheTagsEndpointGetsSent();
+
+        _listTagsResponse!.Should().BeASuccess();
+
+        _tagsContext.TagHashes[hash] = _listTagsResponse!.ETag;
+    }
+
+    #endregion
+
+
+    #region When
 
     [When("A GET request to the /Tags endpoint gets sent")]
     public async Task WhenAGETRequestToTheTagsEndpointGetsSent()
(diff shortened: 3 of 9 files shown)


Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Tags/GET.feature
```diff
@@ -8,3 +8,8 @@ User requests available Tags
         Then the response status code is 200 (OK)
         And the response supports the English language
         And the response attributes contain tags
+
+    Scenario: Requesting the tags with the current hash
+        Given the most current hash h
+        When A GET request to the /Tags endpoint gets sent with hash h
+        Then the response status code is 304 (Not modified)
```
tnotheis, line 15: I'd suggest a more generic phrasing (no need to leak the fact that the ETag is a hash; it might as well be a timestamp or an incrementing number)

Also, you should check that the response content is empty.

Given a list of tags l with an ETag e
And l didn't change since the last fetch
When a GET request to the /Tags endpoint gets sent with the If-None-Match header set to e
Then the response status code is 304 (Not modified)
And the response content is empty

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Contexts/TagsContext.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Backbone.ConsumerApi.Tests.Integration.Contexts;
+
+public class TagsContext
+{
+    public Dictionary<string, string> TagHashes { get; set; } = [];
```
tnotheis, line 5: Things that are not shared between different step definitions don't need to be placed in a context. You can just add this as a field to the step definitions class where it is used.

tnotheis: Actually... Do we need this at all? Why don't you just access the ETag directly via the response? Is it because it has a name in the step? Because the name might as well just be a way to make the step description clearer.

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/TagsStepDefinitions.cs
```diff
@@ -10,22 +11,52 @@ public class TagsStepDefinitions
...
+    private readonly TagsContext _tagsContext;
 
-    private ApiResponse<ListTagsResponse>? _listTagsResponse;
+    private CachedApiResponse<ListTagsResponse>? _listTagsResponse;
 
     public TagsStepDefinitions(ResponseContext responseContext, ClientPool clientPool)
     {
         _responseContext = responseContext;
         _clientPool = clientPool;
+        _tagsContext = new TagsContext();
```
tnotheis, line 22: Even though we will probably scratch the context: next time you need to use one, remember to not create it in the step definitions. Instead, let the DI container inject it for you. Otherwise, it cannot be shared between step definitions, which is the main purpose of the contexts.

Modules/Tags/src/Tags.ConsumerApi/Controllers/TagsController.cs
```diff
@@ -1,27 +1,43 @@
...
     [ProducesResponseType(typeof(HttpResponseEnvelopeResult<ListTagsResponse>), StatusCodes.Status200OK)]
     [AllowAnonymous]
-    public async Task<IActionResult> ListTags(CancellationToken cancellationToken)
+    public async Task<IActionResult> ListTags([FromHeader(Name = "If-None-Match")] string? ifNoneMatch, CancellationToken cancellationToken)
     {
         var response = await _mediator.Send(new ListTagsQuery(), cancellationToken);
+        var json = JsonSerializer.SerializeToUtf8Bytes(response);
+        var responseHash = _hasher.ComputeHash(json);
+        var hashString = UrlBase64.Encode(responseHash);
+        var etag = $"\"{hashString}\"";
```
tnotheis, line 34: Why does the etag contain double quotes?

Modules/Tags/src/Tags.ConsumerApi/Controllers/TagsController.cs
```diff
@@ -1,27 +1,43 @@
...
+
+        Response.Headers.ETag = etag;
+
+        if (ifNoneMatch != null && (ifNoneMatch.SequenceEqual(etag) || ifNoneMatch.SequenceEqual(hashString)))
```
tnotheis, line 38: Why do we need to compare with both the etag AND the hashString?

Sdks/ConsumerApi.Sdk/src/Endpoints/Tags/TagsEndpoint.cs
```diff
@@ -6,8 +6,8 @@ namespace Backbone.ConsumerApi.Sdk.Endpoints.Tags;
 
 public class TagsEndpoint(EndpointClient client) : ConsumerApiEndpoint(client)
 {
-    public async Task<ApiResponse<ListTagsResponse>> ListTags()
+    public async Task<CachedApiResponse<ListTagsResponse>> ListTags(CacheControl? cacheControl = null)
     {
-        return await _client.GetUnauthenticated<ListTagsResponse>($"api/{API_VERSION}/Tags");
+        return await _client.GetCachedUnauthenticated<ListTagsResponse>($"api/{API_VERSION}/Tags", null, null, cacheControl);
```
tnotheis, line 11: I would use a named parameter here, just to not confuse anybody with the two `null`s
```suggestion
        return await _client.GetCachedUnauthenticated<ListTagsResponse>($"api/{API_VERSION}/Tags", cacheControl: cacheControl);
```

tnotheis (changes requested):

Modules/Tags/src/Tags.ConsumerApi/Controllers/TagsController.cs
MH321Productions: To comply with the Http schema. Otherwise it can't be parsed by strict libraries.

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/Tags/GET.feature
```diff
@@ -8,3 +8,10 @@ User requests available Tags
         Then the response status code is 200 (OK)
         And the response supports the English language
         And the response attributes contain tags
+
+    Scenario: Requesting the tags with the current hash
+        Given a list of tags l with an ETag e
+        And I didn't change since the last fetch
```
tnotheis, line 14: Wrong letter :)
```suggestion
        And l didn't change since the last fetch
```

BuildingBlocks/src/BuildingBlocks.API/Mvc/ControllerAttributes/HandleHttpCachingAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+            var responseHeaders = response.GetTypedHeaders();
+
+            responseHeaders.ETag ??= GenerateETag(memoryStream);
+
+            if (IsClientCacheValid(requestHeaders, responseHeaders))
+            {
+                response.StatusCode = StatusCodes.Status304NotModified;
+
+                foreach (var header in response.Headers)
+                    if (!HEADERS_TO_KEEP_FOR304.Contains(header.Key))
```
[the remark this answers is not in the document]
MH321Productions: This will change nothing. The filter will only loop through the headers to keep, the if will always return false and therefore no header gets removed. It would be better to invert the filter and remove the if completely.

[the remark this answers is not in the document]
tnotheis: Hmm, that's exactly what I wanted to suggest. Guess my brain turned off temporarily while writing the comment. :D

tnotheis (approved):

PR #1068, opened 2025-03-04
Expect XSRF token on validate API key endpoint of the Admin API

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


MH321Productions:
When using Bruno, I get the XSRF validation error (and therefore a 400 Bad Request):
> The provided antiforgery token was meant for a different claims-based user than the current user.

The cookie and header get set and used during the request. Is this an error in Bruno?

jkoenig134 (dismissed):

jkoenig134 (approved):

PR #1136, opened 2025-05-05
Consumer Api: Access failed count is not incremented correctly when multiple concurrent requests for the same token are made

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/Database/EntityTypeBuilderExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+                v => v
+            );
+        }
+        else
+        {
+            return builder.Property(propertyExpression).HasColumnType("rowversion").HasConversion(
+                v => (byte[])v,
+                v => v
+            );
+        }
```
tnotheis, line 24: The default should be to throw an exception, in order to prevent using an unsupported column type.
```suggestion
        else if(builder.Metadata.Model.IsSqlServer())
        {
            return builder.Property(propertyExpression).HasColumnType("rowversion").HasConversion(
                v => (byte[])v,
                v => v
            );
        }
        else {
        	throw new Exception("No column type for a version has been defined for the database provider '{builder.Metadata.Model.GetDatabaseProvider}'".
        }
```

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/Database/EntityTypeBuilderExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+public static class EntityTypeBuilderExtensions
+{
+    public static PropertyBuilder<object> HasVersion<TEntity>(this EntityTypeBuilder<TEntity> builder, Expression<Func<TEntity, object>> propertyExpression) where TEntity : class
+    {
+        if (builder.Metadata.Model.IsNpgsql())
+        {
+            return builder.Property(propertyExpression).HasColumnType("xid").HasConversion(
+                v => (uint)v,
+                v => v
+            );
```
tnotheis, line 16: I just wondered: isn't it possible to use a shadow property here? So instead of explicitly defining the property in the class and using a lambda here to reference it, you could do something like this:
```c#
return builder
	.Property("Version")
	.HasColumnType("xid")
	.HasConversion(v => (uint)v, v => v);
```

This way the domain model doesn't get polluted with the "unnecessary" Version property.

Modules/Tokens/src/Tokens.Domain/Entities/Token.cs
```diff
@@ -57,6 +61,9 @@ public Token(IdentityAddress createdBy, DeviceId createdByDevice, byte[] content
     public DateTime CreatedAt { get; set; }
     public DateTime ExpiresAt { get; set; }
 
+    [Timestamp]
```
tnotheis, line 64: So you weren't able to do this with an EntityTypeConfiguration?

tnotheis (changes requested):

BuildingBlocks/src/BuildingBlocks.Application/MediatR/DbConcurrencyBehavior.cs
```diff
@@ -0,0 +1,29 @@
+﻿using MediatR;
+using Microsoft.EntityFrameworkCore;
+
+namespace Backbone.BuildingBlocks.Application.MediatR;
+
+public class DbConcurrencyBehavior<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse>
+    where TRequest : IRequest<TResponse>
+{
+    private readonly Random _random = new();
```
tnotheis, line 9: This can be static.

tnotheis: That's okay in this case.

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/Database/EntityTypeBuilderExtensions.cs
MH321Productions: Shadow properties work well and make the code even easier

Modules/Tokens/src/Tokens.Domain/Entities/Token.cs
MH321Productions: It's now included in the EntityTypeConfiguration

BuildingBlocks/src/BuildingBlocks.Infrastructure/Persistence/Database/EntityTypeBuilderExtensions.cs
MH321Productions: For future readers: Shadow properties don't work the way we wanted and we had to reverse this.

tnotheis (approved):

PR #1157, opened 2025-05-12
Migrate Integration tests from Specflow to Reqnroll

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


Applications/AdminApi/test/AdminApi.Tests.Integration/StepDefinitions/AnnouncementsStepDefinitions.cs
tnotheis, line 1: Why do you escape the `/` here, while in the other files you don't?

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/ChallengesStepDefinitions.cs
```diff
@@ -48,36 +48,36 @@ public async Task GivenAChallengeCreatedByAnAnonymousUser(string challengeName)
...
+    [When(@"an anonymous user sends a POST request to the \/Challenges endpoint")]
     public async Task WhenAnAnonymousUserSendsAPostRequestIsSentToTheChallengesEndpoint()
     {
         var client = _clientPool.Anonymous;
 
         _responseContext.WhenResponse = _challengeResponse = await client.Challenges.CreateChallengeUnauthenticated();
     }
 
-    [When($"{RegexFor.SINGLE_THING} sends a POST request to the /Challenges endpoint")]
+    [When(@$"{RegexFor.SINGLE_THING} sends a POST request to the \/Challenges endpoint")]
```
tnotheis, line 59: Why don't you use `{word}` here? There are some other occurrences of this. You should maybe do a global find and replace

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/DevicesStepDefinitions.cs
```diff
@@ -156,7 +156,7 @@ public async Task WhenDeviceSendsADeleteRequestToTheDeviceIdEndpointWithTheDevic
         _responseContext.WhenResponse = await client.Devices.DeleteDevice(deviceId);
     }
 
-    [When($"{RegexFor.SINGLE_THING} sends a DELETE request to the /Devices/{{id}} endpoint with a non existent id")]
+    [When(@"{word} sends a DELETE request to the \/Devices\/\{id\} endpoint with a non existent id")]
```
tnotheis, line 159: You sometimes don't use `^` and `$` in the regular expressions. Is there a specific reason for that?

tnotheis (commented):
`SpecFlow.NUnit` is currently part of the `ignoredPackages.json`. You can remove this now.

tnotheis (changes requested):
`SpecFlow.NUnit` is currently part of the `ignoredPackages.json`. You can remove this now.

Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/DevicesStepDefinitions.cs
tnotheis: I read about the Cucumber expressions shortly after sending my review. They sound pretty interesting. But I think we use too many `/`s and `{`s for it to be readable, which is why I would probably rather use regex. But I'd suggest that we discuss that in tomorrow's daily and vote on what to do.

MH321Productions:
The xUnit tests now work locally too

tnotheis (approved):

PR #1176, opened 2025-05-22
Database export via Admin CLI

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


Applications/AdminUi/packages/admin_api_types/lib/src/relationships/relationship.dart
```diff
@@ -6,7 +6,7 @@ part 'relationship.g.dart';
 class Relationship {
   final String peer;
   final String requestedBy;
-  final String templateId;
+  final String? templateId;
```
jkoenig134, line 9: Why is the template id optional now?

tnotheis: It is so for quite a while. We just never updated the type definitions in the flutter app.

The reason why it is optional is that the relationship template might get deleted.

MH321Productions (approved):
I didn't check the Admin UI stuff, but Julian did already

PR #1185, opened 2025-05-28
Rename methods, controllers, handlers and namespaces to `GetXyz` and `ListXyz` schema

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.


Applications/AdminApi/src/AdminApi/Controllers/AnnouncementsController.cs
```diff
@@ -35,10 +35,10 @@ public async Task<IActionResult> GetAnnouncement(string id, CancellationToken ca
     }
 
     [HttpGet]
-    [ProducesResponseType(typeof(GetAllAnnouncementsResponse), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(ListAllAnnouncementsResponse), StatusCodes.Status200OK)]
```
tnotheis, line 38: I think there was some misunderstanding. It shouldn't be `ListAll`, but just `List`. Because `List´ already implies that there are multiples.
But this shouldn't be much effort. A global find and replace will probably be sufficient.

I will continue my review once the change has been made.

tnotheis (changes requested):

tnotheis (approved):

PR #1212, opened 2025-06-13
SSE: Add keep alive events to prevent the connection from closing

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

# Stress Test Times

| Number of simultaneous connections | Time to create the clients | Time to create the connections | Time to process the events | Successful runs |
|------------------------------------|----------------------------|--------------------------------|----------------------------|-----------------|
| 100                                | 00:00:01                   | 00:00:01                       | 00:32:10                   | 100%            |
| 1000                               | 00:01:20                   | 00:00:39                       | 00:36:40                   | 100%            |
| 2000                               | 00:02:16                   | ?                              | 00:37:51                   | 24%             |

# Notes

- The database seems to be the bottleneck, since the maximum number of simultaneous connections (100) is reached.
- This seems to be very similar to the race condition during the token locking
- Multiple db operations are executed in sequence to create an identity
  - But new operation sequences are blocking the already running ones, therefore failing


Applications/SseServer/test/SseServerTests/EventQueueTests.cs
```diff
@@ -97,6 +99,27 @@ public async Task Dequeue_only_returns_elements_for_the_given_identity()
...
+        // Arrange
+        var config = new Configuration { Sse = new Configuration.SseConfiguration { KeepAliveEventInterval = 1 } };
+        var options = A.Fake<IOptions<Configuration>>();
+        A.CallTo(() => options.Value).Returns(config);
+
+        var queue = new EventQueue(A.Fake<ILogger<EventQueue>>(), options);
+        queue.Register("testAddress", CancellationToken.None);
+
+        // Act
+        await Task.Delay(TimeSpan.FromMilliseconds(500), TestContext.Current.CancellationToken);
```
tnotheis, line 115: Isn't it a bit unsafe to wait only 500ms when the interval is 1000ms?

Applications/SseServer/test/SseServerTests/EventQueueTests.cs
```diff
@@ -97,6 +99,27 @@ public async Task Dequeue_only_returns_elements_for_the_given_identity()
...
+        var eventNames = await queue.DequeueAllFor("testAddress");
+        queue.Deregister("testAddress");
```
tnotheis, line 117: Typically the act part has only one statement. I see that it is hard here. But I'd say that the one statement is the `Task.Delay`. Everything that comes after should go to Assert. 
Actually, the Deregister shouldn't even be necessary. The queue is disposed after the test anyway.

Applications/SseServer/src/SseServer/Configuration.cs
```diff
@@ -13,6 +13,8 @@ public class Configuration
     [Required]
     public InfrastructureConfiguration Infrastructure { get; set; } = new();
 
+    public SseConfiguration Sse { get; set; } = new();
```
tnotheis, line 16: I think it would be better if we call the configuration property just like the service that uses it (=> `SseServer` instead of `Sse`)

tnotheis (changes requested):

Applications/SseServer/test/SseServerTests/EventQueueTests.cs
MH321Productions: I'm not sure why I did it, but it worked every single time

Applications/SseServer/src/SseServer/Configuration.cs
```diff
@@ -29,4 +31,9 @@ public class InfrastructureConfiguration
         [Required]
         public EventBusConfiguration EventBus { get; set; } = new();
     }
+
+    public class SseConfiguration
```
tnotheis, line 35:
```suggestion
    public class SseServerConfiguration
```

tnotheis (approved):

PR #1215, opened 2025-06-16
Add ArchUnit Tests for Domain Events

MH321Productions:
# Readiness checklist

-   [x] I labeled the PR.

Note: The test regarding the constructors of the incoming domain events is currently failing, because some events contain constructors, which @erbenjak is fixing in his PR.

### DIFF
--- a/Backbone.Tests.ArchUnit/DomainEvents.cs
+++ b/Backbone.Tests.ArchUnit/DomainEvents.cs
@@ -1,0 +1,57 @@
+﻿using ArchUnitNET.Domain.Extensions;
+using ArchUnitNET.Fluent.Conditions;
+using ArchUnitNET.xUnit;
+using Backbone.BuildingBlocks.Domain.Events;
+using Backbone.Tooling.Extensions;
+using static ArchUnitNET.Fluent.ArchRuleDefinition;
+
+namespace Backbone.Backbone.Tests.ArchUnit;
+
+public class DomainEvents
+{
+    [Fact]
+    public void IncomingDomainEventsShouldNotHaveConstructors()
+    {
+        Types()
+            .That().AreAssignableTo(typeof(DomainEvent))
+            .And().ResideInNamespace("Backbone.Modules.*.Domain.DomainEvents.Incoming.*", true)
+            .Should().FollowCustomCondition(type =>
+            {
+                if (type.GetConstructors().Any(c => c.Parameters.Any()))
+                    return new ConditionResult(type, false, "should not have constructors");
+
+                return new ConditionResult(type, true);
+            }, "")
+            .Because("otherwise the JSON Serializer would call the constructor instead of serializing directly, which would cause an invalid serialization")
+            .Check(Backbone.ARCHITECTURE);
+    }
+
+    [Fact]
+    public void OutgoingDomainEventsShouldHaveCorrespondingIncomingDomainEvents()
+    {
+        var outgoing = Types()
+            .That().AreAssignableTo(typeof(DomainEvent))
+            .And().ResideInNamespace("Backbone.Modules.*.DomainEvents.Outgoing.*", true)
+            .GetObjects(Backbone.ARCHITECTURE)
+            .ToList();
+
+        Types()
+            .That().AreAssignableTo(typeof(DomainEvent))
+            .And().ResideInNamespace("Backbone.Modules.*.DomainEvents.Incoming.*", true)
+            .Should().FollowCustomCondition(type =>
+            {
+                var correspondingTypes = outgoing.Where(t => t.NameMatches(type.Name)).ToList();
+                if (correspondingTypes.IsEmpty()) return new ConditionResult(type, false, "does not have a corresponding outgoing domain event");
+
+                foreach (var property in type.GetPropertyMembers())
+                {
+                    foreach (var t in correspondingTypes.Where(t => !t.HasPropertyMemberWithName(property.Name)))
+                        return new ConditionResult(type, false, $"does not have a corresponding property for '{property.Name}' in its corresponding domain event ({t.FullName})");
+                }
+
+                return new ConditionResult(type, true);
+            }, "")
+            .Because("Reason")
+            .Check(Backbone.ARCHITECTURE);
+    }
+}
--- a/Backbone.Tests.ArchUnit/DomainEvents.cs
+++ b/Backbone.Tests.ArchUnit/DomainEvents.cs
@@ -1,6 +1,8 @@
-﻿using ArchUnitNET.Domain.Extensions;
+﻿using ArchUnitNET.Domain;
+using ArchUnitNET.Domain.Extensions;
 using ArchUnitNET.Fluent.Conditions;
-using ArchUnitNET.xUnit;
+using ArchUnitNET.Fluent.Syntax.Elements.Types;
+using ArchUnitNET.xUnitV3;
 using Backbone.BuildingBlocks.Domain.Events;
 using Backbone.Tooling.Extensions;
 using static ArchUnitNET.Fluent.ArchRuleDefinition;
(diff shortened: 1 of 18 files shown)


erbenjak:
The constructors of the incoming domain events are removed in PR #1218

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Backbone.Backbone.Tests.ArchUnit;
+
+public class DomainEvents
+{
+    [Fact]
+    public void IncomingDomainEventsShouldNotHaveConstructors()
+    {
+        Types()
+            .That().AreAssignableTo(typeof(DomainEvent))
+            .And().ResideInNamespace("Backbone.Modules.*.Domain.DomainEvents.Incoming.*", true)
```
tnotheis, line 17: I think we might as well just check for ".*Incoming.*". I don't think there will ever be and outgoing domain event that is in a namespace containing "Incoming". :-)

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+            .Should().FollowCustomCondition(type =>
+            {
+                if (type.GetConstructors().Any(c => c.Parameters.Any()))
+                    return new ConditionResult(type, false, "should not have constructors");
```
tnotheis, line 21:
```suggestion
                    return new ConditionResult(type, false, "should not have a constructor with parameters");
```

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                return new ConditionResult(type, true);
+            }, "")
+            .Because("otherwise the JSON Serializer would call the constructor instead of serializing directly, which would cause an invalid serialization")
```
tnotheis, line 25:
```suggestion
            .Because("objects are only created by the JSON Serializer anyway. Adding a constructor with parameters only causes errors during deserialization if the parameter names don't match the property names.")
```

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                foreach (var property in type.GetPropertyMembers())
+                {
+                    foreach (var t in correspondingTypes.Where(t => !t.HasPropertyMemberWithName(property.Name)))
+                        return new ConditionResult(type, false, $"does not have a corresponding property for '{property.Name}' in its corresponding domain event ({t.FullName})");
+                }
+
+                return new ConditionResult(type, true);
+            }, "")
+            .Because("Reason")
```
tnotheis, line 54: That's not a good reason. ;)

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        var outgoing = Types()
+            .That().AreAssignableTo(typeof(DomainEvent))
+            .And().ResideInNamespace("Backbone.Modules.*.DomainEvents.Outgoing.*", true)
+            .GetObjects(Backbone.ARCHITECTURE)
+            .ToList();
+
+        Types()
+            .That().AreAssignableTo(typeof(DomainEvent))
+            .And().ResideInNamespace("Backbone.Modules.*.DomainEvents.Incoming.*", true)
```
tnotheis, line 40: Since you're using the incoming domain events multiple times, it probably makes sense to introduce a field for them.
And for the sake of consistency, you could do the same for the outgoing ones

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+                return new ConditionResult(type, true);
+            }, "")
+            .Because("otherwise the JSON Serializer would call the constructor instead of serializing directly, which would cause an invalid serialization")
+            .Check(Backbone.ARCHITECTURE);
+    }
+
+    [Fact]
+    public void OutgoingDomainEventsShouldHaveCorrespondingIncomingDomainEvents()
+    {
+        var outgoing = Types()
```
tnotheis, line 32:
```suggestion
        var outgoingDomainEvents = Types()
```

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+                    return new ConditionResult(type, false, "should not have constructors");
+
+                return new ConditionResult(type, true);
+            }, "")
+            .Because("otherwise the JSON Serializer would call the constructor instead of serializing directly, which would cause an invalid serialization")
+            .Check(Backbone.ARCHITECTURE);
+    }
+
+    [Fact]
+    public void OutgoingDomainEventsShouldHaveCorrespondingIncomingDomainEvents()
```
tnotheis, line 30:
```suggestion
    public void IncomingDomainEventsShouldHaveCorrespondingOutgoingDomainEvents()
```

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+            .And().ResideInNamespace("Backbone.Modules.*.DomainEvents.Incoming.*", true)
+            .Should().FollowCustomCondition(type =>
+            {
+                var correspondingTypes = outgoing.Where(t => t.NameMatches(type.Name)).ToList();
+                if (correspondingTypes.IsEmpty()) return new ConditionResult(type, false, "does not have a corresponding outgoing domain event");
+
+                foreach (var property in type.GetPropertyMembers())
+                {
+                    foreach (var t in correspondingTypes.Where(t => !t.HasPropertyMemberWithName(property.Name)))
+                        return new ConditionResult(type, false, $"does not have a corresponding property for '{property.Name}' in its corresponding domain event ({t.FullName})");
```
tnotheis, line 49:
```suggestion
                        return new ConditionResult(type, false, $"does not have a corresponding property for '{property.Name}' in its corresponding outgoing domain event ({t.FullName})");
```

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+        Types()
+            .That().AreAssignableTo(typeof(DomainEvent))
+            .And().ResideInNamespace("Backbone.Modules.*.Domain.DomainEvents.Incoming.*", true)
+            .Should().FollowCustomCondition(type =>
+            {
+                if (type.GetConstructors().Any(c => c.Parameters.Any()))
+                    return new ConditionResult(type, false, "should not have constructors");
+
+                return new ConditionResult(type, true);
+            }, "")
```
tnotheis, line 24: You have to add a description here. Because currently the error message this test produces looks as follows:

> "Types that are assignable to "Backbone.BuildingBlocks.Domain.Events.DomainEvent" and reside in namespace with full name matching "Backbone.Modules.*.Domain.DomainEvents.Incoming.*" should  because otherwise the JSON Serializer would call the constructor instead of serializing directly, which would cause an invalid serialization"

(note that there is something missing between "should" and "because".

Backbone.Tests.ArchUnit/DomainEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+                if (correspondingTypes.IsEmpty()) return new ConditionResult(type, false, "does not have a corresponding outgoing domain event");
+
+                foreach (var property in type.GetPropertyMembers())
+                {
+                    foreach (var t in correspondingTypes.Where(t => !t.HasPropertyMemberWithName(property.Name)))
+                        return new ConditionResult(type, false, $"does not have a corresponding property for '{property.Name}' in its corresponding domain event ({t.FullName})");
+                }
+
+                return new ConditionResult(type, true);
+            }, "")
```
tnotheis, line 53: Add a description instead of an empty string

tnotheis (changes requested):

Backbone.Tests.ArchUnit/DomainEvents.cs
MH321Productions: Oops

tnotheis (changes requested):

tnotheis (approved):

PR #1228, opened 2025-06-25
chore: Remove newtonsoft json parser

erbenjak:
# Readiness checklist

-   [X] I labeled the PR.

### DIFF
--- a/Sdks/AdminApi.Sdk/src/Endpoints/Challenges/Types/SignedChallenge.cs
+++ b/Sdks/AdminApi.Sdk/src/Endpoints/Challenges/Types/SignedChallenge.cs
@@ -1,6 +1,7 @@
+using System.Text;
+using System.Text.Json;
 using Backbone.BuildingBlocks.SDK.Crypto;
 using Backbone.Crypto;
-using Newtonsoft.Json;
 
 namespace Backbone.AdminApi.Sdk.Endpoints.Challenges.Types;
 
@@ -9,9 +10,9 @@ public class SignedChallenge
     public SignedChallenge(string challenge, ConvertibleString signature)
     {
         Challenge = challenge;
-        Signature = ConvertibleString.FromUtf8(
-            JsonConvert.SerializeObject(new CryptoSignatureSignedChallenge { alg = CryptoHashAlgorithm.SHA512, sig = signature.BytesRepresentation }
-            )).Base64Representation;
+        var sig = new CryptoSignatureSignedChallenge { alg = CryptoHashAlgorithm.SHA512, sig = signature.BytesRepresentation };
+        var json = JsonSerializer.Serialize(sig, new JsonSerializerOptions { IncludeFields = true });
+        Signature = ConvertibleString.FromByteArray(Encoding.UTF8.GetBytes(json)).Base64Representation;
     }
 
     public string Challenge { get; internal set; }
--- a/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/IdentitiesStepDefinitions.cs
+++ b/Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/StepDefinitions/IdentitiesStepDefinitions.cs
@@ -1,9 +1,9 @@
-﻿using Backbone.BuildingBlocks.SDK.Crypto;
+﻿using System.Text.Json;
+using Backbone.BuildingBlocks.SDK.Crypto;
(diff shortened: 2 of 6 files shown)


Sdks/AdminApi.Sdk/src/Endpoints/Challenges/Types/SignedChallenge.cs
erbenjak, line 1: i think this behaviour was missunderstood - as also happend to me initially

after serializing the json was always empty and a short google search leads you down a rabbit hole of System.Json not being able to parse byte[], but the true error was that fields were ignored

BuildingBlocks/src/BuildingBlocks.Infrastructure/EventBus/RabbitMQ/EventBusRabbitMq.cs
```diff
@@ -244,8 +238,7 @@ public async Task Publish(DomainEvent @event)
         var eventName = @event.GetEventName();
 
         _logger.LogInformation("Creating RabbitMQ channel to publish a '{EventName}'.", eventName);
-
-        var message = JsonConvert.SerializeObject(@event, JSON_SERIALIZER_SETTINGS);
+        var message = JsonSerializer.Serialize(@event, @event.GetType());
```
tnotheis, line 241: The autistic Me forces me to write this comment:

Can you re-add the blank line between this one and the LogInformation please? 😉

Modules/Devices/src/Devices.Domain/Entities/Identities/Identity.cs
```diff
@@ -330,7 +330,8 @@ public void ChangeFeatureFlags(Dictionary<FeatureFlagName, bool> featureFlags)
         {
             FeatureFlags.Set(keyValuePair.Key, keyValuePair.Value);
         }
-        RaiseDomainEvent(new FeatureFlagsOfIdentityChangedDomainEvent(this));
+
+        RaiseDomainEvent(new FeatureFlagsOfIdentityChangedDomainEvent(Address.Value));
```
tnotheis, line 334: We usually pass the whole domain object so that the event's constructor can do the mapping (just as we did it before your changes). Why did you decide to go a different way?

Modules/Files/src/Files.Domain/DomainEvents/Outgoing/FileUploadedDomainEvent.cs
```diff
@@ -11,6 +11,6 @@ public FileUploadedDomainEvent(File file) : base($"{file.Id}/Created")
         FileId = file.Id.ToString();
     }
 
-    public string FileId { get; private set; }
-    public string Owner { get; private set; }
+    public string FileId { get; init; }
+    public string Owner { get; init; }
```
tnotheis, line 15: Isn't a private setter okay in case of an outgoing event?

Modules/Synchronization/test/Synchronization.Application.Tests/Tests/DomainEvents/IdentityDeletionProcessStartedDomainEventHandlerTests.cs
```diff
@@ -32,7 +32,8 @@ public async Task Does_nothing_if_initiator_is_deletion_process_owner()
     {
         // Arrange
         var deletionProcessOwner = CreateRandomIdentityAddress();
-        var identityDeletionProcessStartedDomainEvent = new IdentityDeletionProcessStartedDomainEvent { Address = deletionProcessOwner, DeletionProcessId = "some-deletion-process-id", Initiator = deletionProcessOwner };
+        var identityDeletionProcessStartedDomainEvent = new IdentityDeletionProcessStartedDomainEvent
+        { Address = deletionProcessOwner, DeletionProcessId = "some-deletion-process-id", Initiator = deletionProcessOwner };
```
tnotheis, line 36: The formatting seems a bit weird.

tnotheis (changes requested):

Modules/Files/src/Files.Domain/DomainEvents/Outgoing/FileUploadedDomainEvent.cs
erbenjak: sure - but I wanted to keep things consistant. Else this would be the only DomainEvent class with private setters... that seems odd to me

PR #1270, opened 2025-08-06
Docker compose: omit first time authentication for seq

erbenjak:
# Readiness checklist

-   [X] I labeled the PR.


tnotheis:
I changed the label because we only use bug in case of a bug in *production* code. The change you made only affects development though.
If we kept the bug label, it would show up in the release notes, which would be confusing.

PR #1282, opened 2025-08-21
One database connection is kept open per active SSE connection

tnotheis:
# Readiness checklist

-   [x] I labeled the PR.


tnotheis:
After doing some research, it seems I overreacted a bit with this fix. 😄 
Having a DbContext doesn't mean that a database connection is blocked, because the DbContexts only take a connection from the connection pool when they run an operation, and return it into the pool immediately afterwards.
So the only downside of having multiple DbContexts is the additional use of memory. But I looked into a memory dump, and a single DbContext instance only seems to take ~2kB of memory, which should be negligible. Even if there are 50 concurrent SSE connections, it's just 1MB.

PR #1488, opened 2026-07-09
Update renovate config

tnotheis:
# Readiness checklist

- [x] I labeled the PR.
- [x] I added/updated tests (if necessary).
- [x] I self-reviewed the PR.


.github/renovate.json
```diff
@@ -0,0 +1,49 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": ["github>js-soft/renovate-config"],
+    "reviewers": ["Milena-Czierlinski", "britsta"],
```
Milena-Czierlinski, line 4: @tnotheis Do I see it correctly that Britta and I are now reviewers of every backbone renovate PR? Shouldn't that be Jakob and you?

tnotheis: Hmm. I guess my codex prompt wasn't specific enough in this case. :) I told it to keep the original reviewers and copy the remaining properties. But in case of the backbone, there were only "additionalReviewers" set. I force-pushed the fix onto the main branch.

PR #1536, opened 2026-09-03
Add Consumer API OpenID4VP verifier site

tnotheis:
## Summary
- extend the existing Consumer API `/r/{referenceId}` onboarding page with an embedded mobile-first OpenID4VP verifier
- render the verifier markup in `AppOnboarding.cshtml`; the TypeScript bundle now only controls existing DOM nodes and performs browser-only loading, decryption, validation, and state updates
- parse the NMSHD reference hash, fetch Token/RelationshipTemplate content, decrypt XCHACHA20_POLY1305 token content with `@nmshd/crypto`, and fall back to the previous onboarding flow when the content is not a VP token
- validate `TokenContentVerifiablePresentation.value` with the existing Credo-based verifier and render loading/valid/invalid/closed states without a mock app/phone shell
- extract displayed credential values from the Verifiable Presentation only; missing values are omitted instead of being replaced with sample/default credential data
- build Vite output into `OpenId4VpVerifierSite/dist` and let the Consumer API MSBuild target copy it into `wwwroot/openid4vp-verifier` at local build time; generated `wwwroot` assets are no longer committed
- build verifier assets in a separate Docker Node stage and copy them into the .NET build stage, so the DHI .NET SDK image does not need npm

## Notes
- For this first version only reference algorithm `3` / `XCHACHA20_POLY1305` is handled.
- Credential status checks remain disabled for the minimal verifier.
- The Vite build warnings about `node:fs` externalization and direct `eval` come from transitive crypto/Credo dependencies.
- Local targeted MTP test execution starts the integration test assembly but hangs without producing a result in this environment; the integration test project itself builds successfully.

## Screenshot
image


Applications/ConsumerApi/test/ConsumerApi.Tests.Integration/Features/AppOnboarding/GET.feature
```diff
@@ -20,3 +20,8 @@ User requests app onboarding page
         When a call is made to the /r/resourceId endpoint for the app "enmeshed"
         Then the response contains a link with the domain name "apps.apple.com"
         And the response contains a link with the domain name "play.google.com"
+
+    Scenario: User requests app onboarding page and verifier bootstrap is contained on the response page
+        Given an http client with user agent "unknown"
+        When a call is made to the /r/resourceId endpoint for the app "enmeshed"
+        Then the response contains the OpenID4VP verifier bootstrap
```
tnotheis: This test doesn't deliver a lot of value and can be removed. Remember to remove the step definition as well.

tnotheis:
@codex review

